PR #27, opened 2020-03-25
Update link to workflow doc

joelrebel:
A minor fix :)


gauravgahlot:
LGTM @joelrebel 
Thanks for the fix. :+1:

PR #39, opened 2020-04-02
Added nginx setup in docker-compose setup process

parauliya:


setup_with_docker_compose.sh
```diff
@@ -35,7 +35,26 @@ function setup_network () {
...
+    chmod +x workflow-helper.sh
+    
+    cd /tmp
+    curl 'https://packet-osie-uploads.s3.amazonaws.com/osie-v19.10.23.00-n=55,c=be58d67,b=master.tar.gz' -o osie.tar.gz
+    tar -zxvf osie.tar.gz
+    cd /tmp/'osie-v19.10.23.00-n=55,c=be58d67,b=master'
+    cp -r grub /packet/nginx/misc/osie/current/
+    cp modloop-x86_64 /packet/nginx/misc/osie/current/
+    cp initramfs-x86_64 /packet/nginx/misc/osie/current/
+    cp vmlinuz-x86_64 /packet/nginx/misc/osie/current/
```
nathangoulding, line 53: why is this not cp *?

parauliya: Because there are lot of others file as well which will acquire a lot of size. That's why after copying required files I have deleted the folder in the next line, which is of size 2.2 GB.

PR #42, opened 2020-04-03
Creating setup with "Terraform scripts" in Packet infrastructure

parauliya:
I have added a combination of `terraform  and bash scripts` so when you run the command "terraform apply" it will automatically create provisioner and worker machine and the compelete stack on the provisioner will be up. 
There is `input.tf` file in which user can provide the values required for the setup like `Packer API AUTH Token`, `project id`, `host ip`, `netmask` etc. 
I have also added the instructions for that in docs/setup.md file.


vielmetti:
Please include a description in the PR text to help reviewers know your intent and thoughts when proposing a patch.

PR #45, opened 2020-04-05
Grouped all things demo together

kdeng3849:
- combined setup with terraform with setup with docker compose 
  - automate just the creating servers part with tf, then manually run the setup script
- included the contents from the `first-good-workflow` branch to avoid having to switch branches back and forth


vielmetti:
Please provide a description of the changes to assist in reviews, thanks!

PR #63, opened 2020-04-13
Fix for Issue#61: Restarting tink will not reset the database and hardware data will persist. 

parauliya:
Root Cause of issue:
      When tink tries to start it calls connecetdb  which was creating the connection with the db and calling the trauncate function which trucate the hardware table in the database because of which all the entries in the hardware table were being deleted.

Fix:
       I have removed the tructate function because of which trucation of harware table will not happen and all the entries of harware table will persist as it is even after the restart of tink.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -19,12 +19,6 @@ func ConnectDB(lg log.Logger) *sql.DB {
 		logger.Error(err)
 		panic(err)
 	}
-	if err := truncate(db); err != nil {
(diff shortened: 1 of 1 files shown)


mmlb:
can you fixup the commit so that this PR message is actually in the commit message too please. Also the "title" of your commit should be more descriptive instead of just "fix issue #" that doesn't really help when looking through history.

mmlb:
diff lgtm though

PR #64, opened 2020-04-14
add endpoints for crud operations & update hardware data model

kdeng3849:
Changes: 
- Added endpoints for the following operations ([unofficial api](https://gist.github.com/kdeng3849/11df3ed5d6ad4a003f9047386afbfcad)):
  - `tink hardware push/all/mac/ip/id`
  - `tink template create/get/list/delete`
  - `tink workflow create/get/list/delete/state/events`
- Added basic auth to above endpoints
  - Added two new envvars `TINK_AUTH_USERNAME` and `TINK_AUTH_PASSWORD` to set basic auth username and password
- Changed the `data` field in `WorkflowTemplate` to be a string instead of bytes (in template proto)
- Updated [`Hardware`](https://github.com/tinkerbell/tink/pull/64/files#diff-730638eb839b6ba1934978ee53a1aafeR54) struct to new data model
- Updated `docker-compose.yml` to reflect additions of new envvars across all three prs

TODO:
- [ ] update setup.sh and docker-compose.yml to account for new envvars
  - [ ] setup.sh
  - [x] docker-compose.yml

Related PRs:
boots: https://github.com/tinkerbell/boots/pull/27
hegel: https://github.com/tinkerbell/hegel/pull/13


gauravgahlot:
@kdeng3849 Can you please rebase the branch?

cli/tink/cmd/hardware/push.go
```diff
@@ -34,7 +34,13 @@ var pushCmd = &cobra.Command{
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		for _, j := range args {
-			if _, err := client.HardwareClient.Push(context.Background(), &hardware.PushRequest{Data: j}); err != nil {
+			fmt.Println(j)
```
gauravgahlot, line 37: Do we really need this?

cli/tink/cmd/template/create.go
```diff
@@ -62,7 +62,7 @@ func readTemplateData() []byte {
 }
 
 func createTemplate(c *cobra.Command, args []string) {
-	req := template.WorkflowTemplate{Name: templateName, Data: readTemplateData()}
+	req := template.WorkflowTemplate{Name: templateName, Data: string(readTemplateData())}
```
gauravgahlot, line 65: Since the returned value of `readTemplateData` is converted to `string` in most of the places, I think it would make sense to change the return type to `string` from `[]byte`.

alexellis:
Looks like another rebase is required on this PR @kdeng3849

cmd/tink-cli/cmd/hardware/ip.go
```diff
@@ -28,11 +29,15 @@ var ipCmd = &cobra.Command{
 	},
 	Run: func(cmd *cobra.Command, args []string) {
 		for _, ip := range args {
-			hw, err := client.HardwareClient.ByIP(context.Background(), &hardware.GetRequest{IP: ip})
+			hw, err := client.HardwareClient.ByIP(context.Background(), &hardware.GetRequest{Ip: ip})
```
mmlb, line 32: Is there a way for the generated code to influence the name for the generated struct fields? Looking at `Ip` -> `IP` and `Mac` -> `MAC` and maybe others?

kdeng3849: I didn't really look into it too much since it seemed to work fine with boots without any issue. Is there a reason/necessity for the names to be capitalized? I could change them back to uppercase in the proto if needed

mmlb: its a code nit, but acronyms should be in all caps. It also matches the types in the stdlib so not being an oddball would be good.

kdeng3849: okay, I'll change it back then

kdeng3849: I remember why I decided on all lowercase, it's because there are problems when the message name is the same as the field name. For example:
```
message RAID {
	string name = 1;
	string level = 2;
	repeated string devices = 3;
	int64 spare = 4;
}

repeated RAID RAID = 5;
```
this is giving me the error:
```
Generating hardware.pb.go...
hardware/hardware.proto:164:41: "RAID" is already defined in "github.com.tinkerbell.tink.protos.hardware.Hardware.Metadata.Instance.Storage".
hardware/hardware.proto:175:42: "RAID" is not defined.
```
So I decided on all lowercase for the field name (which is also the convention for protobuf).  
I could still rename the message name to something like "Raid", so I guess it all boils down to which convention to follow (go's or protobuf's). Which one should I go with?

mmlb: ah ok, sigh. lets go with protobuf then. I'm curious is there an equivalent of `json:"SomeKeyName"` metadata we can add so we can both?

mmlb: looks like no. https://github.com/golang/protobuf/issues/555. In that case lets stay with protobuf norms. Transforming protoc generated structs into human readable/meaningful structs is something that I think always makes sense and we could address then.

kdeng3849: Hmm what do you mean by have both? like in terms of how the json is marshaled/unmarshalled?

kdeng3849: okay, gotcha

cmd/tink-cli/cmd/hardware/all.go
```diff
@@ -24,7 +25,11 @@ var allCmd = &cobra.Command{
 		var hw *hardware.Hardware
 		err = nil
 		for hw, err = alls.Recv(); err == nil && hw != nil; hw, err = alls.Recv() {
-			fmt.Println(hw.JSON)
+			b, err := json.Marshal(hw)
+			if err != nil {
+				log.Fatal(err)
```
yetang-equinix, line 30: This is just to retrieve hardware info, may be one hardware data cannot be converted to json properly, we can just log an error and go on for next. Why do we have to log fatal, which will crash the cli agent and make it harder to debug.

grpc-server/hardware.go
```diff
@@ -23,39 +23,33 @@ func (s *server) Push(ctx context.Context, in *hardware.PushRequest) (*hardware.
...
-		State string
-	}
-	err := json.Unmarshal([]byte(in.Data), &h)
-	if err != nil {
+	hw := in.Data
+	if hw.Id == "" {
 		metrics.CacheTotals.With(labels).Inc()
 		metrics.CacheErrors.With(labels).Inc()
-		err = errors.Wrap(err, "unmarshal json")
+		err := errors.New("id must be set to a UUID")
```
yetang-equinix, line 30: We can indicate that current hw.Id is nil string in this case (add hw.Id to error log). This will give extra info for debugging.

grpc-server/hardware.go
```diff
@@ -75,30 +69,19 @@ func (s *server) Push(ctx context.Context, in *hardware.PushRequest) (*hardware.
...
 		select {
-		case ch <- in.Data:
+		case ch <- string(data):
 		default:
 			metrics.WatchMissTotal.Inc()
-			logger.With("id", h.ID).Info("skipping blocked watcher")
+			logger.With("id", hw.Id ).Info("skipping blocked watcher")
 		}
 	}
 	s.watchLock.RUnlock()
```
yetang-equinix, line 80: Logging is slow, it should not be done between watch lock/unlock. Critical logging for error case is fine. But for information logging, we should do it after unlock.

http-server/http-server.go
```diff
@@ -3,34 +3,72 @@ package httpserver
...
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/tinkerbell/tink/protos/hardware"
+	"github.com/tinkerbell/tink/protos/template"
+	"github.com/tinkerbell/tink/protos/workflow"
 )
 
 var (
 	gitRev         = "unknown"
 	gitRevJSON     []byte
+	grpcListenAddr = "localhost:42113"
```
yetang-equinix, line 28: this port number should be in a config file, not hard coded  here.

http-server/http-server.go
```diff
@@ -3,34 +3,72 @@ package httpserver
...
 )
 
 // SetupHTTP setup and return an HTTP server
 func SetupHTTP(ctx context.Context, lg log.Logger, certPEM []byte, modTime time.Time, errCh chan<- error) {
 	logger = lg
+
+	cp := x509.NewCertPool()
+	ok := cp.AppendCertsFromPEM(certPEM)
+	if !ok {
+		logger.Error(errors.New("parse cert"))
```
yetang-equinix, line 43: please add returned error code to logging

http-server/http-server.go
```diff
@@ -3,34 +3,72 @@ package httpserver
...
+	}
+
 	http.HandleFunc("/cert", func(w http.ResponseWriter, r *http.Request) {
 		http.ServeContent(w, r, "server.pem", modTime, bytes.NewReader(certPEM))
 	})
 	http.Handle("/metrics", promhttp.Handler())
 	setupGitRevJSON()
 	http.HandleFunc("/version", versionHandler)
 	http.HandleFunc("/_packet/healthcheck", healthCheckHandler)
+	http.Handle("/", BasicAuth(mux))
```
nathangoulding, line 71: I think this should be wrapped in an if statement that checks if `TINK_AUTH_USERNAME` or `TINK_AUTH_PASSWORD` are set.

http-server/http-server.go
kdeng3849: How should the config file look like? Would it suffice if I just have the port be read from an environment variable and set a default port for when the variable unset?

http-server/http-server.go
kdeng3849: What are you referring to with "returned error code"? If you meant the `ok`, that is just a boolean

yetang-equinix: agree

http-server/http-server.go
yetang-equinix: Yes, there shall be a default value if there is no configured value. If we have quite some values, a file will be easy to manage.

kdeng3849: I think the closest to the config file you were referring to is the [`docker-compose.yml`](https://github.com/tinkerbell/tink/blob/c49b6c2dec7df60a2588c18617ea2e72c3aeb68d/deploy/docker-compose.yml#L18) that is currently used to run all of the services.  
I have added two new environment variables to allow the user to specify their own ports.

Please refer to this commit here: https://github.com/tinkerbell/tink/pull/64/commits/aee3b9f33e1ddae86c91db0999c9b3fc022bbf0e

deploy/docker-compose.yml
```diff
@@ -128,32 +131,6 @@ services:
     volumes:
       - ./state/webroot:/usr/share/nginx/html/
 
-  cacher:
```
yetang-equinix, line 131: So we totally get ride of cacher from tinkerbell deployment. But do we still deploy cacher from other place? I can still see Hegel still interact with cacher in the other PR.

kdeng3849: yeah, cacher is still used in other places, just not in tinkerbell

PR #77, opened 2020-04-23
Move all the binaries and respective files into cmd/

mmlb:
Keeps all binaries and their support files grouped nicely instead of
littered randomly.

Signed-off-by: Manuel Mendez <manny@packet.com>


gauravgahlot:
@mmlb Please resolve the conflicts.

mmlb:
@gauravgahlot updated PTAL

cmd/tink-server/Dockerfile
```diff
@@ -7,6 +7,6 @@ EXPOSE 42114
 RUN apk add --update ca-certificates && \
     apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing cfssl
 
-COPY deploy/db/migrate /migrate
-COPY deploy/db/docker-entrypoint-initdb.d/tinkerbell-init.sql /init.sql
+COPY db/migrate /migrate
+COPY db/docker-entrypoint-initdb.d/tinkerbell-init.sql /init.sql
```
gauravgahlot, line 11: It should be `deploy/db/`.

gauravgahlot:
@mmlb It would be great if we can conclude this PR asap.

mmlb:
@gauravgahlot should be good now, PTAL.

mmlb:
@gauravgahlot hmm that makes sense, I'll split the deploy files into those needed by tink-server and those for other things then.

PR #78, opened 2020-04-24
Fixes #38: Using persistent volume to hold registry data

### LINKED ISSUE #38: Use persistent volume to hold registry data
**Current**
The private registry created with the `docker-compose.yml` used to set up the workflow environment stores all the data in an anonymous volume. 

**Expected**
The registry should use persistent volumes.

parauliya:
In this fix we are using "/var/tinkerbell/registry" directory locally to persist the data of private registry we are creating to have worker and action images.

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -92,6 +92,7 @@ services:
       REGISTRY_AUTH_HTPASSWD_PATH: /auth/htpasswd
     volumes:
       - ./certs:/certs
+      - /var/tinkerbell/registry:/var/lib/registry
     depends_on:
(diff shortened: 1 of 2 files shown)


gauravgahlot:
@parauliya Can you rebase the branch, please?

parauliya:
> @parauliya Can you rebase the branch, please?

Done. Branch Updated.

mmlb:
This seems fine, though I don't like to assume we can grab a dir in /var. It would be nice to have support for a $PREFIX or similar so users can tailor where our files/dirs land. I'll open that up as a separate issue though.

PR #88, opened 2020-04-30
Merge hardware and target.

parauliya:
Now while creating a workflow you can directly provide hardware/devices mac addr.
And now there is no need of targets. This approach will support both single and multiple workers as well:

Current `template` look like this for `worker` field: ( Has the `Target` reference )
- `worker: "{{index .Targets "machine1" "mac_addr"}}"` 

After this PR the `template` will look like this for `worker` field: (`Target` reference not required)
- `worker: {{.device_1}}`              

Right Now the `workflow create` command is as follows for single and multiple worker: ( Has the `Target` reference )
-  `tink workflow create -t <temp-id> -r <target-id>` 

After this PR the `workflow create` command it will look like:  ( `Target` reference not required )
1. For single worker
   ` tink workflow create -t <template_id> -r '{ "device_1": "mac/IP/Id" }'`   

2. For Multiple worker you can just increase the key value pair in the workflow create command:
   `tink workflow create -t <template_id>  -r '{ "device_1": "mac/IP", "device_2": "mac/IP"}'`  

**_Note: In the above case user can replace the keys device_1/device_2 with any name which should contain only `letters, numbers and Underscore` and should be in sync with template._**


grpc-server/workflow.go
```diff
@@ -267,35 +268,29 @@ func createYaml(ctx context.Context, sqlDB *sql.DB, temp string, tar string) (st
...
 func renderTemplate(tempData string, tarData []byte) (string, error) {
-	type machine map[string]string
-	type Environment struct {
-		Targets map[string]machine `json:"targets"`
-	}
-
-	var env Environment
-	t := template.New("workflow-template")
-	_, err := t.Parse(tempData)
+	var hardware map[string]interface{}
```
gauravgahlot, line 275: Is there any specific reason why we are using `map[string]interface{}` instead of `map[string]string`?
Because in this case the following will also work, while it is incorrect:
```
$ tink workflow create -t <template-id> -r '{"device_1": 12345}'
```

cli/tink/cmd/workflow/create.go
```diff
@@ -40,14 +35,14 @@ var createCmd = &cobra.Command{
 func addFlags() {
 	flags := createCmd.PersistentFlags()
 	flags.StringVarP(&template, "template", "t", "", "workflow template")
-	flags.StringVarP(&target, "target", "r", "", "workflow target")
+	flags.StringVarP(&hardware, "hardware", "r", "", "workflow target hardwares")
```
gauravgahlot, line 38: Replace `r` with `h`.

gauravgahlot, line 38: Change the description from "workflow target hardwares" to "workflow hardware" or "workflow targeted hardware".

cli/tink/cmd/workflow/create.go
```diff
@@ -40,14 +35,14 @@ var createCmd = &cobra.Command{
...
 
-	createCmd.MarkPersistentFlagRequired(fTarget)
+	createCmd.MarkPersistentFlagRequired(fHardware)
 	createCmd.MarkPersistentFlagRequired(fTemplate)
 }
 
 func createWorkflow(c *cobra.Command, args []string) {
-	req := workflow.CreateRequest{Template: template, Target: target}
+	req := workflow.CreateRequest{Template: template, Target: hardware}
```
gauravgahlot, line 45: Please update the proto, such that `Target` is `Hardware`.

deploy/db/docker-entrypoint-initdb.d/tinkerbell-init.sql
```diff
@@ -25,21 +25,10 @@ CREATE TABLE IF NOT EXISTS template (
...
-
-CREATE INDEX IF NOT EXISTS idx_rid ON targets (id);
-CREATE INDEX IF NOT EXISTS idx_rdeleted_at ON targets (deleted_at NULLS FIRST);
-CREATE INDEX IF NOT EXISTS idxgin_rtype ON targets USING GIN (data JSONB_PATH_OPS);
-
 CREATE TABLE IF NOT EXISTS workflow (
 	id UUID UNIQUE NOT NULL
 	, template UUID NOT NULL
-	, target UUID NOT NULL
+	, devices JSONB NOT NULL
```
gauravgahlot, line 31: I think it should be `hardware`.

grpc-server/workflow.go
```diff
@@ -267,35 +268,29 @@ func createYaml(ctx context.Context, sqlDB *sql.DB, temp string, tar string) (st
 	if err != nil {
 		return "", err
 	}
-	tarData, err := db.TargetsByID(ctx, sqlDB, tar)
-	if err != nil {
-		return "", err
-	}
-	return renderTemplate(string(tempData), []byte(tarData))
+	return renderTemplate(string(tempData), []byte(tar))
```
gauravgahlot, line 271: Please rename `tar` to something more meaningful.

grpc-server/workflow.go
```diff
@@ -267,35 +268,29 @@ func createYaml(ctx context.Context, sqlDB *sql.DB, temp string, tar string) (st
...
 }
 
 func renderTemplate(tempData string, tarData []byte) (string, error) {
```
gauravgahlot, line 274: Please rename `tarData` to something more meaningful.

test/framework/utils.go
```diff
@@ -12,12 +12,7 @@ var Log = logger
 
 // SetupWorkflow ... Set up workflow
 func SetupWorkflow(tar string, tmpl string) (string, error) {
-	//Add target machine mac/ip addr into targets table
-	targetID, err := CreateTargets(tar)
-	if err != nil {
-		return "", err
-	}
-	logger.Infoln("Target Created : ", targetID)
+	targetID := "c9d6faa4-08a2-4285-ae6c-f3401211bd56"
```
gauravgahlot, line 15: I think we should update the test framework according to the new structure.

grpc-server/workflow.go
gauravgahlot: I understand this will cover cases like:
```
{"device_one": {"mac": "macaddr"}, "device_two": "device-two-value"}
```

However, we must add a few validations around what is being provided as hardware value at the time of workflow creation.

parauliya: `interface{}` is being used as a value because in future there is a possibility there can be nested dirs in the input and you have to fetch the mac addr or IP addr from the nested key:value pair. In that case you will just need to change the template and you will be able to create the workflow without changing the code. But yes some validation must be required before placing them in the workflow.

cli/tink/cmd/workflow/create.go
parauliya: The -h flag is already reserve for `help` I guess because initally I used -h flag but it gave me the error of duplicate flags.

cli/tink/cmd/workflow/create.go
parauliya: I don't this it is required because `hardware` is also a kind of target so I have changed the name and type of `target` column in the `workflow` table to `devices` and `JSONB`.

deploy/db/docker-entrypoint-initdb.d/tinkerbell-init.sql
parauliya: In case of `multiple` devices in a `hardware` `devices` suits better even if a `hardware` has single `device`.

grpc-server/workflow.go
parauliya: replaced it with `devices`.

grpc-server/workflow.go
parauliya: replaced it with `devices`

test/framework/utils.go
parauliya: Yes, this will be done once we enable the test framework to run.

cli/tink/cmd/workflow/create.go
parauliya: I think `workflow targeted hardware` suits better so channged.

docs/writing-workflow.md
```diff
@@ -76,17 +65,17 @@ The following command creates a workflow template and returns a UUID:
...
+ $ tink workflow create -t edb80a56-b1f2-4502-abf9-17326324192b -r '{"device_1":"mac/IP"}'
 ```
 
 The above command returns a UUID for the workflow thus created. The workflow ID can be used for getting further details about a workflow. Please refer the [Tinkerbell CLI reference](cli/workflow.md) for the same.
 
-It's a good practice to verify that the targets have been well substituted in the template. In order to do so, use the following command:
+It's a good practice to verify that the worker have been well substituted in the template. In order to do so, use the following command:
 ```yaml
- $ tink workflow get edb80a56-b1f2-4502-abf9-17326324192b
+ $ tink workflow get <workflow Id returns from the above command>
```
gauravgahlot, line 78: Should be "workflow Id **returned** from the above command"

alexellis:
Does this mean that the target creation is no longer required?

PR #92, opened 2020-05-04
Reading hardware data from stdin

gauravgahlot:
At present, the hardware data is pushed as shown below and is not the right (or best) way to do it:
```
$ tink hardware push 'hadware-data-json'
```

With the code changes in place, a user can now pipe the hardware data from a JSON file. For example, if the data is saved in `/tmp/data.json`, you can push the data with either of the following:
```
$ cat /tmp/data.json | tink hardware push
$ tink hardware push --file /tmp/data.json
```
Also, if there is no error a confirmation message saying `Hardware data pushed successfully` is received after the data is pushed.

### DIFF
--- a/cli/tink/cmd/hardware/push.go
+++ b/cli/tink/cmd/hardware/push.go
@@ -3,7 +3,6 @@
 package hardware
 
 import (
-	"bufio"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -18,8 +17,8 @@ import (
 )
 
 var (
-	filePath string
-	fPath    = "file-path"
+	file  string
+	sFile = "file"
 )
 
 // pushCmd represents the push command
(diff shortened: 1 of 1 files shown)


nathangoulding:
I think having the capability to read the hardware data from a file is definitely advantageous, especially for large blobs and/or data that might write to a file asynchronously.

My suggestion is that we add the stdin as you've done it, but also have a `-f <file>` parameter to support the existing file-based mechanism.

gauravgahlot:
@nathangoulding That makes sense, I will do the changes. 
@mikemrm I didn't update the docs here because I'm of the opinion that [tinkerbell.org](https://tinkerbell.org/) should serve as the sole documentation resource. And, so we can remove all the duplicate documentation from Tink.

alexellis:
This looks great guys. Nice work on it. 👍

PR #96, opened 2020-05-05
added terraform into deploy dir

kdeng3849:
- Ported the terraform scripts from the `demo-v2` branch
- Specified in `input.tf` to use user api key
- ~~moved `setup.sh` into deploy directory~~


parauliya:
> * Ported the terraform scripts from the `demo-v2` branch
> * Specified in `input.tf` to use user api key
> * moved `setup.sh` into deploy directory

I think instead of moving setup.sh into deploy dir you can directly copy that file to the provisioner machine throught terraform script at a certain place and then you can login to provisioner and run that script.

kdeng3849:
@parauliya I actually think it's better to keep it separate. I remember there were sometimes problems with terraform not being able to ssh into the provisioner and then it just stops the entire operation. 
I also think it's better to align the steps for both setup with terraform and manual setup because it would get confusing if there's going to be "if you created the servers with terraform run the script here, otherwise the script is over there", etc.

parauliya:
> @parauliya I actually think it's better to keep it separate. I remember there were sometimes problems with terraform not being able to ssh into the provisioner and then it just stops the entire operation.
> I also think it's better to align the steps for both setup with terraform and manual setup because it would get confusing if there's going to be "if you created the servers with terraform run the script here, otherwise the script is over there", etc.

I agree with you on that but still I do not see a reason for moving the `setup.sh` into `deploy` folder else you have to do few changes in the setup.sh as well as @gauravgahlot mentioned in his comment.

kdeng3849:
> I agree with you on that but still I do not see a reason for moving the `setup.sh` into `deploy` folder else you have to do few changes in the setup.sh as well as @gauravgahlot mentioned in his comment.

The thought was to keep all things setup together and also to have the generated `envrc` file in the same directory as the `docker-compose.yml` for more convenient access after logging back into the provisioner machine. (I found myself having to do `source ../envrc` quite often and it does get a bit cumbersome after a few times)

kdeng3849:
@parauliya @gauravgahlot I will first move `setup.sh` back to outside deploy and then create a separate pr for that so as to not block the terraform scripts getting merged.

PR #107, opened 2020-05-13
Support for CentOS 8 for being a provisioner.

parauliya:
After this change we can also setup provisioner machine on which CentOS 8 is installed.


mrmrcoleman:
Let's please make sure there is an issue to either update or add docs for this @parauliya

PR #117, opened 2020-05-14
Move tink-server deploy/db files into top-level deploy dir

mmlb:
tink-server does not do anything with these files currently. These files
really do belong in the tink-server container image but we need better
migration handling -server before these file can come into it.


mmlb:
I split them like this since tink-server/Dockerfile was adding them. I thought tink-server had an entrypoint script that would run migrations. That isn't the case though.

deploy/docker-compose.yml
```diff
@@ -41,13 +41,14 @@ services:
...
-    build:
-      context: db
+    image: postgres:10-alpine
     restart: unless-stopped
     environment:
       POSTGRES_DB: tinkerbell
       POSTGRES_PASSWORD: tinkerbell
       POSTGRES_USER: tinkerbell
+    volumes:
+      - ./db/tinkerbell-init.sql:/etc/docker-entrypoint-initdb.d/:ro
```
iaguis, line 51: I think this should be
```suggestion
      - ./db/tinkerbell-init.sql:/docker-entrypoint-initdb.d/:ro
```

iaguis: Sorry, actually this:

```suggestion
      - ./db/tinkerbell-init.sql:/docker-entrypoint-initdb.d/tinkerbell-init.sql:ro
```

mmlb: right you are, I thought it docker-compose would to the "obvious" thing.

mmlb: fixed in latest push

alexellis:
Good to see a fix for this. When do we think this PR could get merged?

Looks like a few people (and now me, are running into this)

Screenshot 2020-05-18 at 14 07 11

deploy/docker-compose.yml
```diff
@@ -41,13 +41,14 @@ services:
...
-    build:
-      context: db
+    image: postgres:10-alpine
     restart: unless-stopped
     environment:
       POSTGRES_DB: tinkerbell
       POSTGRES_PASSWORD: tinkerbell
       POSTGRES_USER: tinkerbell
+    volumes:
+      - ./db/tinkerbell-init.sql:/docker-entrypoint-initdb.d/tinkerbell-init.sql:ro
```
alexellis, line 51: I'm not actually finding a local folder called `db` in the deploy folder after running setup.sh?

Screenshot 2020-05-18 at 14 10 48

alexellis:
If anyone else is landing here run:

```
docker exec -ti deploy_db_1 psql -U tinkerbell  
```

Then paste in from https://raw.githubusercontent.com/tinkerbell/tink/29d8cde68c019de76c40bc6ffaebf60209c2e5a8/deploy/db/tinkerbell-init.sql

parauliya:
Hi @gauravgahlot @mmlb , This PR needs to be merged asap, without this PR setup script will not work since `db` folder has been removed from the `deploy` folder in tink repo.

mmlb:
Yep looking for some review.

deploy/docker-compose.yml
mmlb: Because setup.sh fetches from master. I commented out the git fetch bit and used my own clone of this pr and it worked out fine.

mmlb:
I just ran through tf and setup off this branch and it all worked out. Can I get a :+1: @parauliya, @nathangoulding, @mikemrm

alexellis:
Thanks @mmlb 💪

PR #126, opened 2020-05-21
Fixed gofmt and golint issues and added the check in CI

parauliya:
This PR has the following things
    1. Fixed the goimports and golint issue in the current code.
    2. Add check for goimports and golint issues in the CI pipeline


cmd/tink-worker/worker.go
```diff
@@ -12,8 +12,8 @@ import (
 	"strings"
 	"time"
 
-	pb "github.com/tinkerbell/tink/protos/workflow"
 	"github.com/sirupsen/logrus"
+	pb "github.com/tinkerbell/tink/protos/workflow"
```
mmlb, line 17: please remove the empty line

client/main.go
```diff
@@ -68,6 +68,7 @@ func Setup() {
 	HardwareClient = hardware.NewHardwareServiceClient(conn)
 }
 
+// NewTinkerbellClient : Function to get a new hardware client
```
mmlb, line 71: Please change the format to be more conversational, for example:
```
NewTinkerbellClient creates a new hardware client
```

test/framework/worker.go
```diff
@@ -110,6 +110,7 @@ func captureLogs(ctx context.Context, cli *dc.Client, id string) {
 	fmt.Println("Logging Finished for container ", id)
 }
 
+// StartWorkers : function to start workers
```
mmlb, line 113: ditto

Makefile
```diff
@@ -18,3 +18,8 @@ run: ${binaries}
 test:
 	go clean -testcache
 	go test ./test -v
+
+verify:
+	./hack/verify-gofmt.sh
+	./hack/verify-golint.sh
+	./hack/verify-govet.sh
```
mmlb, line 25: I'd prefer we don't run these scripts in favor or just running the tools directly in the Makefile. Also, without the `find` bits either, that doesn't seem necessary when I ran the tools locally.

mmlb:
Also lets use `goimports` instead of `gofmt` since it does better handling of imports.

cmd/tink-worker/worker.go
parauliya: Removed.

client/main.go
parauliya: Done.

test/framework/worker.go
parauliya: Done

Makefile
parauliya: Yes, I agree, but to make it a part of CI we should have something we can fail when code is not properly formatted as per the `goimport` and `golint` fails.

parauliya: Done that.

mmlb:
@parauliya can you rebase instead of merge master in please? merge makes history messier.

mmlb:
Besides that :lgtm:

parauliya:
> @parauliya can you rebase instead of merge master in please? merge makes history messier.

@mmlb , I rebased my branch to upstream/master but this hisrotries are still there as I messed it up earlier, although the code changes in the files are only related to this fix.

PR #133, opened 2020-05-26
`tink template create` now accepts data from STDIN 

gauravgahlot:
At present, we create a template as shown below:
```
$ tink template create --name example --path /tmp/example.tmpl 
```

With the code changes in place, a user can now pipe the template data from a file. For example, if the data is saved in `/tmp/example.tmpl`, you can create a template with either of the following:
```
$ cat /tmp/example.tmpl | tink template create -n example
$ tink template create --name example --path /tmp/example.tmpl
```
In addition to the above two methods, you can also do something like below:
```
tink template create --name hello-world --path /dev/stdin <<EOF
version: '0.1'
global_timeout: 600
tasks:
  - name: hello world
    worker: {{.device_1}}
    actions:
      - name: hello-world
        image: hello-world
        timeout: 60
EOF
```


cmd/tink-cli/cmd/template/create.go
```diff
@@ -36,40 +57,55 @@ func addFlags() {
 	flags.StringVarP(&filePath, "path", "p", "", "path to the template file")
 	flags.StringVarP(&templateName, "name", "n", "", "unique name for the template (alphanumeric)")
 
-	createCmd.MarkPersistentFlagRequired(fPath)
+	// createCmd.MarkPersistentFlagRequired(fPath)
```
rgl, line 60: You should just remove this commented line as its not used anymore.

rgl:
>  --path /dev/stdin

You should remove this from the example that is on this PR description as there is no need to specify the path when it reads from stdin.

gauravgahlot:
> > --path /dev/stdin
> 
> You should remove this from the example that is on this PR description as there is no need to specify the path when it reads from stdin.

I would prefer to keep it to make sure that people know about this.

rgl:
> > > --path /dev/stdin
> > 
> > 
> > You should remove this from the example that is on this PR description as there is no need to specify the path when it reads from stdin.
> 
> I would prefer to keep it to make sure that people know about this.

If you want to be able to use `--path` with stdin you should also implement the de-facto-standard way of configuring the application to read from stdin or write to stdout by using `-` as a path. For example, in wget (and tar etc) you do `wget -O - example.com` to write to stdout.

This because I'm not sure if all unices have a `/dev/stdout` device, but if Tinkerbell is Linux only, we are good with that :-)

cmd/tink-cli/cmd/template/create.go
```diff
@@ -22,54 +22,88 @@ var (
...
 	if err != nil {
-		log.Fatal(err)
+		log.Println(err)
 	}
 	return data
 }
 
-func createTemplate(c *cobra.Command, args []string) {
-	req := template.WorkflowTemplate{Name: templateName, Data: readTemplateData()}
+func createTemplate(c *cobra.Command, data []byte) {
```
grahamc, line 84: While we're at it, I think we should reduce the globals used in this function and pass `templateName` as a parameter to the function to reduce spooky action. Could we do the same for `filePath` in `readTemplateData`?

cmd/tink-cli/cmd/template/create.go
```diff
@@ -22,54 +22,88 @@ var (
...
+	},
 	Run: func(c *cobra.Command, args []string) {
-		validateTemplate()
-		createTemplate(c, args)
+		var data []byte
+		if isInputFromPipe() {
+			data = readDataFromStdin()
+		} else {
+			data = readTemplateData()
+		}
```
grahamc, line 44: What if we reversed this logic a bit, and instead got an io reader:

```
path, _: c.Flags().GetString(fPath)


if path != "" {
  reader, err = os.Open(filePath)
  ...
} else {
  reader = os.Stdin
}
```

and then just `readAll(reader)`? As it stands, it will fairly eagerly read the input from stdin even if I've passed a file.

cmd/tink-cli/cmd/template/create.go
gauravgahlot: AFIK, since `templateName` and `filePath` are value receivers for the flags, they have to be global. But if there is a better to do that, please do let me know.

PR #140, opened 2020-05-28
Accept path to OSIE tar as env for tinkerbell setup

gauravgahlot:
As the `setup.sh` executes, it downloads the latest of Osie. This will not efficient for local setup with Vagrant, as you may have to destroy your provisioner VM for various reasons. 

The setup script will first read the `TB_OSIE_TAR` variable. If it is set, the setup will skip the download and extract the file provided `tar` instead. If not, it will go ahead and download the latest of Osie.

For your Vagrant setup, you can download the tar and then sync the containing directory with the provisioner VM. Now, set the environment variable to the synced folder. For instance:
```
// in Vagrantfile
config.vm.synced_folder '/var/osie', '/vagrant', type: 'rsync'

// export the env
export TB_OSIE_TAR=/vagrant/latest.tar.gz
```

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -296,8 +296,14 @@ setup_osie() {
 		mkdir -p "$osie_current"
     	mkdir -p "$tink_workflow"
     	pushd /tmp
-    	curl 'https://tinkerbell-oss.s3.amazonaws.com/osie-uploads/latest.tar.gz' -o osie.tar.gz
-    	tar -zxf osie.tar.gz
+
+        if [ -z "${TB_OSIE_TAR:-}" ]; then
+            curl 'https://tinkerbell-oss.s3.amazonaws.com/osie-uploads/latest.tar.gz' -o osie.tar.gz
+            tar -zxf osie.tar.gz
+        else
+            tar -zxf "$TB_OSIE_TAR"
+        fi
+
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -296,8 +296,14 @@ setup_osie() {
 		mkdir -p "$osie_current"
     	mkdir -p "$tink_workflow"
     	pushd /tmp
-    	curl 'https://tinkerbell-oss.s3.amazonaws.com/osie-uploads/latest.tar.gz' -o osie.tar.gz
-    	tar -zxf osie.tar.gz
+
+        if [ -z $TB_INTERFACE ]; then
```
grahamc, line 300:
```suggestion
        if [ -z "${TB_OSIE_TAR:-}" ]; then
```

the `${...:-}` says "if this variable is undefined, just make the result empty." Quoting it just in case the path to the tarball has spaces or other spooky characters in it.

Also, it looks like the indentation doesn't quite match. I just opened https://github.com/tinkerbell/tink/pull/143 so that if you your editor configured with editorconfig, it will just indent correctly automatically. Specifically, these lines of code are indented 2 spaces too far, and then the stuff inside the `if` and `else` should be indented with a tab.

PR #141, opened 2020-05-29
Network configuration using netplan for Ubuntu versions greater than 16.04

gauravgahlot:

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -232,19 +232,66 @@ setup_networking() {
 	if ! is_network_configured; then
 		cidr=$(echo $TINKERBELL_NETWORK | grep -Eo "\/[[:digit:]]+" | grep -v "^$" | tr -d "/")
 		case "$1" in
-		ubuntu)
-			if [ ! -f /etc/network/interfaces ]; then
-				echo "$ERR file /etc/network/interfaces not found"
-				exit 1
-			fi
+			ubuntu)
+				v16=16.04
+				version=$(. /etc/os-release && echo "$VERSION_ID")
+				if (( $(echo "$version > $v16" | bc -l) )); then
+					echo "$INFO found Ubuntu $version, using 'netplan' for network configuration"
 
-			if grep -q $TINKERBELL_HOST_IP /etc/network/interfaces; then
-				echo "$INFO tinkerbell network interface is already configured"
-			else
-				# plumb IP and restart to tinkerbell network interface
-				if grep -q $TINKERBELL_NETWORK_INTERFACE /etc/network/interfaces; then
-					echo "" >>/etc/network/interfaces
-					write_iface_config
+					# this is required to fix the bug described here
+					# https://bugs.launchpad.net/netplan/+bug/1810043
+					apt-get install -y --no-install-recommends netplan.io >> /dev/null
+					cat >/etc/netplan/${TINKERBELL_NETWORK_INTERFACE}.yaml <<EOF
+---
+network:
+  version: 2
+  renderer: networkd
+  ethernets:
+    $TINKERBELL_NETWORK_INTERFACE:
+      addresses:
+        - $TINKERBELL_HOST_IP/$cidr
+        - $TINKERBELL_NGINX_IP/$cidr
+EOF
+					ip link set $TINKERBELL_NETWORK_INTERFACE nomaster
+					netplan apply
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -232,27 +232,60 @@ setup_networking() {
...
+  renderer: networkd
+  ethernets:
+    $TINKERBELL_NETWORK_INTERFACE:
+      addresses:
+        - $TINKERBELL_HOST_IP/$cidr
+        - $TINKERBELL_NGINX_IP/$cidr
+EOF
+					ip link set $TINKERBELL_NETWORK_INTERFACE nomaster
+					netplan apply
+					echo "$INFO waiting for the network configuration to be applied by systemd-networkd"
```
grahamc, line 256: I think there is probably a systemd service we could wait on with something like:

```
while ! systemctl is-active netplan-....; do
  sleep 1
done
```

or perhaps some other unit. This would ensure we don't move past this step if it failed.

setup.sh
```diff
@@ -232,27 +232,60 @@ setup_networking() {
...
+---
+network:
+  version: 2
+  renderer: networkd
+  ethernets:
+    $TINKERBELL_NETWORK_INTERFACE:
+      addresses:
+        - $TINKERBELL_HOST_IP/$cidr
+        - $TINKERBELL_NGINX_IP/$cidr
+EOF
```
grahamc, line 253: Generating YAML on the fly with bash feels a bit scary. In particular, magic values can turn in to other things like "no" being a bool, and `1.10` being just `1.1`. What would you think about using (and installing if it isn't yet available) `jq` for this? Using jq guarantees the output will be correct and that we didn't make an error in indentation, whitespace, and also makes sure that all of the values are properly quoted.

Here is an example of what that would look like:

```bash

TINKERBELL_NETWORK_INTERFACE=10.5.3.1
TINKERBELL_HOST_IP=10.5.3.2
TINKERBELL_NGINX_IP=10.5.3.3
cidr=24

jq -n \
   --arg interface "$TINKERBELL_NETWORK_INTERFACE" \
   --arg cidr "$cidr" \
   --arg host_ip "$TINKERBELL_HOST_IP" \
   --arg nginx_ip "$TINKERBELL_NGINX_IP" \
   '{
  network: "hi",
  renderer: "networkd",
  ethernets: {
    ($interface): {
      addresses: [
        "\($host_ip)/\($cidr)",
        "\($nginx_ip)/\($cidr)"
      ]
    }
  }
}'
```

This prints:

```
{
  "network": "hi",
  "renderer": "networkd",
  "ethernets": {
    "10.5.3.1": {
      "addresses": [
        "10.5.3.2/24",
        "10.5.3.3/24"
      ]
    }
  }
}
```

since JSON is valid YAML, this output works directly with netplan.

setup.sh
```diff
@@ -232,27 +232,60 @@ setup_networking() {
...
 		case "$1" in 
 			ubuntu)
-				if [ ! -f /etc/network/interfaces ] ; then
-					echo "$ERR file /etc/network/interfaces not found"
-					exit 1
-				fi
+				v16=16.04
+				version=$(. /etc/os-release && echo "$VERSION_ID")
+				if (( $(echo "$version > $v16" | bc -l) )); then
+					echo "$INFO found Ubuntu $version, using 'netplan' for network configuration"
```
grahamc, line 238: It looks like 17.10 is the first release to use netplan by default. Maybe we should change this to `>= 17.10`?

setup.sh
```diff
@@ -232,27 +232,60 @@ setup_networking() {
...
+  ethernets:
+    $TINKERBELL_NETWORK_INTERFACE:
+      addresses:
+        - $TINKERBELL_HOST_IP/$cidr
+        - $TINKERBELL_NGINX_IP/$cidr
+EOF
+					ip link set $TINKERBELL_NETWORK_INTERFACE nomaster
+					netplan apply
+					echo "$INFO waiting for the network configuration to be applied by systemd-networkd"
+					sleep 3
```
grahamc, line 257: I wonder if netplan creates a systemd unit we could wait for, like this:
 
```sh
while ! systemctl is-active netplan-something-or-other; do
  sleep 1
done
```

grahamc:
Oops, I meant to also say -- looks pretty good!

grahamc:
These commits were included in #147

PR #154, opened 2020-06-04
Codespell fixes

mmlb:
Fixes spelling mistakes and add codespell to ci so won't happen again.

### DIFF
--- a/db/workflow.go
+++ b/db/workflow.go
@@ -111,7 +111,7 @@ func insertIntoWfWorkerTable(ctx context.Context, db *sql.DB, wfID uuid.UUID, wo
 	INSERT INTO
 		workflow_worker_map (workflow_id, worker_id)
 	SELECT $1, $2
-	WHERE 
+	WHERE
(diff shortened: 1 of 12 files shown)


db/workflow.go
```diff
@@ -573,16 +573,16 @@ func GetWorkflowContexts(ctx context.Context, db *sql.DB, wfID string) (*pb.Work
...
 	if err == nil {
 		return &pb.WorkflowContext{
 			WorkflowId:           wfID,
 			CurrentWorker:        cw,
 			CurrentTask:          ct,
 			CurrentAction:        ca,
 			CurrentActionIndex:   cai,
-			CurrentActionState:   cas,
+			CurrentActionState:   case,
 			TotalNumberOfActions: tact}, nil
```
grahamc, line 586: oops this cas -> case is a mistake.

mmlb: ah right

mmlb: fixed, ptal

PR #167, opened 2020-06-08
Pull nixos/nix image always

mmlb:
Seeing some current TLS cert issues, trying this to see if it helps.


grahamc:
I think this is sorted, closing.

PR #170, opened 2020-06-15
docker-compose: remove EFK

grahamc:
docker-compose: remove EFK

Fetching EFK adds several minutes to the setup time and start-times.

I think most hackers will want to just see the logs on the console,
and be interested in EFK once things are more prod-like. Using EFK
means the user actually has to do some initial config on their
cluster, and it doesn't just work out of the box.

One problem is OSIE wants to talk specifically to ElasticSearch, but
I wonder if we could swap it out with syslog and a syslog listener?


grahamc:
Unfortunately we have this to reckon with: https://github.com/tinkerbell/osie/blob/master/installer/workflow-helper.sh#L60-L62

grahamc:
I got a general :+1: from @mmlb. We should probably adjust the OSIE workflow helper to stop using fluentbit, and just use syslog like ipxe does:

https://github.com/tinkerbell/boots/blob/master/ipxe/dhcp_options.go#L68

and like the entrypoint of osie already does:

```
[grahamc@Petunia:~/projects/github.com/tinkerbell]$ cat ./osie/docker/entrypoint.sh
...
echo "Logging to $RLOGHOST with $container_uuid"
# we tee to fd3 so that output can go out on stdout instead stderr
exec 3>&1
exec 2> >(tee /proc/self/fd/3 | logger -n "$RLOGHOST" -P 514 -t "$container_uuid")
exec 1>&2
```

grahamc:
fixup'd, @mmlb !

alexellis:
( The description still says EFK instead of ELK )

alexellis:
Have you made this an option or removed it completely as I consumed the logs API of elastic search in tinkerbot for the /logs command

alexellis:
One thing to bear in mind is that log collection is hard when working on PCs in the lab where there is no serial output. The only way I was able to get output was by searching in ElasticSearch when doing this on a regular PC. Is this PR taking that away?

PR #175, opened 2020-06-17
ci-checks: run all the checks on checked-in files; check shellcheck

grahamc:
Don't run on files which are not being committed, and run all
the checks even if one step failed.

This PR should fail (validating that the CI check still fails) and then I will add additional commits fixing its complaints.

### DIFF
--- a/test/push_images.sh
+++ b/test/push_images.sh
@@ -1,10 +1,6 @@
 #!/bin/bash
 
-for i in {1..10}; do
-	docker login -u username -p password localhost
-	if [ $? -eq 0 ]; then
-		break
-	fi
+while ! docker login -u username -p password localhost; do
 	sleep 1
 done
 docker push localhost/update-data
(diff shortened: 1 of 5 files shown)


grahamc:
@mmlb do you think it is worth adding `git` to the shell.nix for this, or should we go back to using wildcards? I wanted to use `git ls-files` so that untracked files in a checkout don't gum up the works.

mmlb:
I'm fine with adding git to shell.nix. The shell.nix's aren't super pure for the projects and I've kind of wanted to move towards that way as much as possible where it makes sense.

test/push_images.sh
```diff
@@ -1,10 +1,6 @@
 #!/bin/bash
 
-for i in {1..10}; do
-	docker login -u username -p password localhost
-	if [ $? -eq 0 ]; then
-		break
-	fi
+while ! docker login -u username -p password localhost; do
```
mmlb, line 3: Does this really belong in this PR? I'm fine if you want to earmark it in, just wand to double check.

grahamc: Maybe not, it just seemed like the easiest fix for `$i` being unused. I could swap it out to just echo the iteration. What do you think?

PR #177, opened 2020-06-18
Fixed Issue 80: Modified the test framework for tink repo

parauliya:
This framework was written a while ago but is commented and there have been a lot of changes since then.

1. Remove the target concepts completely
2. Changed the way of writting templates
3. Changed the way of creating workflows
4. Generating certs are not the part of docker-compose anymore

TODO:
This framework works well for single worker. There are few changes required in the framework to make it work for multiple workers


gianarb:
I would like to help you to move this PR forward, I think it is valuable but it needs documentation (in the code itself). As a developer, I need to open https://pkg.go.dev/github.com/tinkerbell/tink@v0.0.0-20200714130438-0100e535bd94/test/framework?tab=doc and I need to be able to figure out how and when to use this framework quickly.

gianarb:
This conversation with @mmlb happened in the public Tinkerbell channel, so I am reporting it here for visibility @mrmrcoleman . From my understanding, we can move from drone to something else that matches our needs (docker), probably GitHub Action that is free for OSS project or Travis. (I vote for actions).

Screenshot 2020-07-20 at 16 20 15

gianarb:
btw I moved this to blocked because I think we have to change CI provider first

mmlb:
We can always have 2 CI for the time being.

gianarb:
@parauliya can you modify this issue in order to run as GitHub action? I think @mmlb is right and we can have a e2e pipeline that runs on its own.

ideally, it will be nice to have one label based on this one #224 . if you can prepare a single command that runs the e2e tests I can help you hook it up as Gh action.

gianarb:
We are moving to #277 , this should not be required anymore

PR #178, opened 2020-06-18
Pinned registry to 2.6

mrmrcoleman:
It looks like htpasswd was removed from the upstream registry image. Pinned to 2.7 and added apache2-utils.

### DIFF
--- a/deploy/registry/Dockerfile
+++ b/deploy/registry/Dockerfile
@@ -1,5 +1,5 @@
 FROM registry:2
-RUN apk add --no-cache --update curl
+RUN apk add --no-cache --update curl apache2-utils
 ARG REGISTRY_USERNAME
 ARG REGISTRY_PASSWORD
 RUN mkdir -p /certs /auth
--- a/deploy/registry/Dockerfile
+++ b/deploy/registry/Dockerfile
@@ -1,5 +1,5 @@
-FROM registry:2
-RUN apk add --no-cache --update curl
+FROM registry:2.7
+RUN apk add --no-cache --update curl apache2-utils
 ARG REGISTRY_USERNAME
 ARG REGISTRY_PASSWORD
 RUN mkdir -p /certs /auth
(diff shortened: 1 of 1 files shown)


mmlb:
@mrmrcoleman can you split into 2 commits please? I'd prefer first for the apache-utils with a message that it went away in 2.7. And the second pinning the image to 2.7 with a message along the lines of so it doesn't happen again ;).

mrmrcoleman:
@mmlb It seems 2.7 doesn't work, 2.6 does. Want me to pin it there without the apk add?

mrmrcoleman:
@mmlb Updated to show the steps in between.

alexellis:
Since we are deriving a new custom image anyway, why don’t we use the later image and just add it back in?

alexellis:
Adding this step to your custom Dockerfile would save your butt for when someone comes along and innocently bumps the version back up to 2.7.

```
/ # apk add --update apache2-utils
(1/5) Installing libuuid (2.34-r1)
(2/5) Installing apr (1.7.0-r0)
(3/5) Installing expat (2.2.9-r1)
(4/5) Installing apr-util (1.6.1-r6)
(5/5) Installing apache2-utils (2.4.43-r0)
Executing busybox-1.31.1-r9.trigger
OK: 7 MiB in 20 packages
/ # htpasswd
Usage:
	htpasswd [-cimBdpsDv] [-C cost] passwordfile username
	htpasswd -b[cmBdpsDv] [-C cost] passwordfile username password
```

deploy/registry/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry:2
+FROM registry:2.6.2
 RUN apk add --no-cache --update curl
```
alexellis, line 2: A solution that would last the tests of time would be the following:

```
RUN apk add --no-cache --update curl apache2-utils
```

Instead of pinning to an earlier version.

mrmrcoleman:
I'm fine with either approach. Perhaps I misunderstood what @mmlb wanted to see?

gianarb:
As we discussed during a chat this is blocking set up and people are asking all over what is going on.
A better fix can land as a separate PR!

Issue #181, opened 2020-06-22
Vagrant: Worker needs more RAM to pull images


alexellis:
When using Vagrant, the worker needs more RAM than 500MB since it uses a tmpfs / Ramdisk to pull images, we saw this fail for https://github.com/alexellis/tink-workflow which has workflow actions of around 429mb.

I would suggest 4GB would be a sensible default

The primary hard disk could also be too small - I would suggest using the Ubuntu base image for the worker, so that it has at least 32GB.

alexellis:
@mrmrcoleman @grahamc this is urgent for a successful webinar, but I think we should use Packet due to the issues faced when Vagrant crashed my Mac and rebooted it. After coming up again, Tinkerbell was left in an unrecoverable state.

<img width="2202" alt="Screenshot 2020-06-23 at 15 24 09" src="https://user-images.githubusercontent.com/6358735/85416264-288f6800-b566-11ea-9d54-6875e38c560b.png">

<img width="1836" alt="Screenshot 2020-06-23 at 15 25 50" src="https://user-images.githubusercontent.com/6358735/85416270-2c22ef00-b566-11ea-8dcc-69db6e66d3bc.png">

mrmrcoleman:
@alexellis Working on this now.

mrmrcoleman:
https://github.com/tinkerbell/tink/pull/188 feedback please.

gauravgahlot:
@mrmrcoleman @alexellis Can we close this issue?

alexellis:
It seems like it, since @mrmrcoleman sent a PR. I would suggest that the team try out the "closing keywords"

Just add to your commit message:

```
Closes #181
```

The word "Fixes" also works the same, then, when the PR is merged, the issue will also be closed, subject to some constraints.

PR #188, opened 2020-06-23
Added 4G memory for worker to prevent OOM errors when booting OSIE

mrmrcoleman:
Intended to address https://github.com/tinkerbell/tink/issues/181 by simply upping the amount of memory on the worker.


alexellis:
@gianarb do you have merge access?

PR #190, opened 2020-06-24
Update tutorial's final step

alexellis:
Once bootstrapped, we need to be able to shell into the
provisioner to run the commands. We also should have something
that can be copy/pasted without thinking or unpicking.

Signed-off-by: Alex Ellis (OpenFaaS Ltd) <alexellis2@gmail.com>

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -480,7 +480,7 @@ check_prerequisites() (
 )
 
 whats_next() (
-	echo "$NEXT  1. Enter /vagrant/deploy and run: source ../envrc; docker-compose up -d"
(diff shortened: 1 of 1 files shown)


alexellis:
@gianarb @mrmrcoleman 
PTAL

gianarb:
I presume the main issue here is that the `setup.sh` is used also for other setups procedure like the one on Packet right?
This PR won't work for both

alexellis:
We already have a hard coded reference to vagrant in there. This just makes the command runnable rather than something you have to parse mentally

gianarb:
> We already have a hard coded reference to vagrant in there

I think you need to help me to figure out where those references are. @grahamc is the setup.sh written to be provider agnostic?

mrmrcoleman:
We will fix this by removing the reference from both the setup.sh and the docs.

PR #197, opened 2020-06-29
ENG-6658 user defined metadata

kdeng3849:
## Description

- Changes the metadata field in the hardware proto to be a string instead of a defined structure.
- Changes the REST api endpoints and handlers to be manually defined rather than using the ones in the generated gateway files.

Related PRs:
boots: https://github.com/tinkerbell/boots/pull/58
hegel: https://github.com/tinkerbell/hegel/pull/16

## Why is this needed
- Metadata is meant to be a free-for-all space where users can input any data not directly relevant to tink but still used in other parts of the system. As such, it should not have a strictly defined structure.
- The handlers generated from the proto file works fine, but doesn't allow for much control over how the request body gets parsed and how the response is formatted. As a result, feels very clunky.

Fixes: https://github.com/tinkerbell/tink/issues/199

## How Has This Been Tested?

- Manually using a tinkerbell/packet setup
- Unit tests for http endpoints

Users shouldn't notice a difference in pushing/getting hardware data as compared to before, except that now, any and all fields they've pushed under metadata will actually be stored. No migration needed.

For the push data for the hardware push rest endpoint, there is now no need to wrap the hardware data within `{"data": {}}`. For usage examples, refer to this [gist](https://gist.github.com/kdeng3849/11df3ed5d6ad4a003f9047386afbfcad).

## Checklist:

I have:

- [x] added unit or e2e tests


http-server/http_handlers.go
```diff
@@ -0,0 +1,558 @@
...
+	if _, err := tmpl.Parse(data); err != nil {
+		return err
+	}
+	return nil
+}
+
+// writeResponse appends a new line after res
+func writeResponse(w http.ResponseWriter, status int, res string) {
+	w.WriteHeader(status)
+	if _, err := w.Write([]byte(fmt.Sprintln(res))); err != nil {
```
kdeng3849, line 555: Should I add a new line at the end of a response or just leave `res` as is?

Issue #204, opened 2020-07-03
Add Validation of hardware data before pushing it to avoid duplication of mac addresses in the database


parauliya:
## Expected Behaviour
If someone tries to push the hardware data which contains a mac address which is already present in the database, the API will fail with the error explaining that "this mac address is already present in the database.

## Current Behaviour
Currently there is no validation for this so a if user tries to push a hardware data which contains a mac which is already present in the data base, a new hardware data entry will be created with a mac address which is already present in the database which will lead to have a duplication of MAC addresses in the database. After this validation that hardware data will not be allowed to push which has a mac address which is already present in the database.

## Possible Solution
Add a validation before pushing the hardware data where you can check for all the MAC address present in the hardware data that if there is any entry for any of the mac address of hardware data and if there is an enrtry found this hardware data is not allowed to be pushed in the database.

## Your Environment
* Operating System and version (Linux, Windows, MacOS).
   Each and every OS
* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:
   In any of the above environment this issue will occur.

PR #205, opened 2020-07-03
Fix issue #204: Added validation to avoid duplication of mac addresses in the hardware database

parauliya:
## Description

If someone tries to push the hardware data which contains a mac address which is already present in the database, the API will fail with the error explaining that "this mac address is already present in the database.

## Why is this needed

Fixes: [https://github.com/tinkerbell/tink/issues/204](https://github.com/tinkerbell/tink/issues/204)

## How Has This Been Tested?
I have tried to push the hardware data with different id but same mac address which is already present in the database and it did't allow me to push the hardware data by saying that that particular mac address is present in the database

The exising user will not be able to push the hardware data which contains the mac address which is already present in the database.
No Migration steps/scripts needed.

### DIFF
--- a/grpc-server/hardware.go
+++ b/grpc-server/hardware.go
@@ -32,8 +32,11 @@ func (s *server) Push(ctx context.Context, in *hardware.PushRequest) (*hardware.
 		return &hardware.Empty{}, err
 	}
 
-	// TODO: somewhere here validate json (if ip addr contains cidr, etc.)
-
+	// TODO: Add more validation for hardware data in the definition of below function if required
+	err := s.validateHardwareData(ctx, hw)
+	if err != nil {
+		return &hardware.Empty{}, err
+	}
 	logger.With("id", hw.Id).Info("data pushed")
 
 	var fn func() error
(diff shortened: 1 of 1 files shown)


grpc-server/hardware.go
```diff
@@ -285,3 +288,24 @@ func (s *server) Delete(ctx context.Context, in *hardware.DeleteRequest) (*hardw
...
+			newhw := hardware.Hardware{}
+			err := json.Unmarshal([]byte(data), &newhw)
+			if err != nil {
+				logger.Error(err, "Failed to unmarshal hardware data")
+				return err
+			}
+			if newhw.Id == hw.Id {
+				return nil
+			}
+			return errors.New("conflicting hardware MAC address " + interfaces[i].GetDhcp().GetMac() + " provided with hardware data/info")
```
gauravgahlot, line 307: Please move all the error messages (across the file) to the top as constants. This is a practice we should follow going forward; no more inline messages. 
Eg:
```
const (
    conflictingMAC = "conflicting hardware MAC address %v provided with hardware data/info"
)

// you can now use it as 
return errors.New(fmt.Errorf(conflictingMAC, interfaces[i].GetDhcp().GetMac()))
```

grpc-server/hardware.go
```diff
@@ -32,8 +32,11 @@ func (s *server) Push(ctx context.Context, in *hardware.PushRequest) (*hardware.
...
 	}
 
-	// TODO: somewhere here validate json (if ip addr contains cidr, etc.)
-
+	// TODO: Add more validation for hardware data in the definition of below function if required
+	err := s.validateHardwareData(ctx, hw)
+	if err != nil {
+		return &hardware.Empty{}, err
+	}
 	logger.With("id", hw.Id).Info("data pushed")
```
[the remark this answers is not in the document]
parauliya: I actually changed it i thin forgot to push.

grpc-server/hardware.go
parauliya: To move all the error messages across the file is a bit tidy at this moment. And also we need to do this for all the files so may that can be done in a seperate issue but for my changes I will follow this practise.

Issue #196, opened 2020-06-29
Improved logic for worker connections


nathangoulding:
## Description

There are a couple issues manifesting themselves which need resolution:

- Workers are connecting to tink and attempting to sleep for 3 seconds between "do I have more work" checks. The sleep is buggy however, and results in thousands of lines of log messages being dumped due to it continually retrying. Solution: switch to a gRPC stream instead of `sleep 3`.
- When fetching workflows, failed and completed workflows are being returned; tink should only return workflows that are in progress or pending.

## Why is this needed

- Incorrect sleep leading to log messages crashing user's terminal sessions, overflowing RAM / disk, etc.

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests
- [ ] provided instructions on how to upgrade

parauliya:
Hi @nathangoulding, 
There are few things I would like to discuss about :
 1. I think adding gRPC stream between tink-worker and tink-server is not a good solution because of the following reasons
      

               -   This will convert the protocol between worker and server from stateless to stateful which will add the condition based on stream connectivity.
               -   If there are multiple workers then server has to maintain sessions for each worker.
               -   This will complicate things between tink-worker and tink-server.
               -   Performance wise stream does not give any benefit.

 2.  For this issue we can do few optimistion and fix the sleep so that terminal should not filled up with too many logs.

 3. When fetching workflow tink should only return workflows that are in `progress` or `pending`  state and that too for that particular worker. 
         **This change is very much required at the server side.**

4. There is one more thing which is not there in this issue but we are planning for it which is as follows:
        As per the current implementation the `tink-worker container` on the` worker machine` exits as soon as it executes all the tasks assigned to it but we would like that `container` to keep running and polling to the server if there is a new `workflow` created for it so that it should start executing that wrokflow automatically. I would like to raise an issue for this enhancement and then do it after the fix of this issue.  
           
Please let me know your views on the same.

nathangoulding:
Responses inline:

> This will convert the protocol between worker and server from stateless to stateful which will add the condition based on stream connectivity.

If there is an established session it will use it; otherwise, it will wait for a worker to fetch it like it does today.

> If there are multiple workers then server has to maintain sessions for each worker.

Correct, this is how it works with osie-runner and hegel in production today.

> This will complicate things between tink-worker and tink-server.

I'm not sure I understand this justification.

> Performance wise stream does not give any benefit.

[...]

Issue #72, opened 2020-04-22
Worker should not exit once all workflows are finished


invidian:
Currently worker polls for new workflows while some other workflows are running, but when they are all finished, it exists. That seem counter-intuitive and requires from user to reboot the machine to run more workflows, which may not be even possible, if one workflow already installed the OS and then machine will boot from the disk.

It seems worker should wait infinitely for new workflows. See also #71

PR #216, opened 2020-07-13
Fix Issue #196: Filtered the workflow context sent to worker by server

parauliya:
## Description
There are following change in this PR:
1. Filtered the workflow context sent to worker by server on the basis of status and worker_id.
2. Changed Info logs to Debug during sleep in worker side
3. Changed a bit of logic in worker for not repeating the same workflow execution again.

Following things are fixed in second commit of this PR
1. Modified the filter from the server side while sending the workflow context
2. Added a gRPC stream for getting the workflow context
3. Continue the loop for worker even after completing the workflow so that it can fetch any new workflow.

## Why is this needed

This resolves the following issue.
Fixes: #196 #222 #72 

## How Has This Been Tested?
This has been tested by creating a sample worklfow and execute it.
Following scenarios was executed for the testing:
1. One workflow has one worker
2. One workflow has two workers
3. Two workflows and each workflow has two workers

It will enhance the user experience as logging will be lesser as compared to the current one on worker side.
There is no migration script needed.


executor/executor.go
```diff
@@ -160,3 +163,31 @@ func GetWorkflowDataVersion(context context.Context, workflowID string, sdb *sql
 	}
 	return &pb.GetWorkflowDataResponse{Version: version}, nil
 }
+
+func isApplicableToSend(context context.Context, wfContext *pb.WorkflowContext, workerID string, sdb *sql.DB) bool {
```
gianarb, line 167: It will be amazing to get in a written format when a worker can receive workload (I think that's what this function does). 

This function also requires a test case with a few cases (table testing please).

parauliya: Sure, I can write a comment over it. And regarding the test cases, we have not written unit test or Table test for any of the  function. We are planning to write all those tests at once by creating a new issue in which we will list out the functions and then we can track the progress over that.

gianarb: Are we in a rush with this? otherwise tested and documented atomic unit of works are awesome!

parauliya: Yes, I agree with you of testing the unit of work are necessary. But we are surely in a bit of rush for this one. Since this is required in master branch as soon as possible. I will create another issue which will also be related to the #196 . I will try to cover that in to the fix of that issue.

db/workflow.go
```diff
@@ -345,7 +345,7 @@ func GetWorkflowDataVersion(ctx context.Context, db *sql.DB, workflowID string)
 }
 
 // GetfromWfWorkflowTable : gives you the current workflow
-func GetfromWfWorkflowTable(ctx context.Context, db *sql.DB, id string) ([]string, error) {
+func GetfromWfWorkflowTable(db *sql.DB, id string) ([]string, error) {
```
gianarb, line 348: I am not sure this function name describes what it does. I understand it was already like that but can we evaluate the idea to change it to something different? Like `SelectWorkflowByID`?

Kind of like the idea to use SQL like keywords to describe functions that interact with SQL database 🍬

parauliya: This function does not used by any external user. I will update the comment though.

gianarb: Even internal users need clarity :)

executor/executor.go
```diff
@@ -160,3 +158,36 @@ func GetWorkflowDataVersion(context context.Context, workflowID string, sdb *sql
 	}
 	return &pb.GetWorkflowDataResponse{Version: version}, nil
 }
+
+// The below function check whether a particular workflow context is applicable or needed to
+// be send to a worker based on the state of the current action and the targeted workerID.
+func isApplicableToSend(context context.Context, wfContext *pb.WorkflowContext, workerID string, sdb *sql.DB) bool {
+	if wfContext.GetCurrentActionState() != pb.ActionState_ACTION_FAILED ||
+		wfContext.GetCurrentActionState() != pb.ActionState_ACTION_TIMEOUT {
```
nathangoulding, line 166: Let's flip this logic around and return early to reduce indentation.

executor/executor.go
nathangoulding: I think this is one of those methods that definitely needs 100% test coverage

protos/workflow/workflow.proto
```diff
@@ -39,7 +40,7 @@ service WorkflowSvc {
             get: "/v1/workflows/{id}/events"
         };
     };
-    rpc GetWorkflowContexts(WorkflowContextRequest) returns (WorkflowContextList) {}
+    rpc GetWorkflowContexts(WorkflowContextRequest) returns (stream WorkflowContext) {}
```
nathangoulding, line 43: Is changing to a stream backwards compatible?

executor/executor.go
parauliya: sure

protos/workflow/workflow.proto
parauliya: No, I don't think so, but Since this function is only called from worker side so I have changed this only instead of having a new gRPC API. Is that ok?

parauliya:
> ker/strea
This handles both of your cases as follows:
1. When the worker do the iPXE boot for the first time, it first executes all the existing workflow which are targeted to it and then it will keep on running on that worker machine. And If you create a new workflow which is targeted to this worker then that workflow will be executed automatically.
2. worker will first execute all the workflows which were created before it was started and then it will go for the new workflows which are targated to it.

parauliya:
> We still need to handle the scenarios where we have an already connected worker/stream, and 1) a new workflow gets created that targets that worker, or 2) an already executing workflow reaches a step where the worker gets new work. How is this achieved in this situation?

This handles both of your cases as follows:

When the worker do the iPXE boot for the first time, it first executes all the existing workflow which are targeted to it and then it will keep on running on that worker machine. And If you create a new workflow which is targeted to this worker then that workflow will be executed automatically.
worker will first execute all the workflows which were created before it was started and then it will go for the new workflows which are targated to it.

nathangoulding:
> This handles both of your cases as follows:
> 
> 1. When the worker do the iPXE boot for the first time, it first executes all the existing workflow which are targeted to it and then it will keep on running on that worker machine. And If you create a new workflow which is targeted to this worker then that workflow will be executed automatically.
> 2. Worker will first execute all the workflows which were created before it was started and then it will go for the new workflows which are targated to it.

You're right. When we add the ability for workers to fetch-and-wait instead of fetch-and-quit, as addressed in issue #222, we will need to update this so that the server sends work to the workers that are already connected and waiting.

Issue #169, opened 2020-06-11
Smoke or E2E tests


gauravgahlot:
The idea is to have smoke or E2E tests that:
 - setup a provisioner and a worker with Vagrant
 - execute a workflow
 - be part of the CI pipeline

Initial tests may cover the following workflows:
 - hello-world
   - services covered - tink, boots, osie
 - install Alpine Linux on worker
   - services covered - tink, boots, osie, hegel, cacher

gauravgahlot:
@mrmrcoleman @grahamc 
Please update the description if I've missed anything from our discussion.

mrmrcoleman:
Hey @gauravgahlot it would also be good to consider which events should trigger the CI. The end to end covers all (?) of the services, so does it run whenever any of the repos change?

mrmrcoleman:
@gauravgahlot Which example are you going to use for the alpine work? (install Alpine Linux on worker)

gianarb:
@gauravgahlot did you already select a target platform for CI/CD? I doubt github action, circleci, travis allowed to use vagrant.
Maybe @grahamc knows more because he tweeted about this

mmlb:
our internal drone.packet.net exists *solely* to run VMs (for osie tests) :D, we can use it.

mmlb:
But its an older version of drone. We can fire up a separate setup off of latest drone.io version and that way also keep external stuff off of our internal build hosts.

I can't recall if semaphore or circle ci let us bring our own build agents and thus allow us to run VMs in them.

gianarb:
@mmlb do you think it is useful to setup e2e tests in a way that they can support multiple platform? Right now we have

1. Vagrant
2. Packet

Tomorrow we will may have something else and the workflow I am picturing is the one we have for kubernetes.

1. PR comes in somebody review it and when it is ready to get tests e2e it adds the label
2. When the label is there one job per platform starts and reports its status
3. You can retest with a comment like:

```
/retest vagrant
/retest packet
```

Ideally, this does not even need a runner that supports VMs because we will reach packet-api and use a device (or more based on the tests itself) for every build.

Does it have a sense? Or is it too much? 😄

mrmrcoleman:
My 2 cents.

For Vagrant we'd want to test it in the way the user would use it. This means we'll need a build runner that supports virtualization.

I think testing the actual workflows might be tricky, but happy to think through it with you.

gianarb:
[...]

PR #224, opened 2020-07-20
[WIP] e2e test for vagrant setup

gianarb:
## Description

This PR uses vagrant to spin up VMs and it execute automatically the [Vagrant Setup](https://tinkerbell.org/setup/local-with-vagrant/)

## Why is this needed

We would like to get automatically tested the Vagrant Setup guide because it has to work in order to offer a great first impression to the community.

Fixes: #169 

## How Has This Been Tested?

This code tests things by design 😎

No impact

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/.github/workflows/vagrant-packet.yaml
+++ b/.github/workflows/vagrant-packet.yaml
@@ -1,0 +1,17 @@
+name: Setup with Vagrant on Packet
+on:
+  push:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  vagrant-setup:
+    if: contains(github.event.pull_request.labels.*.name, 'ci-check/vagrant-setup')
+    runs-on: self-hosted
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Vagrant Test
+      run: |
+        export VAGRANT_DEFAULT_PROVIDER="virtualbox"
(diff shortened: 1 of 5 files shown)


mrmrcoleman:
@gianarb It's probably a good idea to update the README to explain a little bit about how this works, in case people need to debug build failures or add new steps.

gianarb:
Thanks @mrmrcoleman ! The PR does not even work yet! I will be sure to add everything I can add troubleshooting and debug

mrmrcoleman:
@gianarb molto bene!

gianarb:
At this point we are in the position where:

```
go test -v ./test/_vagrant
```
This command will run the Vagrant Getting Started. The package is called `_vagrant` because by default `go test` does not look at packages that start with `_`. This is an easy trick to exclude tests that should not run as part of the `go test ./...` made at the root of the project.

gianarb:
I realized this PR does not have a lot of sense. The e2e tests use `./setup.sh` and it downloads `:latest` from Docker Registry.  So we are not testing against that code.

I still think this is valuable as a workflow and at least as getting started for something that will be useful as soon as we define a release workflow and a starter `startup.sh`.

At the end the goal was to have a test case for Vagrant and this is what it is.

PR #234, opened 2020-08-05
Removed redundant documentation from tink repository

gauravgahlot:
## Description

The changes remove the redundant documentation from this repository so that [tinkerbell.org](tinkerbell.org) serves as the sole documentation source. 

## Why is this needed

It makes documentation updates easier, as we now have only one source to update, i.e. [tinkerbell.org](tinkerbell.org).

No impact.


gianarb:
I thought tinkerbell.org were using this as a source for its documentation :( 

@DailyAlice will have the last call on this I presume

DailyAlice:
Sounds good to me. I'm all for removing from here and having the one source out of tinkerbell.org.

Issue #233, opened 2020-08-05
From Drone to Github Action


gianarb:
In the ongoing effort to release the pressure to our internal drone.io and to improve independency for Tinkerbell as a project, we should move our CI/CD to GitHub Actions.

I had a look at it, but I would like to discuss here how to re-do it to see if we can make some improvements along the way.

What does it currently do to commit and pull request:

1. go test
2. runs ./ci-checks.sh
3. build docker images
4. go import, golint
5. push cli docker image
6. push tink server docker image
7. push tink worker docker image

What I think we should do in the order:

1. go fmt, go lint, goimport, #230 , go test
2. runs ci-checks.sh

I am not against but I am not 100% sure we should push images for all the commits, it sounds a waste and confusing in our registry for community people coming and looking for a particular version. I think we should always build the images to see if everything works but we should push them only when a commit lands on the master branch using the `latest` tag maybe or `unstable`

if you agree with that I will proceed with the implementation:

## THIS ISSUE IS NOT ABOUT RELEASE. It will have its own space

/cc. @mmlb @thebsdbox

thebsdbox:
I agree with the `push` of images... things should be built internally if possible (or perhaps with a `-devel`) including all of the testing tags etc.. the actual `push` should only occur on an actual release..

gianarb:
Can you elaborate @thebsdbox `things should be built internally if possible`? What does it mean internally?
I presume a release will still be managed via GitHub action, but in its own way when a tag gets pushed.

displague:
> I am not against but I am not 100% sure we should push images for all the commits, it sounds a waste and confusing in our registry for community people coming and looking for a particular version. I think we should always build the images to see if everything works but we should push them only when a commit lands on the master branch using the latest tag maybe or unstable

This sounds reasonable to me. The images I would expect to find published would be "latest", any special long-lived branches, and tagged releases.  Additional git-sha based images are noise.

When tests are run for a PR, e2e tests should avoid publishing images and rely on pushing the image onto the test nodes directly. This avoids caching problems but also keeps our tests focused on the images themselves and not the image distribution.

E2E tests on 'latest' and tagged releases may benefit from using registry copies of images, but this creates a chicken-and-egg problem. Do you publish images before the E2E test, for the E2E test? Or do you wait for the E2E test to pass (using local copies of the images) before publishing them?

thebsdbox:
>Can you elaborate @thebsdbox things should be built internally if possible?
[...]

PR #235, opened 2020-08-05
Move to GitHub action from Drone

gianarb:
## Description

Move from Drone to Github Action

## Why is this needed

Fixes: #233 

## How Has This Been Tested?

I am using `act` to run GitHub Action locally.

none

- [x] go fmt
- [x] go vet
- [x] golang-ci
- [x] go test
- [x] It runs the ci-check script
- [x] build docker image tink-cli
- [x] build docker image tink-worker
- [x] build docker image tink-server

When merged to master:

- [x] push docker image tink-cli
- [x] push docker image tink-worker
- [x] push docker image tink-server

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,104 @@
+name: For each commit and PR
+on:
+  push:
+  pull_request:
+
+jobs:
+  validation:
+    runs-on: ubuntu-latest
+    env:
+      CGO_ENABLED: 0
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.14.6'
+    - name: go fmt
+      run: go fmt ./...
+    - name: go vet
+      run: go vet ./...
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v1
+      with:
+        version: v1.30
+    - name: go vet
+      run: go test ./...
+    - name: Build binaries
+      run: make
+    - name: Upload tink-server binary
+      uses: actions/upload-artifact@v2
+      with:
+        name: tink-server
+        path: ./cmd/tink-server/tink-server
+    - name: Upload tink-cli binary
+      uses: actions/upload-artifact@v2
+      with:
+        name: tink-cli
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - name: go vet
+      run: go vet ./...
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v1
+      with:
+        version: v1.30
+    - name: go vet
+      run: go test ./...
+    - run: make
+      name: Build binaries
```
mmlb, line 29:
```yaml
- name: Build binaries
  run: make
```
to match the order of all the other actions

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - name: quay.io/tinkerbell/tink-server
+      uses: docker/build-push-action@v1
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+        registry: quay.io
+        repository: tinkerbell/tink-server
+        path: ./cmd/tink-server/
+        push: ${{ startsWith(github.ref, 'refs/heads/master') }}
+        tags: latest
```
mmlb, line 78: Please add
```yaml
tag_with_ref: true
tag_with_sha: true
```
too

gianarb: Those options will push:

```
Successfully tagged quay.io/tinkerbell/tink:latest
Successfully tagged quay.io/tinkerbell/tink:pr-235-merge
Successfully tagged quay.io/tinkerbell/tink:sha-f009303
```
I don't think we want that

mmlb: I do. Maybe not `tag_wit_ref` for pushes, but definitely for tags. But I do want the sha so we can have point-in-time refs to images, and not just latest that keeps getting pushed up.

gianarb: Ok I will leave this `tag_with_sha: true`.
About tags, that's a separate discussion btw. It is not covered here and I will add it as soon as we get closer to a release lifecycle plan.

it does not look we are close to it

mmlb: Agree.

Issue #222, opened 2020-07-20
tink-worker container should keep on running and seeking for workflow tasks to execute 


parauliya:
tink-worker container should keep on running even after the completion of all the current workflows. 

## Expected Behaviour
tink-worker container should not exit if all the workflows executed successfully.
 
## Current Behaviour
Currently tink-worker container exits once all the workflows provided to it are completed successfully.

## Possible Solution
1. Use `gRPC stream`  instead of  `sleep` for polling workflow.
2. The infinite loop inside the worker container should keep on running irrespective of worklfows state or completion.
     ( using `sleep` for polling over it ).

## Context
Right now there worker exits after the execution of workflows. So if there is a new workflow for the same worker it need to be rebooted.
Right now there is a sleep in worker container between asking for a workflow. This leads to a lot of unnecessary logs which crashes the terminal or overflowing the RAM/disks etc.
This also covers the remaining part of #196 once it is fixed.

Cbkhare:
this issue seems to be duplicate of https://github.com/tinkerbell/tink/issues/72

parauliya:
> this issue seems to be duplicate of #72

Not exactly duplicate but yes it contains the same so we can close the previous one which is #72 .

PR #237, opened 2020-08-06
Added a new rpc call workflow.GetWorkflowContextList to be used by boots

gauravgahlot:
## Description
As the worker starts, Boots checks [if there is a workflow](https://github.com/tinkerbell/boots/blob/master/packet/endpoints.go#L48) defined for that worker. If yes, it will continue to PXE and it will not otherwise. (more context [here](https://github.com/tinkerbell/boots/pull/50)).

The PR introduces a new RPC that can be used by Boots to get the workflow contexts without having to maintain a gRPC stream. 

## Why is this needed

Recently there have been [changes](https://github.com/tinkerbell/tink/issues/222) in Tink, to allow a worker to maintain a gRPC stream with the server to get the workflows. The same RPC was being by boots. 

## How Has This Been Tested?
Still a WIP. 

No direct impact on users.

## Checklist:

I have:

- [ ] added unit or e2e tests


Cbkhare:
Let me first test it with boots then merge it.

Cbkhare:
changes done in accordance with https://github.com/tinkerbell/tink/issues/238

Cbkhare:
Tested the workflow execution using this PR. changes are working correctly.
For more ref follow https://github.com/tinkerbell/boots/pull/64

Cbkhare:
> Can we write a unit test for this?! I would like to have a table test that insert a set of fixtures and checks what we expect. This is a crucial part of the application @gauravgahlot can you add that to this PR?

@gianarb Will it be fine if we merge this PR and @gauravgahlot can add table test for it in a separate PR. To fix issues related to #222 , this is a kind of priority.

gianarb:
Yes I agree but let's get it done as soon as it is merged please

Cbkhare:
> Yes I agree but let's get it done as soon as it is merged please

Thanks @gianarb ,

PR #239, opened 2020-08-06
Make golint-ci happy

gianarb:
We will use golint-ci with staticcheck and some other useful analysis.

This PR makes golint-ci happy


gauravgahlot:
@gianarb you have conflicts now. :smile:

Issue #219, opened 2020-07-16
Document all public interfaces


displague:
## Expected Behaviour

As a developer, the interfaces of the client should be easy to work with.

The endpoints should be documented in a way that describes what the purpose of each endpoint is and what the parameters do.

## Current Behaviour

Endpoints are not documented: https://github.com/tinkerbell/tink/blob/0100e535bd94c14b95eeedbbe46c576d1f344301/protos/hardware/hardware.pb.go#L2023

This makes features like #195 (OpenAPI generated documentation and clients) less valuable.

## Possible Solution
Document (as code comments) each of the public functions, types, and fields for the 3 Client interfaces.
https://github.com/tinkerbell/tink/blob/master/client/main.go#L66-L68

gianarb:
Hello @displague ! Thank you for opening this PR. I think comments has to be improved and you identify a very valuable place to start with!

Are you interested in opening a PR with the 3 comments you suggested?! 

Thanks

rugwirobaker:
I'm just getting started going through the codebase and I think this would be very much helpful. If no else has I'd like to pick up.

PR #243, opened 2020-08-07
Add documentation to proto files

displague:
## Description

This PR:
* makes `protos/protoc.sh` executable in environments without protoc (and c++) installed
* docker provides a reproducible protoc environment for regenerating pb.go files (jaegertracing/protobuf, libprotoc 3.8.0)
* adds documentation to the hardware.proto file

## Why is this needed

This starts the effort needed for #219

Additional .proto file docs would be needed to close that issue.

## How Has This Been Tested?

Let me know how I should test this.

I would like to include a `make protos` task.
Should this be run by CI to check for uncommitted drift?

## Checklist:

I have:

- [ ] added unit or e2e tests


gianarb:
I like the end goal: documentation and easy to replicate. But this adds a dependency with Docker and I don't like that in the same way.

If this is something you would like to add-in CI we use `nix` and shell.nix to resolve dependencies in the ci-checks.sh can you follow the same pattern?

displague:
> this adds a dependency with Docker and I don't like that in the same way.

What are your thoughts on taking this PR as-is for the documentation changes and the protoc.sh changes (when the existing tests pass)? 

If `protoc` is already installed, it won't use Docker.  The protoc.sh script generates the `.pb.go` files that should be included in each commit (that modifies .proto files), it needs to be run by the developer locally.

The `protoc.sh` script is not usable without `protoc` installed, which is complicated because you need to use a package manager, download and unzip, or compile it (requiring a c++ environment).  This makes me ❤️ `go get` even more.

I haven't added any build time changes in this PR. We could make this a part of CI later.

Independent of that, we can create a `make protos` task, which could replace the `protoc.sh` script.  

Then, as I see it, `make protos` will have to do one of four things:

1. Require the expected version of `protoc` to be installed, with an additional requirement for the expected version of the `proto` includes from Google
2. or install these tools for you (in a version and build path isolated way)
3. or use an existing Docker image
4. or use and maintain our own Docker image 

1 and 3 are the most sustainable, 3 is the easiest on new contributors.

kdeng3849:
Small nitpick, can you run `goimports` so that the imports inside the `pb.go` files get grouped properly?

mmlb:
https://github.com/tinkerbell/tink/pull/283/commits/6f757ecf1c299e1d17c3d674da80e087da744819 is relevant to this, addresses @kdeng3849's feedback too.

displague:
@kdeng3849 @mmlb thanks for the goimports addition!

This is rebased and ready for another look.

kdeng3849 (approved):

protos/protoc.sh
```diff
@@ -1,9 +1,33 @@
...
+# Alternative images like grpc/go are very dated and do not
+# include the needed plugins and includes.
+#
 set -e
 
+GOPATH=${GOPATH:-$(go env GOPATH)}
+
+if command -v protoc >/dev/null; then
+	GW_PATH="$GOPATH"/src/github.com/grpc-ecosystem/grpc-gateway
+	GO111MODULES=on go get github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
```
mmlb, line 15: Fetching should be unnecessary since its a required dep and should have already been fetch by `go` at the correct version. That being the case you can also get the file from the go mod src path, though this requires knowing the version number used. For example I've got:
```sh
ls -1d /home/manny/go/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@*/
/home/manny/go/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.14.6/
/home/manny/go/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.14.7/
/home/manny/go/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.14.8/
```
We can pull out the version number from go.mod pretty easily though.

protos/protoc.sh
```diff
@@ -1,9 +1,33 @@
...
+
+	PROTOC="protoc -I/usr/local/include -I$GW_PATH/third_party/googleapis --plugin=protoc-gen-grpc-gateway=$GOPATH/bin/protoc-gen-grpc-gateway"
+else
+	IMAGE=jaegertracing/protobuf:0.2.0
```
mmlb, line 19: This image also doesn't guarantee the same versions of grpc-gateway as we have embedded. I'm not sure its that big of a deal but does feel a bit wrong.

protos/protoc.sh
displague: We could build the container with the values that we want to use: https://github.com/jaegertracing/docker-protobuf/blob/master/Dockerfile#L1-L10.

A series or `go list -m all | grep github.com/grpc-ecosystem/grpc-gateway | awk '{print $2}'` (for each dependency) and `docker build --build-arg ...` could do this if we want to build our own version of the image.

mmlb: Yeah, and this PR doesn't make things worse in that respect anyway, so :+1:

mmlb (approved):

PR #247, opened 2020-08-09
Refactoring db package for unit tests

gauravgahlot:
## Description

Changes include: 
- [x] refactoring `db` package
- [x] added mocked implementation for the `database` interface

## Why is this needed

For mocking database calls to write unit tests for gRPC server. 

No impact.

## How Has This Been Tested?

I have tested the changes by executing an end-to-end workflow.


grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
 func (s *server) GetWorkflowContexts(req *pb.WorkflowContextRequest, stream pb.WorkflowSvc_GetWorkflowContextsServer) error {
-	return exec.GetWorkflowContexts(req, stream, s.db)
+	wfs, err := getWorkflowsForWorker(s.db, req.WorkerId)
+	if err != nil {
+		return err
+	}
+	for _, wf := range wfs {
+		wfContext, err := s.db.GetWorkflowContexts(context.Background(), wf)
+		if err != nil {
+			return status.Errorf(codes.Aborted, "invalid workflow %s found for worker %s", wf, req.WorkerId)
```
Cbkhare, line 26: Error Statement seems to be incorrect. Is it always `invalid workflow` if `err!=nil`. It could also mean sql query failed or connection to db ?

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
-	return exec.GetWorkflowContextList(context, req, s.db)
+	wfs, err := getWorkflowsForWorker(s.db, req.WorkerId)
+	if err != nil {
+		return nil, err
+	}
+	wfContexts := []*pb.WorkflowContext{}
+	for _, wf := range wfs {
+		wfContext, err := s.db.GetWorkflowContexts(context, wf)
+		if err != nil {
+			return nil, status.Errorf(codes.Aborted, "Invalid workflow %s found for worker %s", wf, req.WorkerId)
```
Cbkhare, line 47: Error statement, same as above.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	if len(wfID) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "workflow_id is invalid")
+	}
+	if len(req.GetTaskName()) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "task_name is invalid")
+	}
+	if len(req.GetActionName()) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "action_name is invalid")
+	}
+	fmt.Printf("Received action status: %s\n", req)
```
Cbkhare, line 77: we should this `fmt` and use log.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	wfContext, err := s.db.GetWorkflowContexts(context, wfID)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "workflow context not found for workflow %s", wfID)
```
Cbkhare, line 80: this error statement seems to be valid with my above comment on same point. May be we can refine the statement and it here as well.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	}
+	wfActions, err := s.db.GetWorkflowActions(context, wfID)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "workflow actions not found for workflow %s", wfID)
+	}
+
+	// We need bunch of checks here considering
```
Cbkhare, line 87: type, considering is mentioned twice

Cbkhare: Also are we planning to add check or are they here ?

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	}
+	return &pb.WorkflowContextList{
+		WorkflowContexts: wfContexts,
+	}, nil
 }
 
 // GetWorkflowActions implements tinkerbell.GetWorkflowActions
 func (s *server) GetWorkflowActions(context context.Context, req *pb.WorkflowActionsRequest) (*pb.WorkflowActionList, error) {
-	return exec.GetWorkflowActions(context, req, s.db)
+	wfID := req.GetWorkflowId()
```
Cbkhare, line 58: `WorkflowActionsRequest` struct has the `WorkflowId` field is described as omit empty in `workflow.pb.go`. If `req` does not have this field then `GetWorkflowId()` call would panic as this method does not check whether the field is empty or not. Though it does check whether req is empty or not but it does not serve the purpose. 

For you implementation to work correctly, I would suggest you to add a check in `GetWorkflowId` to verify if req has field `WorkflowId` or not.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	if wfs == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "Worker not found for any workflows")
+	}
+	return wfs, nil
+}
+
+func getWorkflowActions(context context.Context, db db.Database, wfID string) (*pb.WorkflowActionList, error) {
+	actions, err := db.GetWorkflowActions(context, wfID)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "workflow_id is invalid")
```
Cbkhare, line 183: Same comment on error statement.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	}
+	return actions, nil
+}
+
+// The below function check whether a particular workflow context is applicable or needed to
```
Cbkhare, line 188: `isApplicableToSend` ...

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	}
+	return &pb.GetWorkflowDataResponse{Data: data}, nil
 }
 
 // GetWorkflowMetadata returns metadata wrt to the ephemeral data of a workflow
 func (s *server) GetWorkflowMetadata(context context.Context, req *pb.GetWorkflowDataRequest) (*pb.GetWorkflowDataResponse, error) {
-	return exec.GetWorkflowMetadata(context, req, s.db)
+	data, err := s.db.GetWorkflowMetadata(context, req)
+	if err != nil {
+		return &pb.GetWorkflowDataResponse{Data: []byte("")}, status.Errorf(codes.Unknown, err.Error())
```
Cbkhare, line 155: instead of err.Error() we should also send some error statement.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
 // Get Workflow Ephemeral Data
 func (s *server) GetWorkflowData(context context.Context, req *pb.GetWorkflowDataRequest) (*pb.GetWorkflowDataResponse, error) {
-	return exec.GetWorkflowData(context, req, s.db)
+	wfID := req.GetWorkflowID()
+	if len(wfID) == 0 {
+		return &pb.GetWorkflowDataResponse{Data: []byte("")}, status.Errorf(codes.InvalidArgument, "workflow_id is invalid")
+	}
+	data, err := s.db.GetfromWfDataTable(context, req)
+	if err != nil {
+		return &pb.GetWorkflowDataResponse{Data: []byte("")}, status.Errorf(codes.Unknown, err.Error())
```
Cbkhare, line 146: same as below.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+		return nil, status.Errorf(codes.FailedPrecondition, "reported action name not matching in actions info")
+	}
+	wfContext.CurrentWorker = action.GetWorkerId()
+	wfContext.CurrentTask = req.GetTaskName()
+	wfContext.CurrentAction = req.GetActionName()
+	wfContext.CurrentActionState = req.GetActionStatus()
+	wfContext.CurrentActionIndex = actionIndex
+	err = s.db.UpdateWorkflowState(context, wfContext)
+	if err != nil {
+		return &pb.Empty{}, fmt.Errorf("failed to update the workflow_state table. Error : %s", err)
```
Cbkhare, line 109: shouldn't we be sending status code as mentioned inline 100. 

Same comment for line 115.

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+		}
+	}
+	action := wfActions.ActionList[actionIndex]
+	if action.GetTaskName() != req.GetTaskName() {
+		return nil, status.Errorf(codes.FailedPrecondition, "reported task name not matching in actions info")
+	}
+	if action.GetName() != req.GetActionName() {
+		return nil, status.Errorf(codes.FailedPrecondition, "reported action name not matching in actions info")
+	}
+	wfContext.CurrentWorker = action.GetWorkerId()
```
Cbkhare, line 102: same as above comment on GetWorkflowId

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+	wfContext.CurrentTask = req.GetTaskName()
+	wfContext.CurrentAction = req.GetActionName()
```
Cbkhare, line 104: All these `Get..` have an incorrect way of handling the struct fields. 

I can understand that these are from prior commits, but it can affect your implementation. How do you think we should handle it. ?

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+		if actions.ActionList[wfContext.GetCurrentActionIndex()].GetWorkerId() == workerID {
+			log.Println("Send the workflow context ", wfContext.GetWorkflowId())
+			return true
+		}
+	}
+	return false
+}
+
+func isLastAction(wfContext *pb.WorkflowContext, actions *pb.WorkflowActionList) bool {
+	return int(wfContext.GetCurrentActionIndex()) == len(actions.GetActionList())-1
```
Cbkhare, line 219: the same problem of `Get` methods which I mentioned above.

Cbkhare, line 219: return type of `GetCurrentActionIndex` is already int. So, int conversion won't be required?

grpc-server/tinkerbell.go
```diff
@@ -2,47 +2,219 @@ package grpcserver
...
+		if isLastAction(wfContext, actions) {
+			return false
+		}
+		if wfContext.GetCurrentActionIndex() == 0 {
+			if actions.ActionList[wfContext.GetCurrentActionIndex()+1].GetWorkerId() == workerID {
+				log.Println("Send the workflow context ", wfContext.GetWorkflowId())
+				return true
+			}
+		}
+	} else {
```
Cbkhare, line 209: not necessary but we can combine this else --> else if

grpc-server/tinkerbell.go
gauravgahlot: Not that I'm aware of. I will remove the comment and in case we encounter any issues we always come back to add more checks.

grpc-server/tinkerbell.go
gauravgahlot: Yeah, I agree. We should instead return the error as is to the caller.

grpc-server/tinkerbell.go
gauravgahlot: It is required because `GetCurrentActionIndex` returns an `int64` while `len` returns an `int`.

grpc-server/tinkerbell.go
gauravgahlot: I believe you are referring to this one [here](https://github.com/tinkerbell/tink/blob/master/protos/workflow/workflow.pb.go#L731):
```
WorkflowId string `protobuf:"bytes,1,opt,name=workflow_id,json=workflowId,proto3" json:"workflow_id,omitempty"`
```
The `omitempty` is for JSON encoding and doesn't affect getting the value in Go code (like `req.GetWorkflowId()`). So, if workflow id is not set, the method call will always return `""` (an empty string).

grpc-server/tinkerbell.go
gauravgahlot: Marking resolved, based on this [thread](https://github.com/tinkerbell/tink/pull/247#discussion_r468480839).

grpc-server/tinkerbell.go
Cbkhare: I was talking about this method. https://github.com/tinkerbell/tink/blob/master/protos/workflow/workflow.pb.go#L480 

There is no error handling to check if `WorkflowId` is not present and eventually it will panic. 

```
func (x *WorkflowContext) GetWorkflowId() string {
	if x != nil {
		return x.WorkflowId
	}
	return ""
}
```

grpc-server/tinkerbell.go
Cbkhare: Let me know once you have updated, I will resolve the comments.

db/mock/hardware.go
```diff
@@ -0,0 +1,35 @@
+package mock
+
+import (
+	"context"
+)
+
+// DeleteFromDB : delete data from hardware table
+func (d DB) DeleteFromDB(ctx context.Context, id string) error {
```
gianarb, line 8: I just have a feedback. When I create a mock I usually use the form:

```go
type DB struct {
    DeleteFromDBFunc func(ctx context.Context, id string) error
    // and all the function like InsertIntoDB should have its InsertIntoDBFunc...
}
```
In this way the real function will use it:

```go
func (d DB) DeleteFromDB(ctx context.Context, id string) error {
	return d.DeleteFromDBFunc(ctx, id)
}
```

And from the test I can inject the behavior I need:

```go
func TestDBReturnsNotFound(t *testing.T) {
    db := DB{
        DeleteFromDBFunc: func(ctx context.Context, id string) error {
            return errros.Error("not found")
        },
    }
    hardware(db)
}
```

gauravgahlot: Addressed in PR #256

Issue #74, opened 2020-04-23
Cut release tags from tink


alexellis:
Git enables releases to be made, these are shown in the GitHub Releases page.

<img width="1029" alt="Screenshot 2020-04-23 at 10 53 19" src="https://user-images.githubusercontent.com/6358735/80085802-a926ed80-8550-11ea-8199-720b2d5a0494.png">

Tink currently has none, however I would suggest that the project starts to use them, even if they are RC tags or very early alpha tags.

For instance:

0.1.0
0.1.1

and so forth...

These tags can be checked out, show progress being made to internal and external stakeholders, and allow users to reason about "versions"

Example: https://github.com/alexellis/arkade/releases

If you install the Derek app on this repo, it will generate notes with all the PRs and commits pushed automatically.

<img width="1011" alt="Screenshot 2020-04-23 at 10 53 04" src="https://user-images.githubusercontent.com/6358735/80085784-a0361c00-8550-11ea-9a5d-fe70142bc06d.png">

alexellis:
I would also add that you do not need to push Docker images or cut binaries for each release, although that would be a nice addition.

thebsdbox:
https://github.com/tinkerbell/tink/issues/94 The version subcommand will also become part of the same piece of work.

gianarb:
We decided to release the only sandbox for now because this is a too fast-moving project at this stage and the responsibility is not totally split. It means that it is very hard to tag all the projects by themself. 

I will close this issue for now and have a look at https://github.com/tinkerbell/sandbox/releases if you want a pinned version and binaries as well!

PR #251, opened 2020-08-15
Versions and completions

### LINKED ISSUE #94: Add version subcommand
With `tink` binary, it should be possible to check version of both CLI tool and ideally server, similar to `kubectl version` functionality, as currently, there is no way to reference which version of the binary is in use when reporting the issues.

displague:
## Description

* Fixes testing via Makefile `make test` (did not test anything before)
* Fixes #94 based on the git SHA and eventual git tags (blocked by #74)
  * Adds `tink-cli --version` output
  * Because `tink-server` and `tink-worker` do not currently take parameters, the version is now emitted at start.
    * `tink-server` emits this at INFO level (because it is a long-lived process)
    * `tink-worker` emits this at DEBUG level (because the default run is quiet, afaict)
* Adds a `tink-cli completion` subcommand

## Why is this needed

Fixes: #94

## How Has This Been Tested?

* `tink-cli completion --help`
* `tink-cli completion bash`
* ```
   cmd/tink-cli/tink-cli --version
   tink version f9500b2
   ```
* ```
   PACKET_VERSION="x" PACKET_ENV="t" ROLLBAR_TOKEN="t" cmd/tink-server/tink-server 
   {"level":"info","ts":1597460174.581983,"caller":"tink-server/main.go:30","msg":"starting version f9500b2","service":"github.com/tinkerbell/tink"}
   {"level":"info","ts":1597460174.582356,"caller":"metrics/metrics.go:58","msg":"initializing label values","service":"github.com/tinkerbell/tink"}
   ```
* ```
   TINKERBELL_CERT_URL="l" PACKET_VERSION="x" PACKET_ENV="t" ROLLBAR_TOKEN="t" WORKER_LOG_LEVEL=debug cmd/tink-worker/tink-worker
   DEBU[0000] Starting version f9500b2                     
   INFO[0000] RETRY_INTERVAL not set. Using default, 3 seconds 
   INFO[0000] MAX_RETRY not set. Using default, 3 retries. 
   FATA[0000] fetch cert: Get "l": unsupported protocol scheme ""
   ```

<!--- Fixes a bug, unblocks installation, removes a component of the stack etc -->
<!--- Requires a DB migration script, etc. -->


## Checklist:

I have:

- [ ] added unit or e2e tests


displague:
@gianarb I can move completions to another PR and rebase either this PR or that one after merge.

If it helps, I can create a markdown and/or man-page generation process here.  `packet-cli` performs a similar step using spf13/cobra.  I can add this as a separate PR.

https://github.com/packethost/packet-cli/tree/master/docs is generated on each build and commits must include docs/ without build jitter. What are your thoughts, @DailyAlice?

gianarb:
I don't understand. I think we should just write somewhere how to set up completion 
https://kubernetes.io/docs/tasks/tools/install-kubectl/#enabling-shell-autocompletion

gianarb (approved):
It looks like documentation is on hold for now. This is good to go!

Issue #73, opened 2020-04-23
Provide release binary of tink CLI


alexellis:
A statically-linked release binary should be made available of the tink CLI. Users shouldn't have to download go to build the CLI, or wonder where to get it from. I couldn't find docs explaining where to find it. There also are no release tags of tink, but there should be, even if the tag is 0.1.0

For an example, see how we do this with arkade or inlets

Build config for Travis/Make:

https://github.com/alexellis/arkade/blob/master/.travis.yml
https://github.com/alexellis/arkade/blob/master/Makefile

Install script:

https://github.com/alexellis/arkade/blob/master/get.sh

alexellis:
Ref: #74

alexellis:
No external tools like goreleaser are strictly needed. Go supports cross-compilation with a couple of flags. See the Makefile example above for instance.

At a minimum the CLI should have a Linux `x86_64` and MacOS option.

invidian:
There is also https://github.com/tinkerbell/boots/issues/4. Perhaps one solution should be picked for all binaries.

I personally like Goreleaser, as it automates building and uploading artifacts to GitHub, making GitHub releases really easy.

alexellis:
In my opinion a Makefile would suffice, goreleaser is a popular, but complex tool.

Bumping this up. @parauliya @gauravgahlot is there an update on this? For instance, after running the setup script, it would be ideal to be able to run "tink" without exec'ing into a container.

gianarb:
We decided to release not every single project for now, but the entire stack, at the end the services are strictly connected and the project evolves quickly at this stage, it is very time-consuming to look at each project in isolation.

Binaryies are here: https://github.com/tinkerbell/sandbox/releases

I will close this issue for now because it won't be implemented soon

PR #252, opened 2020-08-15
use goreleaser to build binaries on tagged releases

displague:
## Description

* Adds GoReleaser and a GitHub Workflow to run when tags are pushed.
* Binaries are created for several architectures and operating systems. 

See https://github.com/displague/tink/releases for an example (based on this commit) of what a release will look like.

Future iterations could bundle these binaries with documentation as archive files, or build debs, rpms, Docker images, etc.

Goreleaser's own .goreleaser.yml is a handy demonstration: https://github.com/goreleaser/goreleaser/blob/master/.goreleaser.yml

## Why is this needed

Fixes: #73

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -17,6 +17,13 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: 1.14
+      - name: Docker Login
+        if: success() && startsWith(github.ref, 'refs/tags/v')
+        env:
+          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
+          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
+        run: |
+          echo "${DOCKER_PASSWORD}" | docker login --username "${DOCKER_USERNAME}" --password-stdin
       - name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
@@ -24,3 +31,8 @@ jobs:
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          DOCKER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Clear
+        if: always() && startsWith(github.ref, 'refs/tags/v')
+        run: |
+          rm -f ${HOME}/.docker/config.json
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: release
+
+on:
+  push:
+    tags:
+      - "*"
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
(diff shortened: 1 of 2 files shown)


gianarb:
Great work

During my past experience with GoReleaser and multi-arch support for Docker images, I can say that it is not easy to do in a clean way. At the end of the story, we didn't use it for other projects like packet-ccm or CAPP for that reason.

We have to support as part of the release workflow something that can build and push multi-arch Docker images as well, not only the binaries. And I would like to merge a PR that contains both

displague:
I've made changes to publish amd64 Docker images in the latest builds. 

I'm testing changes to add arm64 and armv6/armv7 images.

displague:
Hmm @gianarb, it looks like this would only be possible for goreleaser to push multiple architectures images using a separate Docker repository for each architecture (examples: tinkerbell/tink-server, tinkerbell/tink-server-arm64, tinkberll/tink-server-armv6, etc).

This is due to https://github.com/goreleaser/goreleaser/issues/530

I see four options here:
- close this PR entirely (in favor of another approach in the future)
- create unique Docker repositories for each architecture (and publish binaries and docker images for all arch/os combos)
- build binaries for all arch/os combos and publish only amd64 Docker images (for now, using goreleaser)
- build binaries only, don't publish docker images for now

Which would you prefer?

displague:
A differentiator between this project and packet-ccm/csi-packet is that this project benefits from providing tink-cli binaries where those other projects are primarily consumed as containers.

What binaries do Tinkerbell users need? Only `tink-cli`? Should `tink-server` and `tink-worker` only be provided as containers?

If we take a`docker manifest` or `docker buildx` pattern it would be nice to reuse the same binaries that are archived on release (one of the benefits of the goreleaser pattern).  This would cut down on build resources and would ensure that the different execution methods are still close (using the same binary).

gianarb:
We should not tight ourself to the idea that containers are the only one to run tinkerbell because it is not true.

Tinkerbell today depends on Docker but I see docker as a CRI. Technically I am sure we can even use the same interface.

Anyway, tink should release tink-cli, tink-server and tink-worker as binary + containers version of them in multi-arch and OS.

> If we take adocker manifest or docker buildx pattern it would be nice to reuse the same binaries that are archived on release (one of the benefits of the goreleaser pattern). This would cut down on build resources and would ensure that the different execution methods are still close (using the same binary).

I agree and that's why I think docker + binaries should be part of the same PR because you will see how complicated it is to make with goreleaser

displague:
I see. Work in https://github.com/tinkerbell/proposals/pull/5 should be allowed to finish first.

I'll close this PR in its current form (so others don't wait for this one and so they aren't prevented from trying alternate directions). I may revisit this.  

> a nightly release for tink (other repos will follow that will run as GitHub Action at a certain schedule)

I don't see the need for a nightly release. If we are using semver and pushing latest images from the default-branch (on every commit), what purpose do nightly binaries serve?  I think the only reason nightly makes sense now is that we have not stamped a single tagged release.  Why not just push 0.1.0 (#74) on all of the current projects in their current state?

This is probably a better conversation for https://github.com/tinkerbell/proposals/pull/5/files#diff-5ae2d202d9cad195081db97be153eaa9R139.

gianarb:
nightly is just a very easy way to test the release process continuously

PR #261, opened 2020-08-19
Add codecov as part of ci checks

gianarb:
Osie uses codecov already, this PR extends its use with the tink
repository.

Right now there is not support for PR comments or things like that. It
just updates codecov


displague (dismissed):
I assume merging to master will fix the warning codecov is reporting.

https://docs.codecov.io/docs/error-reference#section-missing-base-commit

gianarb:
Yes @displague

gauravgahlot (approved):
Much awaited.

PR #262, opened 2020-08-19
Increase go test timeout for vagrant tests

gianarb:
Right now there is only one test that runs as part of the vagrant
test suite the default go test timeout (10s) is enough.

It won't stay for long.


thebsdbox (approved):
This is approved.

PR #267, opened 2020-08-26
Added changes for centralized logging and rsyslog

Cbkhare:
Signed-off-by: cbkhare <Chitrabasukhare89@gmail.com>

## Description

Changes to add centralized logging for the Tinkerbell. 

## Why is this needed

- We need a log aggregator for Tinkerbell which can collect log from tink services and tink worker containers. 
- boots PR: https://github.com/tinkerbell/boots/pull/73
- osie PR: https://github.com/tinkerbell/osie/pull/104
- RFD: https://github.com/tinkerbell/proposals/pull/7

Fixes: #

## How Has This Been Tested?
- [x] Tested with basic hello-world example.
- [x] Verified the testing of logrotation.
- [x] Added below test results. 

**log_aggregation**
```
vagrant@provisioner:/vagrant/deploy$ ls -lrt   /var/log/dockerlfs/
total 20
drwxr-xr-x 2 syslog syslog 4096 Sep  8 12:59 provisioner
-rw-r----- 1 syslog adm    9521 Sep  8 13:04 daemon.log
drwxr-xr-x 2 syslog syslog 4096 Sep  8 13:17 192.168.1.5
vagrant@provisioner:/vagrant/deploy$ ls -lrt   /var/log/dockerlfs/192.168.1.5/
total 12
-rw-r----- 1 syslog adm 2985 Sep  8 13:17 hello_world_0e65404d-5abe-4746-b6a9-bba9be867884.log
-rw-r----- 1 syslog adm 4789 Sep  8 13:17 tink-worker.log
```

**logrotation**:
```
vagrant@provisioner:/vagrant/deploy$ ls -lrt   /var/log/dockerlfs/provisioner/
total 36
-rw-r----- 1 syslog adm  513 Sep  8 13:04 deploy_db_1.log-20200908131599572256.gz
-rw-r----- 1 syslog adm 4099 Sep  8 13:17 deploy_boots_1.log-20200908131599572256.gz
-rw-r----- 1 syslog adm  577 Sep  8 13:17 deploy_nginx_1.log-20200908131599572256.gz
-rw-r----- 1 syslog adm  400 Sep  8 13:17 deploy_hegel_1.log-20200908131599572256.gz
-rw-r----- 1 syslog adm 1450 Sep  8 13:17 deploy_tink-server_1.log-20200908131599572256.gz
-rw-r----- 1 syslog adm 6155 Sep  8 13:37 deploy_registry_1.log-20200908131599572256.gz
-rw-r----- 1 syslog adm    0 Sep  8 13:37 deploy_boots_1.log
-rw-r----- 1 syslog adm    0 Sep  8 13:37 deploy_db_1.log
-rw-r----- 1 syslog adm    0 Sep  8 13:37 deploy_hegel_1.log
-rw-r----- 1 syslog adm    0 Sep  8 13:37 deploy_nginx_1.log
-rw-r----- 1 syslog adm    0 Sep  8 13:37 deploy_tink-server_1.log
-rw-r----- 1 syslog adm  147 Sep  8 13:37 deploy_registry_1.log
vagrant@provisioner:/vagrant/deploy$ ls -lrt   /var/log/dockerlfs/192.168.1.5/
total 8
-rw-r----- 1 syslog adm 571 Sep  8 13:17 hello_world_0e65404d-5abe-4746-b6a9-bba9be867884.log-20200908131599572256.gz
-rw-r----- 1 syslog adm 918 Sep  8 13:17 tink-worker.log-20200908131599572256.gz
-rw-r----- 1 syslog adm   0 Sep  8 13:37 hello_world_0e65404d-5abe-4746-b6a9-bba9be867884.log
-rw-r----- 1 syslog adm   0 Sep  8 13:37 tink-worker.log
vagrant@provisioner:/vagrant/deploy$ ls -lrt   /var/log/dockerlfs/
total 12
-rw-r----- 1 syslog adm    1642 Sep  8 13:04 daemon.log-20200908131599572256.gz
-rw-r----- 1 syslog adm       0 Sep  8 13:37 daemon.log
drwxr-xr-x 2 syslog syslog 4096 Sep  8 13:37 192.168.1.5
drwxr-xr-x 2 syslog syslog 4096 Sep  8 13:37 provisioner

``` 

- Log location has been updated. Documentation will need an update.

## Checklist:

I have:


fransvanberckel:
Looking forward: Within Annsible it's all about naming. I am not sure how it gonna handle; container_name/{{.Name}}

fransvanberckel:
And LOG_SERVER_ADDRESS looks more like a url then an address.

Cbkhare:
> And LOG_SERVER_ADDRESS looks to me more like a URL then an ip address.

@fransvanberckel yes, this URL address specifies the connection type as well. The naming convention of this variable is with respect to docker log drivers. https://docs.docker.com/config/containers/logging/syslog/ 

you can find RFD for this change in here https://github.com/tinkerbell/proposals/pull/7/files.

setup.sh
```diff
@@ -485,6 +485,18 @@ whats_next() (
 	echo "$BLANK    Follow the steps described in https://tinkerbell.org/examples/hello-world/ to say 'Hello World!' with a workflow."
 )
 
+confiure_rsyslog_logging() (
```
thebsdbox, line 488: `confiure_rsyslog_logging ` -> `configure_rsyslog_logging `

setup.sh
```diff
@@ -505,6 +517,7 @@ do_setup() (
 
 	setup_osie
 	generate_certificates
+	confiure_rsyslog_logging
```
thebsdbox, line 520: confiure_rsyslog_logging -> configure_rsyslog_logging

thebsdbox:
What happens if we don't want `rsyslog` .. Also where does `rsyslog` come from, who installs it?

setup.sh
Cbkhare: noted.

setup.sh
Cbkhare: done

setup.sh
Cbkhare: done

Cbkhare (commented):
> What happens if we don't want `rsyslog` .. Also where does `rsyslog` come from, who installs it?

If we don't want `rsyslog` then we can update the logging driver as per the steps mentioned in here https://github.com/tinkerbell/tinkerbell.org/pull/135/files#diff-1fe57f07e24a8681cab7f4044badecb4. 

It comes by default in Centos, Debian and Ubuntu.

setup.sh
```diff
@@ -505,6 +517,7 @@ do_setup() (
 
 	setup_osie
 	generate_certificates
+	configure_rsyslog_logging
```
mmlb, line 520: @Cbkhare @gianarb I thought we said we weren't going to make centralized logging a requirement? The docker-compose.yml was a sticky point but it doesn't seem too hard to handle that. Either with a `docker-compose-centralized-logging.yml` (template file) or with `sed` magic.

gianarb: I agree with you. My expectation as an outcome for the centralizing logging work was to get only solid documentation. Not sure how to review this PR.

Cbkhare: I would request some conclusion here as this change has taken some good amount of time and effort. 

Should I add the change to support logging without centralization (aggregation)?

Cbkhare:
On my PRs for the feature of Centralised logging, i have added the support where user can use the Tinkerbell with existing logging system i.e with `docker logs <service-name> `. And, at the same time if user wishes, he/she can also use the centralized logging system as per the choice of his/her log driver by updating some set of env variables. Also as a part of the implementation, we will give the base platform and configuration to run centralized logging with rsyslog.
 I will update the doc https://github.com/tinkerbell/tinkerbell.org/pull/135 in some time, but PR for tink/boots/osie is ready.

setup.sh
Cbkhare: Though now we have the support of logging without centralization, This change of rsyslog would be required to provide the base configuration to the user if at some time he/she wishes to use the centralized logging. 

Conclusively, Tinkerbell has now a addon support for centralised logging along with the existing framework. And this centralized logging would also support plug and play model.

deploy/docker-compose-rsyslog.yml
```diff
@@ -0,0 +1,183 @@
...
+      - ./db/tinkerbell-init.sql:/docker-entrypoint-initdb.d/tinkerbell-init.sql:ro
+      - postgres_data:/var/lib/postgresql/data:rw
+    ports:
+      - 5432:5432
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U tinkerbell"]
+      interval: 1s
+      timeout: 1s
+      retries: 30
+    logging:
```
gianarb, line 54: This is not necessary, let's remove all these extra things. This is the way we should go
https://docs.docker.com/config/containers/logging/configure/

Do not touch docker-compose for logging.

cmd/tink-worker/action.go
```diff
@@ -59,9 +59,10 @@ func executeAction(ctx context.Context, action *pb.WorkflowAction, wfID string)
 
 	failedActionStatus := make(chan pb.ActionState)
 
-	//capturing logs of action container in a go-routine
-	go captureLogs(ctx, id)
-
+	if os.Getenv("CENTRALIZED_LOGGING") != "enable" {
```
gianarb, line 62: `CENTRALIZED_LOGGING` should not be required. If you configure centralize logging it is enable, if you do not configure anything... It is off. This boolean value has no sense.

gianarb (changes requested):

cmd/tink-worker/action.go
Cbkhare: This boolean check is required to start the action containers with respective log-drivers if defined. This check can not be removed.

mmlb (commented):
@Cbkhare I think @gianarb and I have the same concerns for tink-worker. Why does tink-worker need to do anything _at all_? Why not just have the docker daemon configured with the correct log-driver and opts and then tink-worker just works and doesn't need to do anything?

The only gotcha I'd see there is trying to have tags applied that would help to group workflows/actions/tasks for example. I'm not sure if we can configure docker daemon to syslog/fluentd and only specify --log-opt tag=sometag or --log-opt env=some,tags,here w/o adding log-driver.

Cbkhare:
> @Cbkhare I think @gianarb and I have the same concerns for tink-worker. Why does tink-worker need to do anything _at all_? Why not just have the docker daemon configured with the correct log-driver and opts and then tink-worker just works and doesn't need to do anything?

@mmlb If we go by the docker daemon docker approach then the changes will be made on the host. So, for Tinkerbell services changes will be made on the provisioner node and for the tink-worker/action containers changes will be needed on the worker node. This will create a dependency to ensure that after a worker node is booted and before tink-worker container starts, worker node docker daemon is **_pre-configured_** with the appropriate driver, maybe via a script or as a part of `workflow-helper.sh` with minor changes. 

> The only gotcha I'd see there is trying to have tags applied that would help to group workflows/actions/tasks for example. I'm not sure if we can configure docker daemon to syslog/fluentd and only specify --log-opt tag=sometag or --log-opt env=some,tags,here w/o adding log-driver.

yes, you right. With docker daemon, env variables will be removed and instead a file could be referenced. 

Docker daemon change will cover more scope other than just configuring log-drivers, and this is something I am willing to avoid to ensure we can have the feature in place without making any system level changes.

mmlb:
> > @Cbkhare I think @gianarb and I have the same concerns for tink-worker. Why does tink-worker need to do anything _at all_? Why not just have the docker daemon configured with the correct log-driver and opts and then tink-worker just works and doesn't need to do anything?
> 
> @mmlb If we go by the docker daemon docker approach then the changes will be made on the host. So, for Tinkerbell services changes will be made on the provisioner node and for the tink-worker/action containers changes will be needed on the worker node. This will create a dependency to ensure that after a worker node is booted and before tink-worker container starts, worker node docker daemon is **_pre-configured_** with the appropriate driver, maybe via a script or as a part of `workflow-helper.sh` with minor changes.
> 
> > The only gotcha I'd see there is trying to have tags applied that would help to group workflows/actions/tasks for example. I'm not sure if we can configure docker daemon to syslog/fluentd and only specify --log-opt tag=sometag or --log-opt env=some,tags,here w/o adding log-driver.
> 
> yes, you right. With docker daemon, env variables will be removed and instead a file could be referenced.

Hmm I'm not fully understanding the `env` and `labels` options from docker log options page, but really  `tag` is probably good enough for the better nameing/context info.

> Docker daemon change will cover more scope other than just configuring log-drivers, and this is something I am willing to avoid to ensure we can have the feature in place without making any system level changes.

The thing is that this is an ops area. worfklow-helper.sh should definitely setup docker with logging configured and `tink-worker` should probably only deal with adding context via `tag`. This might change in the future, but for *right now*, YAGNI applies.

mmlb:
The boots/osie changes are definitely needed, but tink-server/tink-worker should be left up to "ops" to configure and the "user doc" is good enough for now.

Issue #198, opened 2020-06-30
GlobalTimeout appears unused


grahamc:
Right now, the global timeout is not actually used except at display-time in the `portal` project:

```
[grahamc@Petunia:~/projects/github.com/tinkerbell]$ rg GlobalTimeout
tink/executor/types.go
7:	GlobalTimeout int    `yaml:"global_timeout"`

tink/db/workflow.go
29:		GlobalTimeout int    `yaml:"global_timeout"`

portal/src/client/workflow.go
123:	wf.Timeout = strconv.Itoa(details.GlobalTimeout)

portal/src/pkg/types/workflow.go
26:	GlobalTimeout int    `yaml:"global_timeout"`
```

I propose the following semantics. Given a workflow:

```yaml
global_timeout: 30
tasks:
- name: "example"
  worker: "{{.device_1}}"
  actions:
  - name: "phase-1"
    image: hello-world
    timeout: 20
    command: sleep 20
  - name: "phase-2"
    image: hello-world
    timeout: 20
    command: sleep 20
```

1. phase-1 would execute to completion, take 20 seconds, and the entire workflow would have 10 seconds to complete the rest of the phases.
2. phase-2 would begin, run for 10 seconds, be killed, and the workflow is canceled.

parauliya:
@grahamc I have a suggestion. Since we are parsing the yaml before creating a template. We can add the timeout of all the actions and then we can comapre it with global timeout. If it exceeds we can simply say that this template is invalid.

gianarb:
I like the idea to validate it at API layer

parauliya:
As discussed, There is no use of `global_timeout` as we have `action level timeout` so this needs to be removed from the template.

parauliya:
> As discussed, There is no use of `global_timeout` as we have `action level timeout` so this needs to be removed from the template and code itself.

There was a misunderstanding with the use of "global_timeout" but finally it was decided to have it in the template and add the support in the code.

gianarb:
I had a chat with a bunch of people and we agreed on how this feature will be implemented.

The `GlobalTimeout` is a workflow attribute that acts as a safeguard to guarantee that workflow will end. Tinkerbell can't get to a point where workflow runs forever. The global timeout needs to have a default value in tink-server (15 minutes), the value has to be configurable with a flag or env var when the tink-server starts.

Workflow supports a `global_timeout`, 0 is not a supported value.

gauravgahlot:
[...]

PR #269, opened 2020-08-27
Removed GlobalTimeout from the template definition

parauliya:

## Description

This PR will remove the global_timeout field from template which was unused.

## Why is this needed

Fixes: #198 

## How Has This Been Tested?
I ran an e2e test using vagrant before raising this PR.

Now the existing user need not to have` global_timeout` field in the template.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/tink-cli/sample.tmpl
+++ b/cmd/tink-cli/sample.tmpl
@@ -1,6 +1,5 @@
 version: '0.1'
 name: packet_osie_provision
-global_timeout: 600
 tasks:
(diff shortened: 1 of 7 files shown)


gianarb:
I think this comes from a misunderstanding we had during a meeting.
Right now we think a globalTimeout is needed and it should trigger an action. 

The globalTimeout acts as a safeguard to tell when a workflow HAS to complete. If it does not, it will be marked as a failure and never looked at it again.

PR #270, opened 2020-08-27
Fixed the duplicate name issue while creating a template

parauliya:
Name of the template which is used to provide with -n flag while creating a template should be unique.
But there was an issue because of which it was accepting duplicate name as well.
This PR fixed the above issue.

## Description

Name of the template which is used to provide with -n flag while creating a template should be unique.
But there was an issue because of which it was accepting duplicate name as well.
This PR fixed the above issue.

## Why is this needed
Same as Description.

## How Has This Been Tested?
I tried to create multiple template with same name but now they are not being created because of the same name.

User need to provide the unique name for each template it is creating in the database.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/grpc-server/template_test.go
+++ b/grpc-server/template_test.go
@@ -33,11 +33,12 @@ tasks:
       timeout: 60`
 )
 
-func TestDuplicateTemplateName(t *testing.T) {
+func TestCreateTemplate(t *testing.T) {
 	type (
 		args struct {
-			db   mock.DB
-			name string
+			db        mock.DB
+			name      []string
+			templates []string
 		}
 		want struct {
 			expectedError bool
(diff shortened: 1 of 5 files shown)


gianarb:
Pretty good! Can you add a Table Test that accepts templates as input and validate the answers, please?

Let's make this coverage green bit by bit!

grpc-server/template_test.go
```diff
@@ -0,0 +1,75 @@
...
+tasks:
+  - name: "hello world again"
+    worker: "{{.device_1}}"
+    actions:
+  	- name: "hello_world_again"
+      image: hello-world
+      timeout: 60`
+)
+
+func TestDuplicateTemplateName(t *testing.T) {
```
gianarb, line 36: Can you call it `TestCreateTemplate`? and the table test has to work like this imho:
Table Test is not only for one specific use case, but it has to be flexible enough to allow all the one you can think about for the function you are testing, in our case `CreateTemplate`

```golang
// you don't need a testCases struct just use a []testCase
struct testCase type {
    db mock.DB
    name string
    input []string // or templates
    expect func() (*template.CreateResponse, error)
}
```

In order to figure out if the TableTest is well written can you add three other testcases please?

1. Well formatted and successful template creation (only one)
2. A random validation error if we have any
3. Create 2 templates with a different name, both should work

Thanks

gianarb (changes requested):

grpc-server/template_test.go
parauliya: sure, I will try to add couple of more test cases as you suggested.

parauliya: @gianarb  As mentioned by you I have modified the test struct ( not exactly as you suggested ) so that in future more tests can be added. And I have also added couple of more test cases. But I was not able to test for validation of template because we don't check the validation of a template in `CreateTemplate` function. Template is being validated while creating a workflow here https://github.com/tinkerbell/tink/blob/master/db/workflow.go#L98.

gianarb (approved):
Great! Thanks `+1.74%` ✅

PR #272, opened 2020-08-28
add "tink docs" command for markdown and man page generation

displague:
## Description

This adds a `tink docs` command for generating both markdown and man pages.

This could be introduced in the `Makefile` as a way to ship and keep updated markdown docs in the repository, but I have not enabled that.

## Why is this needed

Documentation about the tink commands is currently hand generated. Cobra allows for these to be autogenerated.

Man pages will help system operators learn about the available options. Man pages are also a requirement for Debian packages, should we want to explore that.

## How Has This Been Tested?

```
make
mkdir docs
cmd/tink-cli/tink-cli docs markdown --path docs/
```

```
mkdir man
cmd/tink-cli/tink-cli docs man --path man/
man man/tink.1
```

none

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/cmd/tink-cli/cmd/completion.go
+++ b/cmd/tink-cli/cmd/completion.go
@@ -32,10 +32,9 @@ func completionCmd(name string) *cobra.Command {
 	# To load completions for each session, execute once:
 	$ tink-cli completion fish > ~/.config/fish/completions/tink-cli.fish
 	`,
-		DisableFlagsInUseLine: true,
-		Hidden:                true,
-		ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
-		Args:                  cobra.ExactValidArgs(1),
+		Hidden:    true,
+		ValidArgs: []string{"bash", "zsh", "fish", "powershell"},
+		Args:      cobra.ExactValidArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			switch args[0] {
 			case "bash":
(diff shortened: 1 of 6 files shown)


cmd/tink-cli/cmd/completion.go
```diff
@@ -32,10 +32,9 @@ func completionCmd(name string) *cobra.Command {
 	# To load completions for each session, execute once:
 	$ tink-cli completion fish > ~/.config/fish/completions/tink-cli.fish
 	`,
-		DisableFlagsInUseLine: true,
```
displague, line 35: Removing this flag so that the `completion` command is included in `--help`.

cmd/tink-cli/cmd/root.go
```diff
@@ -15,6 +15,7 @@ var rootCmd = &cobra.Command{
 	Use:               "tink",
 	Short:             "tinkerbell CLI",
 	PersistentPreRunE: setupClient,
+	DisableAutoGenTag: true,
```
displague, line 18: This prevents the markdown docs from create datestamp jitter, should we include and require updated docs/ in each commit.

thebsdbox:
This looks great, does it require any sort of testing?

displague:
@thebsdbox in an idea world, every line of code is tested. In this case, we have no tests for the `tink-cli` command or its subcommand.

This CLI is modeled after the Packet CLI which includes some E2E tests as examples if we want to require that for this merge. <https://github.com/packethost/packet-cli/tree/master/tests>

displague:
@thebsdbox Getting me to say the wrong thing is a sure way to get me to do the right thing.  Tests added.

thebsdbox (approved):
Awesome, this looks great :-)

PR #275, opened 2020-08-31
ENG-8624 fix operating_system_version/operating_system mismatch

kdeng3849:
Signed-off-by: Kelly Deng <kelly@packet.com>

## Description

This PR fixes the mismatch in field names [`operating_system_version`](https://github.com/tinkerbell/tink/blob/master/protos/packet/packet.proto#L86) vs [`operating_system`](https://github.com/tinkerbell/hegel/blob/a3138d417536903dcdedd674d634e13ebc19fc79/http_handlers.go#L33).

## Why is this needed
This bug mainly affects the users who needs both the `packet.pb.go` and the EC2 endpoint. Because the EC2 endpoint in Hegel uses the field name `operating_system` to be backwards compatible with Kant, users using the `Metadata` object defined in `packet.pb.go` will face a problem with parsing the operating system field since it was defined as `operating_system_version` there. 
The reason for the mismatch is because in the "old" cacher mode, `operating_system_version`, but Kant uses `operating_system` (which we have to be backwards compatible with) so we decided move away from `operating_system_version` and go with `operating_system` for "tink mode".
A user could have worked around this by specifying two fields `operating_system` and `operating_system_version` with the same value, inside the same piece of hardware, but this introduces some redundancy and just isn't ideal.

Related PR: https://github.com/tinkerbell/boots/pull/104

Fixes: #

## How Has This Been Tested?

To be manually tested

Existing users who import the `packet.pb.go` will have to `go get` latest master of the tink repo.

## Checklist:

I have:

- [ ] added unit or e2e tests


gianarb:
Is this a bc break? How does it impact users? Can we add a bit more context to this pull request in the description?
Thanks

displague:
@kdeng3849 Can you rebase this and update with the latest protos/protoc.sh (presumably this should result in a smaller diff, especially if the containerized protoc is used).

To @gianarb's point, can we offer both `operating_system` and `operating_system_version` to avoid a BC break?  Does this make sense in the long-run? (operating_system=debian, operating_system_version=9.0)

kdeng3849:
@displague I'm not sure if that makes sense because the `operating_system`/`operating_system_version` field is actually an [object](https://github.com/tinkerbell/tink/blob/master/protos/packet/packet.proto#L13) (not string and float, unless I misunderstood you there). 

I guess we _could_ offer both for backward compatibility, but I don't think many people were using the code generated from the packet proto to begin with? And I also think `operating_system` makes more sense compared to `operating_system_version` since with `version` in the name people might think it's actually referring to a version number, when it's not. Just feels unintuitive to me.

mmlb:
This isn't really a bc-break as packet.proto is not supported in the oss project. Its being used by one internal client that is currently broken and needs this PR so it works. Either way, packet.proto should be replaced and actually be usable by https://github.com/tinkerbell/proposals/pull/25

mmlb (approved):

mmlb (approved):

mmlb (approved):

PR #276, opened 2020-09-02
Fix Terraform setup

gianarb:
This is part of the work I am doing with the Terraform Setup guide:

https://github.com/tinkerbell/tinkerbell.org/pull/129


gianarb:
Great thanks @displague !

deploy/terraform/main.tf
```diff
@@ -10,43 +10,46 @@ provider "packet" {
...
 
   provisioner "file" {
     source      = "./../../../tink"
     destination = "/root/"
-
-    connection {
-      type        = "ssh"
-      user        = var.ssh_user
-      host        = packet_device.tink-provisioner.network[0].address
-      private_key = file(var.ssh_private_key)
```
displague: @gianarb I removed the `private_key` because Terraform will use your ssh-agent by default.

The problem with defaulting to "~/.ssh/id_rsa" is that this file is usually encrypted, and Terraform can not use that.

My suggestion is that we ask the user to add the key they want to use to their SSH agent before running `terraform apply`.  For most users, this should already be the case.

An alternative is to ask users to `ssh-keygen` create a new key in this directory, and tell Terraform to use that.

If we can make `private_key` optional (use the agent if ssh_private_key is empty) that would be even better.  I can look into this.

deploy/terraform/main.tf
```diff
@@ -10,43 +10,46 @@ provider "packet" {
...
   hostname         = "tink-provisioner"
   plan             = var.device_type
   facilities       = [var.facility]
   operating_system = "ubuntu_18_04"
   billing_cycle    = "hourly"
   project_id       = var.project_id
   user_data        = file("install_package.sh")
+}
+
+resource "null_resource" "tink_directory" {
```
displague, line 30: @gianarb by moving the provisioning to a `null_resource`, provisioning failures (like a missing ssh key) will not require the `tink_provisioner` host to be recreated.  Changes to the tink/ directory can be reapplied (rsync'd) independently. `terraform taint null_resource.tink_directory; terraform apply`

deploy/terraform/main.tf
```diff
@@ -56,21 +59,25 @@ resource "packet_device" "tink-worker" {
...
+  count = var.worker_count
+
+  device_id = packet_device.tink_worker[count.index].id
   type      = "layer2-individual"
 }
 
 # Attach VLAN to provisioner
 resource "packet_port_vlan_attachment" "provisioner" {
-  device_id = packet_device.tink-provisioner.id
+  device_id = packet_device.tink_provisioner.id
```
displague, line 71: I ran into `Error: POST https://api.packet.net/ports/bff62cf9-5f23-49fd-b500-686e3d8730b6/assign: 422 still bonded` once while testing this.

It may make sense to add `depends_on = packet_device_network_type.tink_provisioner_network_type` here 🤔

deploy/terraform/variables.tf
```diff
@@ -8,6 +8,11 @@ variable "project_id" {
   type        = string
 }
 
+variable "worker_count" {
```
displague, line 11: I haven't checked https://github.com/tinkerbell/tinkerbell.org/pull/129 to see if multiple workers will require any additional doc changes.

deploy/terraform/main.tf
gianarb: It sounds like a reasonable dependency to me, so let's be clear about that and write it down 👍

deploy/terraform/variables.tf
gianarb: No, they will just work! Nice!

rugwirobaker:
I'm quite new to terraform, and for it to work for me I have had to `terraform 0.13upgrade .` And I'm currently testing with this PR. 
I'm using the latest terraform version.
```
➜  ~ terraform version
Your version of Terraform is out of date! The latest version
is 0.13.2. You can update by downloading from https://www.terraform.io/downloads.html
Terraform v0.13.0
```
Also something that would  make the process faster for the newbies is to include the file `my_var.tfvars` file mentioned in the docs(I did have to learn terraform so, I guess that's a positive :satisfied:). I'm thinking of something like this:
```
deploy/
├── example.tfvars 
├── install_package.sh
├── my_var.tfvars // user created copy should be added to .gitignore 
├── main.tf
├── outputs.tf
└── variables.tf

```
Example: https://github.com/alexellis/sshdkit where the ssh.yml was ignored but we have an example for it in ssh.example.yml
Small thing but it would make the project a little bit more accessible for people like me.

rugwirobaker:
A small hicup running `./setup.sh`
Got this error on my first run of terraform apply so I redid it and it went away
```
Error: POST https://api.packet.net/ports/d6ddf121-3f74-4b43-9a6c-a25edfc40fa5/assign: 422 still bonded 

  on main.tf line 76, in resource "packet_port_vlan_attachment" "provisioner":
  76: resource "packet_port_vlan_attachment" "provisioner" {

Error: POST https://api.packet.net/ports/e16fe28f-e194-409f-9d06-c7c72d07f7b4/assign: 422 still bonded 

  on main.tf line 83, in resource "packet_port_vlan_attachment" "worker":
  83: resource "packet_port_vlan_attachment" "worker" {
```

But ounce insider the provisioner the `setup.sh` script failed:
```
INFO: verifying prerequisites for ubuntu (18.04)                                                                                   [51/719]
       Found prerequisite: docker                                                                                                          
       Found prerequisite: docker-compose                                                                                                  
       Found prerequisite: ip                                                                                                              
       Found prerequisite: jq                                                                                                              
       Found prerequisite: netplan                                                                                                         
INFO: waiting for the network configuration to be applied by systemd-networkd                                                              
.........

Step 7/7 : EXPOSE 443
 ---> Running in b68d63577cca
Removing intermediate container b68d63577cca
 ---> 9d1bdfff2b16
Successfully built 9d1bdfff2b16
Successfully tagged deploy_registry:latest
Recreating deploy_registry_1 ... 
Recreating deploy_registry_1 ... done
Error response from daemon: Get https://192.168.1.1/v2/: net/http: request canceled while waiting for connection (Client.Timeout exceeded w
hile awaiting headers)  
```

I went on but `docker-compose up -d` could not bring up the nginx container with the following error:
```
ERROR: for deploy_nginx_1  Cannot start service nginx: driver failed programming external connectivity on endpoint deploy_nginx_1 (7ccfa729
954ae70036255926ac45970b2ddfc6797f7d3b2bf6924ea10b3c4f22): Error starting userland proxy: listen tcp 192.168.1.2:80: bind: cannot assign
````

deploy/terraform/main.tf
gianarb: I added the `ssh-agent` note as part of the documentation. We can do the fallback work as follow up work 👍

gianarb:
The support for multiple worker is good but it is not ideal (there are a lot of index 1 fixed around), should we hold it for another PR?

displague (approved):

Issue #278, opened 2020-09-03
force / store mac lowercase


geekgonecrazy:
I created a hardware definition copying directly from proxmox which shows mac address all capital letters.

So when boots queried using the lowercase mac address it would return "unexpected end of json"

## Expected Behaviour

Force / store mac address lowercase

## Current Behaviour

Errors out

## Possible Solution

Force lowercase

## Steps to Reproduce (for bugs)

1. insert
```
{"id":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","metadata":{"facility":{"facility_code":"ewr1","plan_slug":"c2.medium.x86","plan_version_slug":""},"instance":{},"state":"provisioning"},"network":{"interfaces":[{"dhcp":{"arch":"x86_64","ip":{"address":"10.10.6.2","gateway":"10.10.0.1","netmask":"255.255.0.0"},"mac":"E2:8A:5B:5E:45:3E"},"netboot":{"allow_pxe":true,"allow_workflow":true}}]}}
```
2.
boot vm or a machine with mac address.  Which will be sent to lowercase
3. fails to boot

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

detiber:
I created https://github.com/tinkerbell/tink/pull/279, which should address this issue

geekgonecrazy:
Wow that was quick! 🙏

PR #279, opened 2020-09-04
Store MAC as lowercase

detiber:
## Description

Always stores any MAC addresses provided as lowercase to avoid issues with comparisons later on.

## Why is this needed

boots is attempting to compare pxe booted systems using a lowercased MAC address, so if the user inputs the MAC addresses that are capitalized they will not match.

Fixes: #278 

## How Has This Been Tested?
Still in progress, will be testing against the Vagrant getting started guide, and manually changing the MAC used to be uppercased.

This should not effect existing users (at least ones that have functioning environments), since existing hardware definitions using upper cased MAC addresses would be failing already.

## Checklist:

I have:

- [ ] added unit or e2e tests


grpc-server/hardware.go
```diff
@@ -316,3 +320,14 @@ func (s *server) validateHardwareData(ctx context.Context, hw *hardware.Hardware
 	}
 	return nil
 }
+
+func (s *server) normalizeHardwareData(hw *hardware.Hardware) {
+	// Ensure MAC is stored as lowercase
+	if hw != nil && hw.Network != nil {
+		for i := range hw.Network.Interfaces {
+			if hw.Network.Interfaces[i].Dhcp != nil {
```
displague, line 328: `hw.Network.Interfaces[i]` could technically be `nil` too.

displague (dismissed):
I have a nit-pic about potential nil, otherwise this looks good to me.

grpc-server/hardware.go
```diff
@@ -316,3 +320,14 @@ func (s *server) validateHardwareData(ctx context.Context, hw *hardware.Hardware
 	}
 	return nil
 }
+
+func (s *server) normalizeHardwareData(hw *hardware.Hardware) {
```
displague, line 324: Do we want to make this a method? I was going to suggest/offer a test for this function and realized that the harness would include all the complexities of mocking a `server`.

displague:
```patch
diff --git a/grpc-server/hardware.go b/grpc-server/hardware.go
index dffd492..1527439 100644
--- a/grpc-server/hardware.go
+++ b/grpc-server/hardware.go
@@ -40,7 +40,7 @@ func (s *server) Push(ctx context.Context, in *hardware.PushRequest) (*hardware.
 	}
 
 	// normalize data prior to storing in the database
-	s.normalizeHardwareData(hw)
+	normalizeHardwareData(hw)
 
 	// validate the hardware data to avoid duplicate mac address
 	err := s.validateHardwareData(ctx, hw)
@@ -321,7 +321,7 @@ func (s *server) validateHardwareData(ctx context.Context, hw *hardware.Hardware
 	return nil
 }
 
-func (s *server) normalizeHardwareData(hw *hardware.Hardware) {
+func normalizeHardwareData(hw *hardware.Hardware) {
 	// Ensure MAC is stored as lowercase
 	if hw != nil && hw.Network != nil {
 		for i := range hw.Network.Interfaces {
diff --git a/grpc-server/hardware_test.go b/grpc-server/hardware_test.go
new file mode 100644
index 0000000..01d3a6d
--- /dev/null
+++ b/grpc-server/hardware_test.go
@@ -0,0 +1,49 @@
+package grpcserver
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_normalizeHardwareData(t *testing.T) {
+	type args struct {
+		hw *hardware.Hardware
+	}
+	tests := []struct {
+		name string
+		got  args
+		want args
+	}{
+		{
+			name: "NilHardware",
+			got:  args{hw: nil},
+		},
+		{
+			name: "Uppercase",
+			got: args{hw: &hardware.Hardware{
+				Network: &hardware.Hardware_Network{
+					Interfaces: []*hardware.Hardware_Network_Interface{{
+						Dhcp: &hardware.Hardware_DHCP{Mac: "DE:AD:BE:EF"},
+					}},
+				},
+			},
+			},
+			want: args{hw: &hardware.Hardware{
+				Network: &hardware.Hardware_Network{
+					Interfaces: []*hardware.Hardware_Network_Interface{{
+						Dhcp: &hardware.Hardware_DHCP{Mac: "de:ad:be:ef"},
+					}},
+				},
+			},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			normalizeHardwareData(tt.got.hw)
+			assert.Equal(t, tt.want.hw, tt.got.hw)
+		})
+	}
+}
```

detiber: Indeed, already switched it to a standalone function as part of creating the tests

grpc-server/hardware.go
```diff
@@ -297,22 +309,35 @@ func (s *server) Delete(ctx context.Context, in *hardware.DeleteRequest) (*hardw
 }
 
 func (s *server) validateHardwareData(ctx context.Context, hw *hardware.Hardware) error {
-	interfaces := hw.GetNetwork().GetInterfaces()
-	for i := range hw.GetNetwork().GetInterfaces() {
-		data, _ := s.db.GetByMAC(ctx, interfaces[i].GetDhcp().GetMac())
-		if data != "" {
-			logger.With("MAC", interfaces[i].GetDhcp().GetMac()).Info(duplicateMAC)
+	for _, iface := range hw.GetNetwork().GetInterfaces() {
+		mac := iface.GetDhcp().GetMac()
```
detiber, line 313: I'm wondering if it would make sense to add a test for the empty string here prior to calling `s.db.GetByMAC()`.

Would we expect that to be an error condition or would we allow a hardware definition without a mac address?

grpc-server/hardware.go
```diff
@@ -297,22 +309,35 @@ func (s *server) Delete(ctx context.Context, in *hardware.DeleteRequest) (*hardw
...
+			return errors.New(fmt.Sprintf(conflictMACAddr, mac))
 		}
 	}
+
 	return nil
 }
+
+func normalizeHardwareData(hw *hardware.Hardware) {
+	// Ensure MAC is stored as lowercase
+	for _, iface := range hw.GetNetwork().GetInterfaces() {
```
detiber, line 337: I switched things to use the generated proto getters rather than nil checks, since they already include the nil checks and it also eliminates the use of slice indexing.

grpc-server/hardware.go
displague: I was questioning similar, whether we should transform the input at all, just let validate/push throw an error if the hardware mac was not "valid" (lowercase, non-empty).

That said, I don't know hardware address can be treated as optional.

displague: I worry that transforming the data means that any stateful tools that read back the lowercase value will need to be aware of the transform and copy that logic themselves.

grpc-server/hardware.go
```diff
@@ -29,15 +30,26 @@ func (s *server) Push(ctx context.Context, in *hardware.PushRequest) (*hardware.
 	// must be a copy so deferred cacheInFlight.Dec matches the Inc
 	labels = prometheus.Labels{"method": "Push", "op": ""}
 
-	hw := in.Data
-	if hw.Id == "" {
+	hw := in.GetData()
+	if hw != nil {
```
displague, line 34: inverted logic

grpc-server/hardware.go
```diff
@@ -297,22 +309,35 @@ func (s *server) Delete(ctx context.Context, in *hardware.DeleteRequest) (*hardw
...
 				logger.Error(err, "Failed to unmarshal hardware data")
 				return err
 			}
+
 			if newhw.Id == hw.Id {
 				return nil
 			}
-			return errors.New(fmt.Sprintf(conflictMACAddr, interfaces[i].GetDhcp().GetMac()))
+
+			return errors.New(fmt.Sprintf(conflictMACAddr, mac))
```
displague, line 328: `return fmt.Errorf(...)` ?

grpc-server/hardware_test.go
```diff
@@ -0,0 +1,179 @@
...
+							Dhcp: &hardware.Hardware_DHCP{
+								Mac: "02:42:0e:d9:c7:53",
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "Expect MAC to be normalized to lowercase, nultiple interfaces",
```
displague, line 42: nultiple

grpc-server/hardware.go
detiber: The alternative I thought of was to make the comparisons case-insensitive, but with it being stored embedded inside of a stored json blob that seemed a bit more error prone.

grpc-server/hardware.go
detiber: Fixed, thanks for catching that.

gianarb (dismissed):
Great! Thanks

displague (approved):

PR #280, opened 2020-09-07
Added the validation of template while creating it

parauliya:
## Description

This PR will validate a template. If `timeout` field for an action is not provided or provided with 0 value, then that will be considered as infinite timeout for that particular action.

## Why is this needed
This fix is required so that Provisioner should only depends on global timeout.

## How Has This Been Tested?
Added UT for the `CreateTemplate` API.
Also tested through e2e test to ensure that any existing feature doesn't break.

No Impact on existig user

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/cmd/tink-worker/action.go
+++ b/cmd/tink-worker/action.go
@@ -43,15 +43,13 @@ func executeAction(ctx context.Context, action *pb.WorkflowAction, wfID string) 
 		return pb.ActionState_ACTION_IN_PROGRESS, errors.Wrap(err, "DOCKER CREATE")
 	}
 	l.With("containerID", id, "command", action.GetOnTimeout()).Info("container created")
-
-	var timeCtx context.Context
-	var cancel context.CancelFunc
+	// Setting time context for action
+	timeCtx := ctx
 	if action.Timeout > 0 {
-		timeCtx, cancel = context.WithTimeout(context.Background(), time.Duration(action.Timeout)*time.Second)
-	} else {
-		timeCtx, cancel = context.WithTimeout(context.Background(), 1*time.Hour)
+		var cancel context.CancelFunc
+		timeCtx, cancel = context.WithTimeout(ctx, time.Duration(action.Timeout)*time.Second)
+		defer cancel()
 	}
-	defer cancel()
 	err = startContainer(timeCtx, l, id)
 	if err != nil {
 		return pb.ActionState_ACTION_IN_PROGRESS, errors.Wrap(err, "DOCKER RUN")
--- a/db/mock/template.go
+++ b/db/mock/template.go
@@ -8,30 +8,26 @@ import (
 	"github.com/google/uuid"
 )
 
-type template struct {
-	id   uuid.UUID
-	data string
+type Template struct {
+	ID   uuid.UUID
+	Data string
 }
-
-var templateDB = map[string]interface{}{}
 
 // CreateTemplate creates a new workflow template
 func (d DB) CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error {
-	if len(templateDB) > 0 {
-		if _, ok := templateDB[name]; ok {
-			return errors.New("Template name already exist in the database")
-		}
-		templateDB[name] = template{
-			id:   id,
-			data: data,
-		}
-
-	} else {
-		templateDB[name] = template{
-			id:   id,
-			data: data,
-		}
+	if d.TemplateDB == nil {
+		d.TemplateDB = make(map[string]interface{})
 	}
+
+	if _, ok := d.TemplateDB[name]; ok {
+		return errors.New("Template name already exist in the database")
+	}
+
+	d.TemplateDB[name] = Template{
+		ID:   id,
+		Data: data,
+	}
+
 	return nil
 }
 
@@ -42,12 +38,10 @@ func (d DB) GetTemplate(ctx context.Context, id string) (string, string, error) 
 
 // DeleteTemplate deletes a workflow template
 func (d DB) DeleteTemplate(ctx context.Context, name string) error {
-	if len(templateDB) > 0 {
-		if _, ok := templateDB[name]; !ok {
-			return errors.New("Template name does not exist")
-		}
-		delete(templateDB, name)
+	if d.TemplateDB != nil {
+		delete(d.TemplateDB, name)
 	}
+
 	return nil
 }
 
@@ -63,7 +57,5 @@ func (d DB) UpdateTemplate(ctx context.Context, name string, data string, id uui
 
 // ClearTemplateDB clear all the templates
 func (d DB) ClearTemplateDB() {
-	for name := range templateDB {
-		delete(templateDB, name)
-	}
+	d.TemplateDB = make(map[string]interface{})
 }
--- a/db/template.go
+++ b/db/template.go
@@ -9,6 +9,7 @@ import (
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/google/uuid"
 	"github.com/pkg/errors"
+	"github.com/tinkerbell/tink/pkg"
 )
 
 // CreateTemplate creates a new workflow template
@@ -16,6 +17,15 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 	tx, err := d.instance.BeginTx(ctx, &sql.TxOptions{Isolation: sql.LevelSerializable})
 	if err != nil {
 		return errors.Wrap(err, "BEGIN transaction")
+	}
+
+	wf, err := pkg.ParseYAML([]byte(data))
+	if err != nil {
+		return err
+	}
+	err = pkg.ValidateTemplate(wf)
+	if err != nil {
+		return err
 	}
 
 	_, err = tx.Exec(`
--- a/grpc-server/template_test.go
+++ b/grpc-server/template_test.go
@@ -22,13 +22,13 @@ tasks:
       timeout: 60`
 
 	template2 = `version: "0.1"
-name: hello_world_workflow
+name: hello_world_again_workflow
 global_timeout: 600
 tasks:
   - name: "hello world again"
-    worker: "{{.device_1}}"
+    worker: "{{.device_2}}"
     actions:
-  	- name: "hello_world_again"
+    - name: "hello_world_again"
       image: hello-world
       timeout: 60`
 )
@@ -36,9 +36,9 @@ tasks:
 func TestCreateTemplate(t *testing.T) {
 	type (
 		args struct {
-			db        mock.DB
-			name      []string
-			templates []string
+			db       mock.DB
+			name     string
+			template string
 		}
 		want struct {
 			expectedError bool
@@ -50,9 +50,11 @@ func TestCreateTemplate(t *testing.T) {
 	}{
 		"SuccessfullTemplateCreation": {
 			args: args{
-				db:        mock.DB{},
-				name:      []string{"template_1"},
-				templates: []string{template1},
+				db: mock.DB{
+					TemplateDB: make(map[string]interface{}),
+				},
+				name:     "template_1",
+				template: template1,
 			},
 			want: want{
 				expectedError: false,
@@ -61,9 +63,13 @@ func TestCreateTemplate(t *testing.T) {
 
 		"SuccessfullMultipleTemplateCreation": {
 			args: args{
-				db:        mock.DB{},
-				name:      []string{"template_1", "template_2"},
-				templates: []string{template1, template2},
+				db: mock.DB{
+					TemplateDB: map[string]interface{}{
+						"template_1": template1,
+					},
+				},
+				name:     "template_2",
+				template: template2,
 			},
 			want: want{
 				expectedError: false,
@@ -72,35 +78,31 @@ func TestCreateTemplate(t *testing.T) {
 
 		"FailedMultipleTemplateCreationWithSameName": {
 			args: args{
-				db:        mock.DB{},
-				name:      []string{"template_1", "template_1"},
-				templates: []string{template1, template2},
+				db: mock.DB{
+					TemplateDB: map[string]interface{}{
+						"template_1": template1,
+					},
+				},
+				name:     "template_1",
+				template: template2,
 			},
 			want: want{
 				expectedError: true,
 			},
 		},
 	}
-	for name, tc := range testCases {
+
+	for name := range testCases {
+		tc := testCases[name]
 		t.Run(name, func(t *testing.T) {
+			t.Parallel()
 			s := testServer(tc.args.db)
-			tc.args.db.ClearTemplateDB()
-			index := 0
-			res, err := s.CreateTemplate(context.TODO(), &pb.WorkflowTemplate{Name: tc.args.name[index], Data: tc.args.templates[index]})
-			assert.Nil(t, err)
-			assert.NotNil(t, res)
-			if err == nil && len(tc.args.templates) > 1 {
-				index++
-				_, err = s.CreateTemplate(context.TODO(), &pb.WorkflowTemplate{Name: tc.args.name[index], Data: tc.args.templates[index]})
-			} else {
-				return
-			}
-			if err != nil {
+			res, err := s.CreateTemplate(context.TODO(), &pb.WorkflowTemplate{Name: tc.args.name, Data: tc.args.template})
+			if tc.want.expectedError {
 				assert.Error(t, err)
-				assert.True(t, tc.want.expectedError)
 			} else {
 				assert.Nil(t, err)
-				assert.False(t, tc.want.expectedError)
+				assert.NotEmpty(t, res)
 			}
 		})
 	}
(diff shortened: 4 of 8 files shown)


gianarb:
That's nice. A couple of questions:

1. Is it better to have a default vs erroring when it is not passed?
2. this is a bc break, people using this code but with a workflow that does not have a timeout needs to know :) https://github.com/tinkerbell/proposals/pull/15

gauravgahlot:
@parauliya I remember you mentioned about having a default value for `timeout`, but I missed the bc break that time. You were right then. 

@gianarb IMO it would be good to have a default, but we still want to make them aware of the new default. Because it is possible that the new default introduces some unexpected behavior. 

Now, the question is what is the ideal (kinda) default timeout?

gianarb:
You are right @gauravgahlot . Does not matter the decision we take, it is a bc break. I pinged @thebsdbox to figure out what we want because I don't really have a preference. Probably a default timeout applies here, but I am not strong about it

parauliya:
@gianarb @gauravgahlot IMO, Since this is a `action level timeout`, if we have a default value for each action then there are following issues which can be occur:
* There is a possibility the default value is much more than execution time for few actions, in that case, global timeout will be difficult to handle since we will be calculating the total time with the default values which can be more than global timeout provided in the template. And in that case, template will be invalid.
* The reverse of above is also possible where the time taken by an action is much more than the default value and in that case, the action will timeout.

Due to the above points I think it is better to make this field mandatory and user should provide this as a part of template. 
For bc break I think we should think for some different approach.

If we talk about workflow timeout which is global timeout, this can have a default value which will be calculated dynamically by adding the timeout for all the actions and few additional seconds for communication over the network.

pkg/template_validator.go
```diff
@@ -77,3 +82,10 @@ func isValidImageName(name string) error {
 	}
 	return nil
 }
+
+func hasValidTimeout(timeout int64, name string) error {
+	if timeout > 0 {
+		return nil
+	}
+	return fmt.Errorf(errNoTimeout, name)
```
gauravgahlot, line 90: Could you please change all calls of `fmt.Errorf` to `errors.Wrapf` instead?

gauravgahlot (changes requested):

protos/workflow/workflow.pb.go
```diff
@@ -8,6 +8,9 @@ package workflow
 
 import (
 	context "context"
+	reflect "reflect"
```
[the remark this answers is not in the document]
parauliya: These changes have come when I ran the command `git rebase -x 'goimports -w .' master` as @mmlb told you to run it before raising a PR.

pkg/template_validator_test.go
```diff
@@ -45,6 +55,11 @@ func TestParseYAML(t *testing.T) {
...
 			expectedError: true,
 		},
+		{
+			name:          "without action timeout template",
+			content:       []byte(withoutTimeoutTemplate),
+			expectedError: true,
+		},
 	}
 
 	for _, test := range testcases {
```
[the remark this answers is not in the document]
parauliya: Done

gauravgahlot:
> @gianarb @gauravgahlot IMO, Since this is a `action level timeout`, if we have a default value for each action then there are following issues which can occur:
> 
> * There is a possibility the default value is much more than execution time for few actions, in that case, global timeout will be difficult to handle since we will be calculating the total time with the default values which can be more than global timeout provided in the template. And in that case, template will be invalid.
> * The reverse of above is also possible where the time taken by an action is much more than the default value and in that case, the action will timeout.
> 
> Due to the above points I think it is better to make this field mandatory and user should provide this as a part of template.
> For bc break I think we should think for some different approach.
> 
> If we talk about workflow timeout which is global timeout, this can have a default value which will be calculated dynamically by adding the timeout for all the actions and few additional seconds for communication over the network.

Like @gianarb I don't have a preference as well. 
As much as I would like the user to decide the timeout for each action and be flagged with an error when they don't, I don't want the current users to be confused as their templates fail to parse. 

The point is what's more important here is to look for an efficient answer for _How do we convey a bc break (like this one) to our existing users?_

gianarb:
> How do we convey a bc break (like this one) to our existing users?

I linked a proposal about that

db/mock/template.go
```diff
@@ -32,6 +34,17 @@ func (d DB) CreateTemplate(ctx context.Context, name string, data string, id uui
 			data: data,
 		}
 	}
+	wf, err := pkg.ParseYAML([]byte(data))
+	if err != nil {
+		fmt.Println("Could not parse Yaml")
```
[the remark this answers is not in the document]
parauliya: Done

pkg/template_validator_test.go
```diff
@@ -45,6 +55,11 @@ func TestParseYAML(t *testing.T) {
 			content:       []byte(invalidTemplate),
 			expectedError: true,
 		},
+		{
```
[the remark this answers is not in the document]
parauliya: Yeah, I think so

pkg/template_validator_test.go
```diff
@@ -142,6 +162,10 @@ func withDuplicateActionName() workflowModifier {
 	return func(wf *Workflow) { wf.Tasks[0].Actions = append(wf.Tasks[0].Actions, wf.Tasks[0].Actions[0]) }
 }
 
+func withoutTimeoutInAction() workflowModifier {
```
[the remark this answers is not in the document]
parauliya: Done

pkg/template_validator.go
parauliya: `errors.Wrapf` required `err` as the first argument which we do not have in every case.

mmlb: Actually use the correct function. If no formatting is happening, use `errors.Wrap`, if there's no error available to wrap use `errors.New`

pkg/template_validator.go
```diff
@@ -1,18 +1,19 @@
 package pkg
 
 import (
-	"fmt"
+	"errors"
```
mmlb, line 5: this should not be grouped separately

pkg/template_validator.go
```diff
@@ -77,3 +82,10 @@ func isValidImageName(name string) error {
 	}
 	return nil
 }
+
+func hasValidTimeout(timeout int64, name string) error {
```
mmlb, line 86: this is an abuse of errors, should really be a bool. But its matching other similar functions so I'm torn :D. *If* you were to change the other `has.* error {` functions to return bool instead would be great :D. If not please open up an issue for it.

mmlb (changes requested):

gianarb:
@mmlb told me that internally Packet does not have a concept of per action timeout, but only a global one. And now I have big questions about life.

nathangoulding:
@mmlb is right that Packet only has a global timeout. Making per-action timeouts mandatory seems overly burdensome for the end user.

IMHO, we need a global timeout for the workflow, an optional default timeout that applies to all actions, and (if we're feeling like it's necessary) per-action timeouts that can override the default timeout. I don't think that per-action timeouts should be mandatory.

parauliya:
@nathangoulding Now the question is what should be the `default` value of `timeout` for actions?

pkg/template_validator.go
gauravgahlot: PR submitted #284 
I think we need to be careful about such convention.

gianarb:
I had a chat with @thebsdbox and @nathangoulding to build some context around this feature.

Packet does not currently have a per-action timeout and it does not really need it, not sure why Tinkerbell has this feature BUT let's implement it in this way.

No validation required, by default the pre-action timeout is set to 0 and it means `forever`, so no-timeout.
Can you implement the feature in this way @parauliya ?
Thanks

parauliya:
> means

Sure, I am on it.

cmd/tink-worker/action.go
```diff
@@ -39,10 +39,10 @@ func executeAction(ctx context.Context, action *pb.WorkflowAction, wfID string)
 	var cancel context.CancelFunc
```
gianarb, line 39:
```
       timeCtx = ctx
	if action.Timeout > 0 {
                var cancel context.CancelFunc
		timeCtx, cancel = context.WithTimeout(context.Background(), time.Duration(action.Timeout)*time.Second)
		defer cancel()
	}
```
Something like this sounds a bit easier to read according to me.

parauliya: Done

gianarb:
This needs a rebase @parauliya !

gianarb (approved):

gianarb (approved):

grpc-server/template_test.go
```diff
@@ -84,22 +158,24 @@ func TestCreateTemplate(t *testing.T) {
 	for name, tc := range testCases {
 		t.Run(name, func(t *testing.T) {
 			s := testServer(tc.args.db)
```
gianarb, line 160: Let's do a test, do you mind to add at line 160:

`t.Parallel()`

Thanks

gianarb (changes requested):

db/template.go
```diff
@@ -18,6 +19,15 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 		return errors.Wrap(err, "BEGIN transaction")
 	}
 
+	wf, err := pkg.ParseYAML([]byte(data))
```
mmlb, line 22: We should have a `tempate` package with a Parse method, not `pkg`. But this can be a separate issue/PR (#300)

grpc-server/template_test.go
```diff
@@ -22,17 +25,31 @@ tasks:
...
+	data string
+}
+
+var templateDB = map[string]interface{}{}
+
+// ClearTemplateDB clear all the templates
+func clearTemplateDB() {
+	for name := range templateDB {
+		delete(templateDB, name)
+	}
```
mmlb, line 50: can't we just do `templateDB = map[string]interface{}{}`?

grpc-server/template_test.go
```diff
@@ -50,7 +67,26 @@ func TestCreateTemplate(t *testing.T) {
 	}{
 		"SuccessfullTemplateCreation": {
 			args: args{
-				db:        mock.DB{},
+				db: mock.DB{
+					CreateTemplateFunc: func(ctx context.Context, name string, data string, id uuid.UUID) error {
+						if len(templateDB) > 0 {
```
mmlb, line 72: isn't len(templateDB) == 0 always because `clearTemplateDB` is being called?

grpc-server/template_test.go
mmlb: I don't think so, because clearTemplateDB is being called in each test run, which is a global var.

mmlb (changes requested):

grpc-server/template_test.go
```diff
@@ -50,7 +67,26 @@ func TestCreateTemplate(t *testing.T) {
 	}{
 		"SuccessfullTemplateCreation": {
 			args: args{
-				db:        mock.DB{},
+				db: mock.DB{
```
detiber, line 70: Why the switch from `mockDB{}` to overriding `CreateTemplateFunc`? From what I can tell the expected behavior appears to be the same.

grpc-server/template_test.go
gianarb: That's why I left a comment. We should not do that

grpc-server/template_test.go
parauliya: Yes, we can do that. Thanks

grpc-server/template_test.go
parauliya: Yes, we can remove that check in the test case where only one template is being created.

cmd/tink-worker/action.go
```diff
@@ -43,15 +43,13 @@ func executeAction(ctx context.Context, action *pb.WorkflowAction, wfID string)
...
 	}
 	l.With("containerID", id, "command", action.GetOnTimeout()).Info("container created")
-
-	var timeCtx context.Context
-	var cancel context.CancelFunc
+	// Setting time context for action
+	timeCtx := ctx
 	if action.Timeout > 0 {
+		var cancel context.CancelFunc
 		timeCtx, cancel = context.WithTimeout(context.Background(), time.Duration(action.Timeout)*time.Second)
```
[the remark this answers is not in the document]
parauliya: First of all, the global timeout will be controlled from the `tink-server` which is on the provisioner machine so we won't be passing the global timeout here. 
Second if a user has provided a timeout for a particular action I would like to give that action a separate context with that deadine so that in future we can run actions(with timeout) in parllel. That's why I have not used parent context here. But yes if user has not provided the timeout for an action it should run in a parent context which is as  you mentioned `ctx`.

grpc-server/template_test.go
parauliya: No, Not exactly, the definition of `CreateTemplateFunc` will depend on the test case I would like to test.

grpc-server/template_test.go
mmlb: Ah, yes I agree. {Less,No} global vars :allthethings:!

cmd/tink-worker/action.go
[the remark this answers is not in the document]
mmlb: Agree with @detiber. If (when, I have a proposal to write for this today) we have concurrent actions we can avoid the parent context, but for here/now lets continue with using the parent.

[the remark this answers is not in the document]
parauliya: Sure, for now I can change it to parent context for now and in future when we have concurrent actions this needs to be changed again.

gianarb:
@parauliya you should pull your changes and sign the merged commit otherwise CTO won't pass and I can't merge this PR

grpc-server/template_test.go
parauliya: Done

grpc-server/template_test.go
parauliya: Done

cmd/tink-worker/action.go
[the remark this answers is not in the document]
parauliya: Done

grpc-server/template_test.go
parauliya: Done

gianarb (dismissed):

gianarb (approved):

pkg/template_validator.go
```diff
@@ -30,50 +30,49 @@ func ParseYAML(yamlContent []byte) (*Workflow, error) {
 func ValidateTemplate(wf *Workflow) error {
 	taskNameMap := make(map[string]struct{})
 	for _, task := range wf.Tasks {
-		err := hasValidLength(task.Name)
-		if err != nil {
-			return err
+		if hasEmptyName(task.Name) {
+			return errors.New(errEmptyName + "task name")
```
gauravgahlot, line 34:
```suggestion
			return errors.New(errEmptyName + task.Name)
```

pkg/template_validator.go
```diff
@@ -30,50 +30,49 @@ func ParseYAML(yamlContent []byte) (*Workflow, error) {
...
+			return errors.New(errDuplicateTaskName + task.Name)
 		}
 		taskNameMap[task.Name] = struct{}{}
 		actionNameMap := make(map[string]struct{})
 		for _, action := range task.Actions {
-			err := hasValidLength(action.Name)
-			if err != nil {
-				return err
+			if hasEmptyName(action.Name) {
+				return errors.New(errEmptyName + "action name")
```
gauravgahlot, line 47:
```suggestion
				return errors.New(errEmptyName + action.Name)
```

gauravgahlot (changes requested):

gauravgahlot (changes requested):
@parauliya Please clean up the commits.

parauliya:
> @parauliya Please clean up the commits.

Done

parauliya:
I have rephrashed the two commit messages as they were contradictory.

gauravgahlot (approved):

detiber (approved):
Changes look good to me

gianarb (dismissed):

pkg/template_validator.go
```diff
@@ -1,18 +1,18 @@
 package pkg
 
 import (
-	"fmt"
+	"errors"
```
mmlb, line 4: this should be github.com/pkg/errors

mmlb (changes requested):

pkg/template_validator.go
parauliya: Done

gianarb (approved):

mmlb (dismissed):
I'd like to see the commit split into at least 2 commits, but not enough to actually block the PR on it.

gianarb (dismissed):

parauliya:
@gianarb @mmlb @detiber @gauravgahlot, Now the code cove report seems good. 
Please approve the same.

gauravgahlot (approved):

gianarb (approved):

detiber (approved):

Issue #282, opened 2020-09-08
Tink-cli creates workflow with the template Id which doesnot exists


Cbkhare:
Tink-cli creates workflow with the template Id which does not exists

## Expected Behaviour

tink-cli should not allow creating the workflow with below command when the UUID corresponding to the template does not exists. 
```
docker exec -i deploy_tink-cli_1 tink workflow create \
    -t  26da8139-5079-4f6a-8519-2a09feb12338\
    -r '{"device_1":"08:00:27:00:00:01"}'
```

## Current Behaviour
Tink-cli allows creating the workflow with the above command. This throws an error when the worker with the same hardware Id is started. The error is also inconsistent to the point where it actually occurred.

```
ERROR: 2020/09/08 13:10:39 grpc: server failed to encode response: rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil
```

## Possible Solution
Add a code change to in tink-cli to check if the UUID against a template is valid and if it exists. 

## Steps to Reproduce (for bugs)
1. Add hardware from tink-cli.
2. Try to create a workflow with random valid uuid for template Id. 

## Context
- Worker execution fails 

## Your Environment
* ubuntu
* vagrant setup

PR #286, opened 2020-09-09
Tink CLI now flags an error if template ID does not exist, while creating a workflow

gauravgahlot:
## Description

Return error if no template is found for the provided template ID.

## Why is this needed

Fixes: #282 

## How Has This Been Tested?

By attempting to create a workflow with non-existing and existing workflow ID.

Fixes a bug.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/db/mock/mock.go
+++ b/db/mock/mock.go
@@ -4,11 +4,15 @@ import (
 	"context"
 	"time"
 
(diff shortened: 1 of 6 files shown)


gianarb (approved):
Great work! :D

Issue #228, opened 2020-08-03
Move terraform deployment to a separate repository


displague:
## Expected Behaviour

The terraform deployment should be stable and published in the terraform.io registry.

## Current Behaviour

* `deploy/terraform` falls out of sync with the shared setup.sh (used by the Vagrant installer).
* The project not can be used as a module from the terraform registry

## Possible Solution

* Move terraform deployment to a new terraform-packet-tinkrbell repository.
* Publish this module

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

displague:
I have started with this in github.com/displague/terraform-packet-tinkerbell

invidian:
Isn't this done in https://github.com/tinkerbell/tink/pull/289 ?

gianarb:
Yep thanks

PR #289, opened 2020-09-10
Move terraform to sandbox

gianarb:
This is not needed here anymore. Developer can figure out their best
development environment. And this is used as a Setup guide for users. So
`sandbox` is a better place

#228


thebsdbox (approved):
`/lgtm` 

This finally removes all the installer work from the "code" repositories 🎉

Issue #295, opened 2020-09-17
Integrate a migration tool in tink


gianarb:
Tink uses Postgres as datastore.

There are two things I would like to achieve here:

1. We need a way to migrate our database schema as part of #294 . We need to add an `event` table at least.
2. Right now the database initialization is applied via docker-compose. I would like to have a more generic mechanism and even better if it can be managed by tink itself.

I explored a couple of libraries:

* https://github.com/amacneil/dbmate
    * 👍  their readme makes a comparison with the other libraries and dbmate looks like the best one ever!
    * 👎  their readme makes a comparison with the other libraries and dbmate looks like the best one ever!
    * 👎  it is in Go, so we can embed the library, but there is no documentation about how to do it
* https://github.com/golang-migrate/migrate
    * 👍  it has more starts that all the others and community is active for what I can tell
    * 👍  Packet used it internally already
    * 👍  It is designed with embeddability in mind as you can see from their README.
    * 👎  dbmate's README says that dbmate is cooler!
* https://github.com/kyleconroy/sqlc
    * 👎  It does not really do the migration, it is more like code generation from SQL.
    * 👍  I love this library but its not what I am looking for now, it works with dbmate and `migrate`. We should keep it in mind as a possible way to generate some of the code we write manually.

---

I think `migrate` is what I am looking for. As I said the migration will be applied at every boot (it is idempotent on its own, so if there are not migration to apply, we are good).

I will start with the fil source but I am investing the possibility to use a `custom` source (source is where migrations are read from) because I would like to have the migration file written in `*.go` and embedded in the binary itself, to simplify distribution.

thebsdbox:
I'm a little confused with your 👍 👎 ranking, but if we're already using `migrate` and it is relatively active then it seems like the path of least resistance 👍

rugwirobaker:
I can vouch for migrate too. The feature I like most is that I can embed my migrations in the binary using packr.

Cbkhare:
I found this big list of `Go Schema Migration tools` with good comparisons. May it can also help. 

https://povilasv.me/go-schema-migration-tools/ 

Moreover, https://github.com/golang-migrate/migrate seems to be a good choice.

gianarb:
@Cbkhare the article is good but published in `PUBLISHED FEBRUARY 20TH, 2017`. In Go it means FOREVER! eheh

@thebsdbox those are pros and cons based on my judgment

I am evaluating the lib @rugwirobaker shared because I like how it looks easy to embed

PR #296, opened 2020-09-17
Apply migrations when tink-server boots

gianarb:
## Description

It will be nice to have the tink-server capable of applying migrations
if necessary. The idea to self-apply migrations will may look unusual
but it is an easy way to distribute software capable of migrating
database schema without having to run any external procedures.
## Why is this needed

Fixes: #295 

## How Has This Been Tested?

```console
$ docker run -d -it -e  POSTGRES_PASSWORD=tinkerbell -e POSTGRES_USER=tinkerbell -p 5432:5432 postgres
# PGPASSWORD=tinkerbell PGUSER=tinkerbell PGSSLMODE=disable PACKET_VERSION=v1 PACKET_ENV=ga ROLLBAR_TOKEN=ignored go run cmd/tink-server/main.go
{"level":"info","ts":1600772721.3009,"caller":"tink-server/main.go:30","msg":"starting version devel","service":"github.com/tinkerbell/tink"}
migrations  1
records  0
{"level":"info","ts":1600772721.309925,"caller":"tink-server/main.go:54","msg":"Your database schema is not up to date. Please apply migrations running tink-server with env var ONLY_MIGRATION set.","service":"github.com/tinkerbell/tink"}
{"level":"info","ts":1600772721.310011,"caller":"metrics/metrics.go:58","msg":"initializing label values","service":"github.com/tinkerbell/tink"}
```

As suggested you should run tink-server with ONLY_MIGRATION=true

```
$ ONLY_MIGRATION=true PGPASSWORD=tinkerbell PGUSER=tinkerbell PGSSLMODE=disable PACKET_VERSION=v1 PACKET_ENV=ga ROLLBAR_TOKEN=ignored go run cmd/tink-server/main.go
{"level":"info","ts":1600772853.779405,"caller":"tink-server/main.go:30","msg":"starting version devel","service":"github.com/tinkerbell/tink"}
{"level":"info","ts":1600772853.779805,"caller":"tink-server/main.go:40","msg":"Applying migrations. This process will end when migrations will take place.","service":"github.com/tinkerbell/tink"}
{"level":"info","ts":1600772853.819854,"caller":"tink-server/main.go:45","msg":"Running migrations if necessary","service":"github.com/tinkerbell/tink","num_applied_migrations":1}
```

They are not impacted by this change.

## Checklist:

I have:

- [ ] added unit or e2e tests


gianarb:
This is just a PoC to see what other people think about it.

Ideally, I would like to add something in the tink-cli or somewhere that can help us to generate a migration `file.go` and maybe that can also update the actual slice of migrations.

The original CLI tool supports a status command, I would like to expose the same capability but via HTTP endpoint:

(original status command)
```
$ sql-migrate status
+---------------+-----------------------------------------+
|   MIGRATION   |                 APPLIED                 |
+---------------+-----------------------------------------+
| 1_initial.sql | 2014-09-13 08:19:06.788354925 +0000 UTC |
| 2_record.sql  | no                                      |
+---------------+-----------------------------------------+
```

Still open the question about rollback

db/db.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/lib/pq"
 	"github.com/packethost/pkg/log"
 	"github.com/pkg/errors"
+	migrate "github.com/rubenv/sql-migrate"
```
mmlb, line 13: I thought you said https://github.com/golang-migrate/migrate, goimports fail?

grpc-server/grpc_server.go
```diff
@@ -43,16 +43,15 @@ type server struct {
...
 	params := []grpc.ServerOption{
 		grpc.UnaryInterceptor(grpc_prometheus.UnaryServerInterceptor),
 		grpc.StreamInterceptor(grpc_prometheus.StreamServerInterceptor),
 	}
 	logger = log
 	metrics.SetupMetrics(facility, logger)
-	tinkDB := db.Connect(logger)
 	server := &server{
-		db:      tinkDB,
+		db:      db,
```
mmlb, line 54: where did `tinkDB` go/come-from? Was it a global variable? If so why can't it be deleted? Also, this change should be its own commit.

mmlb (changes requested):

db/db.go
gianarb: no, from the community I got this library as a suggestion and it is very easy to embed, kind of cool!

grpc-server/grpc_server.go
gianarb: `func SetupGRPC(ctx context.Context, log log.Logger, facility string, db *db.TinkDB, errCh chan<- error)` it is coming from `main.go`. I had to move it there in order to be able to run the migration during the tink-server's start.

db/db.go
mmlb: ah ok cool, just wondering.

grpc-server/grpc_server.go
mmlb: oh no nm. I thought it was a global var, but only because I failed to see `tinkDB := db.Connect` literally 2 lines up :p

db/migration/202009171251-init-database.go
```diff
@@ -0,0 +1,89 @@
...
+        , worker_id UUID NOT NULL
+);
+
+CREATE TABLE IF NOT EXISTS workflow_data (
+        workflow_id UUID NOT NULL
+        , version INT
+        , metadata JSONB
+        , data JSONB
+);`},
+		Down: []string{"DROP DATABASE tinkerbell"},
```
mmlb, line 87: I um worry about this. I think every thing I've ever read about Down migrations is always "no, don't". Of course I can't find anything on this atm, but I'll keep looking :D.

In general I think its just always easier to think about failing-forward instead of reverting. iirc its less foot-gunny.

db/migration/202009171251-init-database.go
```diff
@@ -0,0 +1,89 @@
+package migration
+
+import migrate "github.com/rubenv/sql-migrate"
+
+func Get202009171251() *migrate.Migration {
+	return &migrate.Migration{
+		Id: "123",
```
mmlb, line 7: I think for minimizing context it would be best to have the Id match the name?

cmd/tink-server/main.go
```diff
@@ -32,7 +33,14 @@ func main() {
 	errCh := make(chan error, 2)
 	facility := os.Getenv("FACILITY")
 
-	cert, modT := rpcServer.SetupGRPC(ctx, logger, facility, errCh)
+	tinkDB := db.Connect(logger)
+	numAppliedMigrations, err := tinkDB.Migrate()
+	if err != nil {
+		panic(err)
+	}
+	log.With("num_applied_migrations", numAppliedMigrations).Info("Running migrations if necessary")
```
mmlb, line 41: We've found out the hard way that its best to allow for migrations to run separately from normal service startup. This way once we have multiple tinks running an upgrade won't have +1 tinks trying to migrate the db.

mmlb (changes requested):

db/migration/202009171251-init-database.go
gianarb: Right now there is no logic that applies rollback :) :) That's one of the topic I  would like to discuss. We can say: "you rollback is one migration forward"

db/migration/202009171251-init-database.go
gianarb: yep

cmd/tink-server/main.go
gianarb: I didn't take this into consideration. Now the idea to migrate as part of the start is way less appealing :) I will close this PR in favor of something else :D because the title is kind of clear

db/migration/202009171251-init-database.go
mmlb: So lets drop this `Down` ;)

db/migration/202009171251-init-database.go
mmlb: It would be great to have a sed or similar check that Id matches `Get.*`.

cmd/tink-server/main.go
mmlb: Well if you start up 2 tink-servers at the same time they may both try to do some costly migration and take up db time and only one will succeed, the other will fail and restart and then continue. It would stay correct, but waste time/db resources.

Having `--migrate-only` would be better imo, as it can be a separate step in the deployment pipeline and would exit when done.

gianarb: Do you think that flag should stay in tink-server? I like that! I will check if the library supports something like a dry run that checks for migrations but does not apply them. In this way I can run it at start and notify the user that there are migration not applied yet

db/migration/202009171251-init-database.go
gianarb: Solved

cmd/tink-server/main.go
gianarb: Done @mmlb and I have updated the PR description with the new workflow

mmlb: yeah the dry run would be pretty nice imo.

cmd/tink-server/main.go
```diff
@@ -32,7 +33,28 @@ func main() {
...
 
-	cert, modT := rpcServer.SetupGRPC(ctx, logger, facility, errCh)
+	tinkDB := db.Connect(logger)
+
+	_, onlyMigration := os.LookupEnv("ONLY_MIGRATION")
+	if onlyMigration == true {
+		logger.Info("Applying migrations. This process will end when migrations will take place.")
+		numAppliedMigrations, err := tinkDB.Migrate()
+		if err != nil {
+			panic(err)
```
mmlb, line 43:
```
log.Fatal(err)
panic(err)
```
so we can get a proper error message logged + rollbar if configured, don't really need the panic but is a good signal to send to the programmer.

cmd/tink-server/main.go
```diff
@@ -32,7 +33,28 @@ func main() {
...
+		}
+		log.With("num_applied_migrations", numAppliedMigrations).Info("Migrations applied successfully")
+		os.Exit(0)
+	}
+
+	numAvailableMigrations, err := tinkDB.CheckAvailableMigrations()
+	if err != nil {
+		panic(err)
```
mmlb, line 51: ditto:
```
log.Fatal(err)
panic(err)
```

db/db.go
```diff
@@ -73,6 +76,21 @@ func Connect(lg log.Logger) *TinkDB {
 	return &TinkDB{instance: db}
 }
 
+func (t *TinkDB) Migrate() (int, error) {
+	return migrate.Exec(t.instance, "postgres", migration.GetMigrations(), migrate.Up)
+}
+
+func (t *TinkDB) CheckAvailableMigrations() (int, error) {
```
mmlb, line 83: Available seems like the wrong word to me... Necessary? Required?

db/db.go
```diff
@@ -73,6 +76,21 @@ func Connect(lg log.Logger) *TinkDB {
...
+	migrations := migration.GetMigrations().Migrations
+	records, err := migrate.GetMigrationRecords(t.instance, "postgres")
+	if err != nil {
+		return 0, err
+	}
+	fmt.Println("migrations ", len(migrations))
+	fmt.Println("records ", len(records))
```
mmlb, line 90: debugging?

db/migration/202009171251-init-database.go
```diff
@@ -0,0 +1,88 @@
...
+
+import migrate "github.com/rubenv/sql-migrate"
+
+func Get202009171251() *migrate.Migration {
+	return &migrate.Migration{
+		Id: "202009171251-init-database",
+		Up: []string{`
+SELECT 'CREATE DATABASE tinkerbell' WHERE NOT EXISTS (SELECT FROM pg_database WHERE datname = 'tinkerbell');
+
+SET ROLE tinkerbell;
```
mmlb, line 11: this should probably happen before the `SELECT 'CREATE...'` if possible.

mmlb (changes requested):

cmd/tink-server/main.go
gianarb: This
> rollbar if configured

and [this](https://github.com/tinkerbell/tink/pull/296#discussion_r492291406) do not play nice together, you can pick only one. You can have rollback or not I presume.

db/migration/202009171251-init-database.go
gianarb: Done

mmlb (approved):

Issue #294, opened 2020-09-16
Event driven implementation


gianarb:
This issue will be used as a way to summarize and centralize work required to solve proposal [[0012] Event driven system](https://github.com/tinkerbell/proposals/pull/17/)

The proposal is coming together and it is time to prototype and work on the actual implementation. The proposal contains directions and ideas about how it can be done, but the real work will be reflected here.

I want to split this work into two big pieces, ReadingPath and WritingPath.

## ReadingPath

The ReadingPath is around how events get out form tink-server. @gauravgahlot is currently having a look at this part.

### Phase 1

* How we can stream Events and serialize them via gRPC.
* How the EventClient and the `Watch` will look like
* Listen on the `event` table and stream out events

### Phase 2: 

* How we can filter events out
* Use the EventClient in the other client to build the `hardware.Watch` function.

### Prerequisite

None for now.

## WritingPath

This is all about how events are written in the database itself. It is for sure the part that has the fewest unknowns.
@parauliya  is currently looking at this.

### Phase 1

* Setup a new event table in Postgres
* Hook* only a bunch of events like:
     * CREATED, DELETED, UPDATED for the workflow resource
     * The event required for the phone home feature: 
           * WORKFLOW_STARTED
           * WORKFLOW_SUCCEEDED
           * WORKFLOW_TIMEOUT
           * WORKFLOW_FAILED
           * ACTION_STARTED
           * ACTION_SUCCEEDED
           * ACTION_TIMEDOUT
           * ACTION_FAILED

*hook means: find the right code path to write the event in Postgres and persist it!

### Phase 2

* Hook all the missing events
* Set up a listener for the event table and hook the ReadingPath with the WritingPath
* Write a garbage collector that cleans up old events, if Postgres does not have a similar feature already.

### Prerequisite

* https://github.com/tinkerbell/tink/issues/295 I noticed we do not really have a way to migrate the database. I think this is required to create the new event table for running tinkerbell. Other than being a better way to manage the database initialization (moving it from docker-compose to tink-server itself). Something like [golang-migrate/migrate](https://github.com/golang-migrate/migrate)

mmlb:
Strongly suggest `Phase1` is done with triggers like https://wiki.postgresql.org/wiki/Audit_trigger_91plus

PR #301, opened 2020-09-21
Implementation for event driven tink

gauravgahlot:
## Description

This PR provides the implementation for event-driven Tinkerbell base model proposed [here](https://github.com/tinkerbell/proposals/pull/17).

**Note**
We would like to keep `Watch` as the function used to stream events for the new event-driven tink. Therefore, the existing `hardware.Watch` has been renamed to `hardware.DeprecatedWatch`.

This will also require some changes in Hegel, so that it can use the `hardware.DeprecatedWatch`.

## Why is this needed

Fixes: #294 

## How Has This Been Tested?

Unit tests and manual testing.

This contains a bc-break. Watch has been renamed to DeprecatedWatch, clients will need to be regenerated and upgraded.

## Checklist:

I have:

- [x] added unit or e2e tests

## Known issues in progress (fixed)
- [x] an event being received twice on the same stream
- [x] `tink hardware watch <id>` fails to correctly print the hardware data for new events


mmlb (changes requested):

protos/event/event.proto
```diff
@@ -0,0 +1,47 @@
...
+}
+
+enum EventType {
+    CREATED = 0;
+    UPDATED = 1;
+    DELETED = 2;
+    WORKFLOW_INPROGRESS = 3;
+    WORKFLOW_FAILED = 4;
+    WORKFLOW_TIMEOUT = 5;
+    WORKFLOW_SUCCESS = 6;
```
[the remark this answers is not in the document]
gauravgahlot: Thanks for pointing this out. We will have the workflow state in data (payload), so it should not be an event type.

protos/event/event.proto
```diff
@@ -0,0 +1,47 @@
+syntax = "proto3";
+
+option go_package = "event";
+
+package github.com.tinkerbell.tink.protos.event;
+
+import "timestamp.proto";
+
+service EventSvc {
+    rpc Watch(WatchRequest) returns (stream Event){}
```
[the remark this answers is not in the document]
gianarb: I think Stream is a bit TOO generic :) but people says that naming is hard like caching.

You have remember that very soon you will be able to filter for a specific workflow or hardware ID, so you can watch it overtime. Stream brings to me a different idea.

gianarb:
I will post the demo here as soon as I will upload it on youtube! Thanks @gauravgahlot for it.

I have a couple of things to say about the UX and on the feature itself.

I think we should not use integer for event types and resource_type we should use strings. They are way more readable as constant and it won't require to actually have a catalog of them. A few bytes to store but who cares. If we start supporting custom events and types it will be impossible to keep a centralized index.

We have to figure out how to effectively deserialize `data` from bytes to struct and json.

The CLI I think should refect that filters we will have to implement something like `--resource-type` --resource-id` `--event-type`. The `--frame` can be a more clear `--from 5m` (that's should be the default value)

Other than that, great start!

protos/events/events.proto
```diff
@@ -0,0 +1,43 @@
...
+
+  // Every consumer when it connects to a stream of events will specify
+  // how old the events returned should be (by default 5m).
+  int32 old_events_time_span = 3; // needs a better name
+}
+
+enum ResourceType {
+  TEMPLATE = 0;
+  HARDWARE = 1;
+  WORKFLOW = 2;
```
nathangoulding, line 34: What about ACTION events as they move from pending to in progress to completed / failed?

protos/events/events.proto
```diff
@@ -0,0 +1,43 @@
...
+}
+
+enum EventType {
+  CREATED = 0;
+  UPDATED = 1;
+  DELETED = 2;
```
nathangoulding, line 40: These are basic CRUD events which are good, but I think not exhaustive enough to include events for the workflow itself (e.g. IN_PROGRESS, etc). Can we work to include those here?

nathangoulding (dismissed):

protos/events/events.proto
gauravgahlot: We initially had added events representing changes in workflow state. However, as @mmlb described in his comment [here](https://github.com/tinkerbell/tink/pull/301#discussion_r492278180) the workflow state change should be represented as an update event. The actual state of the workflow will always be present in the data.

mmlb:
@splaspood fyi

deploy/db/tinkerbell-init.sql
```diff
@@ -74,3 +74,96 @@ CREATE TABLE IF NOT EXISTS workflow_data (
...
+    END IF;
+  END IF;
+
+  RETURN NULL;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER hardware_trigger
+AFTER INSERT OR UPDATE ON hardware
+FOR EACH ROW EXECUTE PROCEDURE hardware_event();
```
mmlb, line 127: we should also ensure the old entries are trimmed away right?

protos/protoc.sh
```diff
@@ -21,7 +21,7 @@ else
 	PROTOC="docker run -v $(pwd):$BASE -w $BASE --rm $IMAGE "
 fi
 
-for proto in hardware packet template workflow; do
+for proto in hardware packet template workflow events; do
```
mmlb, line 24: nit: sort

protos/events/events.proto
gianarb: When I think about a resource I usually think about entities stores in a database. Actions are part of a workflow. They will have their own set of events but they are not a resource in the way I intend them

cmd/tink-cli/cmd/event/watch.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/google/uuid"
+	"github.com/spf13/cobra"
+	"github.com/tinkerbell/tink/client/informers"
+	"github.com/tinkerbell/tink/protos/events"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	ignoreRTAll = "ignoring \"all\" since specific resource type(s) have been provided"
+	ignoreETAll = "ignoring \"all\" since specific event type(s) have been provided"
```
mmlb, line 23: use ``` `..."all"...` ```

cmd/tink-cli/cmd/event/watch.go
```diff
@@ -0,0 +1,137 @@
...
+		"create": events.EventType_CREATED.String(),
+		"update": events.EventType_UPDATED.String(),
+		"delete": events.EventType_DELETED.String(),
+	}
+)
+
+// watchCmd represents the watch command
+var watchCmd = &cobra.Command{
+	Use:   "watch",
+	Short: "watch for event(s) on given resource(s)",
```
mmlb, line 53: `watch for events of given type(s) on given resource(s)`

db/migration/202010071530-init-events-table-and-triggers.go
```diff
@@ -0,0 +1,104 @@
...
+
+import migrate "github.com/rubenv/sql-migrate"
+
+func Get202010071530() *migrate.Migration {
+	return &migrate.Migration{
+		Id: "202010071530-init-events-table-and-triggers",
+		Up: []string{`
+SET ROLE tinkerbell;
+
+CREATE EXTENSION "uuid-ossp";
```
mmlb, line 11: create extension if not exists

db/migration/202010071530-init-events-table-and-triggers.go
```diff
@@ -0,0 +1,104 @@
...
+CREATE TYPE resource_type AS ENUM ('HARDWARE', 'TEMPLATE', 'WORKFLOW');
+CREATE TYPE event_type AS ENUM ('CREATED', 'UPDATED', 'DELETED');
```
mmlb, line 13: to make this migration idempotent we need to do:
```sql
DO $$ BEGIN
    CREATE TYPE resource_type AS ENUM ('HARDWARE', 'TEMPLATE', 'WORKFLOW');
EXCEPTION
    WHEN duplicate_object THEN null;
END $$;

DO $$ BEGIN
    CREATE TYPE event_type AS ENUM ('CREATED', 'UPDATED', 'DELETED');
EXCEPTION
    WHEN duplicate_object THEN null;
END $$;
``
for event_type

File: db/db.go [hunk 11]
```diff
@@ -95,6 +96,15 @@ func (t *TinkDB) CheckRequiredMigrations() (int, error) {
 	return len(migrations) - len(records), nil
 }
 
+// PurgeEvents periodically checks the events table and
+// purges the events that have passed the defined EVENTS_TTL.
```
mmlb, line 100: I thought we were going to do this in the db?

mmlb (changes requested):

db/db.go
gianarb: Yeah, I always forget you use the database a lot. I want to have the TTL easy to swap and configure. I imagine the TTL to be as simple as `--events-ttl 1h`, we can create a trigger with a fixed name and validate if what it does is still compliant with the TTL we set via flag @gauravgahlot ?

It looks simple to me to just start a goroutine for that, but if 🔝 can be done I think we can demand the database to do the cleaning.

I don't want to have that trigger as part of the migration because it will make it hard to configure or change

db/events.go
```diff
@@ -0,0 +1,62 @@
...
+			ID:           id,
+			ResourceID:   resourceID,
+			ResourceType: resourceType,
+			EventType:    eventType,
+			Data:         data,
+			CreatedAt:    createdAt,
+		}
+		if notification.Filter(req.ResourceId, eventTypes, resourceTypes) {
+			continue
+		}
```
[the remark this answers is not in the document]
gauravgahlot: Right, that's how it should be. Thanks, Manny.

db/db.go
gauravgahlot: IMHO I would prefer having something like `tink events prune`. I may accept either of the two flags - `--all` and `--ttl`. 
So, we could either do `tink events prune --all` or `tink events prune --ttl 60m`.

gianarb: it is not a client feature, it is a server one

client/informers/informer.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+}
+
+func (s *sharedInformer) Start(ctx context.Context, req *events.WatchRequest, fn func(e *events.Event) error) error {
+	stream, err := client.EventClient.Watch(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	go processEvents(s.eventsCh, s.errCh, fn)
```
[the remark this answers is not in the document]
gianarb: This is done in its own commit 9cadcf257b752c25da01e266afd0bf7dbf892719

db/db.go
gianarb: In the end we use a separate goroutine and I think we are holding this work for too much, this is the easiest implementation and the repo is always open for new contributions and evolution 👍

client/informers/informer.go
```diff
@@ -0,0 +1,58 @@
+package informers
```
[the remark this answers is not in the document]
gianarb: The informer is an abstraction over the `grpc client`. That's we placed it here

db/db.go
mmlb: yep that sound fine, just wanted to make sure I understood where things stood.

mmlb (changes requested):
almost there, thanks for all the continued updates!

mmlb (approved):

protos/events/events.proto
mmlb: I think lets keep simple CUD events and we can expand if we want to later, but I doubt we'll want/need to.

protos/events/events.proto
mmlb: I can see having an ACTION ResourceType, it would cut down on pushing the full workflow down the event pipe when an action has been updated, but I think we can try it this way for now.

nathangoulding:
Will events fire as actions move to in progress/complete/fail?

mmlb:
*Technically* we do have a bc-break here, but *in reality* its not. Hegel is the only client that connects to Watch that we know of, and only OSIE's osie-runner uses it. But thats only in legacy mode that doesn't even run against tink so not really a big deal. And we can probably delete the `Watch` rpc method.

gauravgahlot:
@mmlb I'm observing an issue with `tink hardware watch`, so I have removed the `ready-to-merge` label. I will debug and update here tomorrow.

mmlb:
> Will events fire as actions move to in progress/complete/fail?

iirc from the demo, yes, through the workflow. @gauravgahlot can you confirm?

gianarb:
There is not an event called ACTION_SUCCEEDED because ACTION is not a resource in our database. There is an UPDATED event for the WORKFLOW resource that gets trigger all the time the resource changes. So also when an Action changes its state.

gauravgahlot:
> > Will events fire as actions move to in progress/complete/fail?
> 
> iirc from the demo, yes, through the workflow. @gauravgahlot can you confirm?

Yes, you are right. However, to make that possible we will need to introduce another trigger on `workflow_state` stable. But that all will be taken care of in the next PR.

gauravgahlot:
@mmlb @gianarb @thebsdbox All the known issues have been fixed and the PR is ready to go ahead.

mmlb:
@gauravgahlot just to make sure, this was fixed? What was it?

> @mmlb I'm observing an issue with `tink hardware watch`, so I have removed the `ready-to-merge` label. I will debug and update here tomorrow.

gauravgahlot:
> @gauravgahlot just to make sure, this was fixed? What was it?
> 
> > @mmlb I'm observing an issue with `tink hardware watch`, so I have removed the `ready-to-merge` label. I will debug and update here tomorrow.

Yes, this has been fixed. The issue was that as a listener starts the first event was received twice. I observed it with tink hardware watch first, however, the same issue was with the tink events watch as well. 

It has been fixed now.

gianarb (dismissed):
This is g2g for my side! Thanks, @gauravgahlot for the hard work!

mmlb (approved):

gauravgahlot:
Updated branch with a rebase, no changes.

gianarb (approved):

mmlb (approved):

PR #314, opened 2020-10-01
Disable codecov path

gianarb:
## Description

I am not sure we are ready to get this granularity, all our PRs are
failing for that reason and I don't feel strong enough to push other
people or myself to make the patch itself green.

There is still a check that looks at diff in coverage for all project, but not for the commit itself

## Why is this needed

Coverage is important but sanity is more.

none

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,3 @@
+coverage:
+  status:
+    patch: off


thebsdbox:
Hmm, interesting.. is the plan to re-enable this at a later point?

gianarb:
Well enabling it is one pull request far away. Codecov has a nice public dashboard can I presume we can use to understand if and how to enable it. We are using the default behavior and I think it is acceptable to tune it based on needs.

mmlb:
I think its actually just failing because we don't have the minimum coverage, https://docs.codecov.io/docs/commit-status#target. Can we try setting it to auto? I believe we've had this same setup before in other internal repos.

gianarb:
I think you linked something related to the project status @mmlb . The one that very often fails in a way that is not activable (at least imho) is the patch one.

What I am not understanding?
Thanks

mmlb:
Hmm I suspect that the project status is affecting the patch status. But there is also a `target: auto` option for patch that we may want to go with. I feel like disabling patch is a big hammer that we may not actually need though. codecov has 0 config as is, I think we should make sure one of the options isn't going to allow PRs that don't make things work pass.

gianarb:
I don't know, the alternative is to keep thing as it is and getting all the PR red because we didn't test a single line of code (in a project that as ~20% coverage it looks like a big push) 😅

Issue #300, opened 2020-09-18
Delete pkg package


mmlb:
pkg packages is an anti pattern. Instead we should break out its contents into smaller/easier to use packages.

detiber:
+1000, we should also move anything that we don't intend to expose as a public API into one or more 'internal' packages

vigneshashokan:
@mmlb If no one is working on this, I can take a stab at this issue.

gauravgahlot:
@vigneshashokan I don't think anyone is working on it yet. I will assign the issue to you. 

Please take read at the [communication](https://github.com/tinkerbell/.github/blob/master/COMMUNICATION.md) doc and do let us know if you face any issues in the slack channel.

Looking forward to your contribution. 😄

mmlb:
@vigneshashokan I've done one part in #315 already.

chrisdoherty4:
Stale issue. Some of this work is being addressed with an overhaul of the Tink repository anyhow.

PR #315, opened 2020-10-01
Move workflow stuff out of pkg

mmlb:
## Description

Move workflow types and funcs out of pkg/ and into its own package. Also cleaned up some of the API and added some more validation while I'm here.

## Why is this needed

Helps address #300 

## How Has This Been Tested?

`go test` passes.

Low impact. A bc-break is occurring here, but its in internalish code and no user is expected to be using this code. If they are `gofmt -w -r 'p.ParseYAML(b) -> p.Parse(b)' .` and removing the call to `ValidateTemplate` is all thats needed.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/db/template.go
+++ b/db/template.go
@@ -23,10 +23,6 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 	if err != nil {
 		return err
 	}
-	err = wflow.ValidateTemplate(wf)
-	if err != nil {
-		return err
-	}
 
 	_, err = tx.Exec(`
 	INSERT INTO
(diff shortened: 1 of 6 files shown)


gianarb:
Is it fixing this as well https://github.com/tinkerbell/tink/issues/317 ?

mmlb:
@gianarb no? I dont see where #317 is valid in non-test code. Every error is returned from `TinkDB.CreateTemplate` which is checked and returned to callers.

detiber (dismissed):
These changes look good to me, down with `pkg` and thank you for internalizing the template validation.

mmlb:
> These changes look good to me, down with `pkg` and thank you for internalizing the template validation.

Thanks for the review.

mmlb:
I'll leave this up for @gianarb til tomorrow.

gianarb (approved):

gianarb (approved):

PR #322, opened 2020-10-07
[tink-worker] Fix use and defaulting of the max-retry flag

detiber:
## Description

Previously we were trying to fetch the wrong flag value when configuring retries for tink-worker, this PR fixes that.

## How Has This Been Tested?
I tested this in my Tilt/kind setup

Fixes issues with using the latest tink-worker for e2e testing/deployment.

It doesn't require any changes for users or require any additional migration.


displague (approved):
This coincides with https://github.com/tinkerbell/tink/blob/1009dc3658d5d514875ab51252f857e4c7c79a37/cmd/tink-worker/cmd/root.go#L85

Issue #312, opened 2020-10-01
Multi arch container support


gianarb:
Hello

Currently, we build our binaries outside a container using `go build` directly. I would like to build them inside a container for a couple of reasons:

1. We serve a docker-compose file to work locally and it uses the [build](https://docs.docker.com/compose/compose-file/#build) directive to refresh a container starting from Dockerfile. As it is today it refresh the container using the old binary. There are workaround but if we build the binaries as part of the Dockerfile the refresh will pick up code changes as well
2. If we build binaries inside a container we can use buildx to support multiple architectures. We can do it as well outside but it sounds less trivial

Multi Arch means tinkerbell on ARM!

thebsdbox:
I'm 💯 x 💯  in support of this, a simple `Dockerfile` like below would work:

```
 # syntax=docker/dockerfile:experimental

FROM golang:1.15-alpine as dev
RUN apk add --no-cache git ca-certificates make
RUN adduser -D appuser
COPY . /src/
WORKDIR /src

ENV GO111MODULE=on
RUN --mount=type=cache,sharing=locked,id=gomod,target=/go/pkg/mod/cache \
    --mount=type=cache,sharing=locked,id=goroot,target=/root/.cache/go-build \
    CGO_ENABLED=0 GOOS=linux make build

FROM scratch
COPY --from=dev /src/tink-server /
ENTRYPOINT ["/tink-server"]
```

gauravgahlot:
A definite **YES** for it. 💯

rugwirobaker:
@gianarb  can I take this on?

gianarb:
Sure @rugwirobaker !! Thank you for your help!

rugwirobaker:
@thebsdbox I haven't used this method of mounting the module cache as a volume before.

> RUN --mount=type=cache,sharing=locked,id=gomod,target=/go/pkg/mod/cache \
>     --mount=type=cache,sharing=locked,id=goroot,target=/root/.cache/go-build \
>     CGO_ENABLED=0 GOOS=linux make build

Is there an advantage over downloading them in their own stage like:
```
# syntax=docker/dockerfile:experimental

FROM golang:1.15-alpine as dev
WORKDIR $GOPATH/src/github.com/tinkerbell/tink
COPY go.mod go.sum  ./
RUN GO111MODULE=on GOPROXY="https://proxy.golang.org" go mod download
COPY . .
CGO_ENABLED=0 GOOS=linux make build


FROM scratch
.....
[...]

PR #323, opened 2020-10-08
Build binaries as part of Docker

gianarb:
## Description

Moving the binary compilation as part of Docker we will be able to
leverage the multi-architecture support Docker has to provide images
that can run on different architecture such as x64_86, ARM.
## Why is this needed

Reference: #312


thebsdbox:
image

cmd/tink-server/Dockerfile
```diff
@@ -1,10 +1,16 @@
...
 
-ENTRYPOINT ["tink-server"]
 EXPOSE 42113
 EXPOSE 42114
 
 RUN apk add --update ca-certificates && \
     apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing cfssl
 
-COPY tink-server /bin/
+ENTRYPOINT ["/usr/bin/tink-server"]
```
mmlb, line 16: I think lets keep the image "metadata" together at the top and then the contents after so it would be like:

```
FROM alpine:3.11

EXPOSE 42113
EXPOSE 42114
ENTRYPOINT ["/usr/bin/tink-server"]

RUN apk add ...
COPY --from=0 ...
```

Notice I re-arranged the COPY, RUN -> RUN, COPY so we can re-use the RUN apk cached entry.

cmd/tink-worker/Dockerfile
```diff
@@ -1,7 +1,10 @@
...
+COPY . /usr/myapp
+WORKDIR /usr/myapp
+RUN make worker
 
+FROM alpine:3.11
+COPY --from=0 /usr/myapp/cmd/tink-worker/tink-worker /usr/bin/tink-worker
 RUN apk add --no-cache --update --upgrade ca-certificates
-
-COPY tink-worker /tink-worker
+ENTRYPOINT [ "/usr/bin/tink-worker" ]
```
mmlb, line 10: ditto, lets do ENTRYPOINT, RUN, COPY to make best use of cache.

cmd/tink-cli/Dockerfile
```diff
@@ -1,8 +1,10 @@
...
+WORKDIR /usr/myapp
+RUN make cli
 
+FROM alpine:3.11
 RUN apk add --no-cache --update --upgrade ca-certificates
-
-COPY tink-cli /bin/tink
-COPY sample.tmpl /tmp
+COPY --from=0 /usr/myapp/cmd/tink-cli/tink-cli /usr/bin/tink-cli
+ENTRYPOINT ["/usr/bin/tink-cli"]
```
mmlb, line 10: ditto, lets do ENTRYPOINT, RUN, COPY to make best use of cache and keep groupings of metadata vs content

mmlb (changes requested):

mmlb (approved):

mmlb (dismissed):

mmlb (dismissed):

mmlb (approved):

Issue #319, opened 2020-10-02
Consistently name client interfaces


invidian:
## Expected Behaviour

One can easily switch between working on code around workflows and templates and naming should follow the same convention, so interface names are not surprising, following [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

## Current Behaviour

- `github.com/tinkerbell/tink/protos/workflow` has `WorkflowSvcClient`
- `github.com/tinkerbell/tink/protos/hardware` has `HardwareServiceClient`
- `github.com/tinkerbell/tink/protos/template` has `TemplateClient`

mmlb:
100% yes!

mmlb:
Thanks @invidian take a look at #345 which addresses this along with other protobuf related warts I encountered.

PR #345, opened 2020-10-19
Show protos some love

mmlb:
## Description

This overhauls and fixes a bunch of warnings/nits in our protobufs. Also, I wasn't able to build the protobufs because of version mismatches and what not.

## Why is this needed

I was not happy with the state of our protobuf build setup, managing protobuf dependencies and general protobuf lint/style issues.

Also,
Fixes: #319 

## How Has This Been Tested?

Builds and `go test ./...` works.

Custom clients using our protobuf files will need to be updated, I don't know of any such clients.


jacobweinstock (dismissed):
you have a `buf.yaml` file in here, but I didn't see `buf` being run anywhere? did i miss it?

tools.go
```diff
@@ -0,0 +1,12 @@
+// +build tools
```
displague, line 1: What would the relationship be between this file and go.mod? Is this adding versioned dependencies that would be pruned by `go mod tidy` when not making a `tools` build?

Could `tools.go` live in the `protoc/` directory with an isolated go.mod in that directory? (`go list -m all` could be used to list the dependencies if that were the case.)

We run into the problem that users will have the dependencies already installed, but at the wrong versions of those dependencies, resulting in proto file jitter depending on the committer.

The biggest potential benefit of this change is that the protobuf tooling version used to create the generated files would match the ones that are used in other parts of this project.  If that is the goal, then the containerized protoc.sh approach should be removed.

The proto builds must be repeatable.  I would like to see a protoc build added to the Github actions to look for proto file diffs that were not incorporated in a commit.

If controlling the local version of the proto tools is problematic, I would recommend that we use our own Dockerfile to control the tooling used in the proto file builds.

mmlb: > What would the relationship be between this file and go.mod? Is this adding versioned dependencies that would be pruned by `go mod tidy` when not making a `tools` build?

This file is used in harmony with go.mod. This is coming from one of the options in the Go Wiki, golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module.

> Could `tools.go` live in the `protoc/` directory with an isolated go.mod in that directory? (`go list -m all` could be used to list the dependencies if that were the case.)

I'd prefer not to do that. For one it seems weird to define a mod in a sub directory since mods are meant to be at a repo root right? Having these deps in the top-level mod also ensures they are always being used.
 
> We run into the problem that users will have the dependencies already installed, but at the wrong versions of those dependencies, resulting in proto file jitter depending on the committer.

This takes care of that since we'd be using the pinned versions of the tools from the go.mod file.
 
> The biggest potential benefit of this change is that the protobuf tooling version used to create the generated files would match the ones that are used in other parts of this project. If that is the goal, then the containerized protoc.sh approach should be removed.

I meant to do this, but forgot. This PR still doesn't deal with `protoc` versions/installations. I deal with that with `shell.nix` personally, the build-in-docker approach in protoc.sh also helps there too. 

> 
> The proto builds must be repeatable. I would like to see a protoc build added to the Github actions to look for proto file diffs that were not incorporated in a commit.

Can do.

> If controlling the local version of the proto tools is problematic, I would recommend that we use our own Dockerfile to control the tooling used in the proto file builds.

I don't particularly like leaning so much on docker for all dev tooling, though I think the happy medium is a nix docker image that reads the shell.nix file and uses those versions :D.

mmlb:
> you have a `buf.yaml` file in here, but I didn't see `buf` being run anywhere? did i miss it?

Not in an automated fasion. I used it locally to run some lint checks.

mmlb:
> > The proto builds must be repeatable. I would like to see a protoc build added to the Github actions to look for proto file diffs that were not incorporated in a commit.
>
> Can do.

actually @displague I'd like to do that in a follow up PR. This one doesn't make things any worse wrt CI and is already pretty big.

go.mod
```diff
@@ -27,16 +27,17 @@ require (
...
+	google.golang.org/genproto v0.0.0-20200921165018-b9da36f5f452
 	google.golang.org/grpc v1.29.1
-	google.golang.org/protobuf v1.25.0 // indirect
-	gopkg.in/yaml.v2 v2.2.5
+	google.golang.org/protobuf v1.25.0
+	gopkg.in/yaml.v2 v2.3.0
 	gotest.tools v2.2.0+incompatible // indirect
 )
+
+replace github.com/stormcat24/protodep => github.com/ackintosh/protodep v0.0.0-20200728152107-abf8eb579d6c
```
mmlb, line 43: replaced with https://github.com/stormcat24/protodep/pull/49, otherwise I was only able to make it work with a bogus `-i` ssh identity file

gianarb:
@mmlb if you are experimenting with `buf` locally can you remove it from the vcs? as you said, you can do it all together in its own PR.

gianarb:
Overall I like what I see and I think we have to merge this PR pretty soon, it is big and will require a lot of work in order to keep it up to date with the codebase and I would like to avoid that.

@mmlb I think you should write a note in `CONTRIBUTORS.md` about how we can generate proto and how the tools package works. Just to have some sort of documentation.

Thanks

mmlb:
> Overall I like what I see and I think we have to merge this PR pretty soon, it is big and will require a lot of work in order to keep it up to date with the codebase and I would like to avoid that.

Yep, I have one more thing to check and make sure isn't broken as it failed for me in some tests yesterday.

> 
> @mmlb I think you should write a note in `CONTRIBUTORS.md` about how we can generate proto and how the tools package works. Just to have some sort of documentation.

I'll do that in a follow up PR to avoid growing this one more.

mmlb:
@displague @jacobweinstock @gianarb PTAL, I've rebased on top of master to pull the `goimports` commit out of here.

gianarb:
> I'll do that in a follow up PR to avoid growing this one more.

Follow up PR for documentation scares me so much xD Often they never arrive!  xD

mmlb:
:D especially with me :D. Ok I'll whip something up then.

mmlb:
hmm we don't even have a CONTRIBUTING.md doc, grabbing from .github.

mmlb:
Alright @gianarb I threw in CONTRIBUTING.md over from boots and added a bit about protobuf files: https://github.com/mmlb/tink/blob/show-protos-some-love/CONTRIBUTING.md

@displague I dropped the use of jaegertracing/protobuf too

gianarb (approved):

PR #342, opened 2020-10-15
ENG-9317 Fix concurrent workflow create bug

kdeng3849:
Signed-off-by: Kelly Deng <kelly@packet.com>

## Description

Lowers the transaction isolation level from `serializable` to `repeatable read` as `serializable` appears to be too overly strict.

## Why is this needed

When creating several workflows in parallel, an error is returned:
```
rpc error: code = Unknown desc = Failed to workflow: inserting e3ea67ac-2d5d-45ce-8cb3-fca42abc896f, INSERT in to workflow: pq: could not serialize access due to read/write dependencies among transactions
```

### example:
```
/ # cat inputs | parallel
Created Workflow:  7c8404ee-52a7-49a9-bacc-57ae9171e8c1
2020/10/14 15:08:18 rpc error: code = Unknown desc = Failed to workflow: inserting e3ea67ac-2d5d-45ce-8cb3-fca42abc896f, INSERT in to workflow: pq: could not serialize access due to read/write dependencies among transactions
Created Workflow:  6981c47b-443b-4d69-932b-4e6b6d1ce678
2020/10/14 15:08:18 rpc error: code = Unknown desc = Failed to workflow: inserting 48de21ee-5bac-4498-96c4-21c4ad46ae33, INSERT in to workflow: pq: could not serialize access due to read/write dependencies among transactions
Created Workflow:  a0921f77-e7f6-4aeb-8c5b-823e1d57168c
Created Workflow:  875a9033-51ce-4b5a-a786-1d3c0520eeb4
2020/10/14 15:08:18 rpc error: code = Unknown desc = Failed to insert in workflow_state: inserting 5f1a1f7c-d264-4339-b0fc-12f5fb72bd81, INSERT in to workflow_state: pq: could not serialize access due to read/write dependencies among transactions
Created Workflow:  cde501b0-51f0-4df8-9844-7c0e581a2e69
Created Workflow:  4acef5ac-5299-45ec-bbe9-56871a6fbe61
Created Workflow:  83a6dc36-bfae-456d-8b46-501e6dd16f03
```

**`inputs` file contents**
```
tink workflow create -t 67a0b806-6315-4a98-a56d-99fb03b30573 -r '{"device_1": "b4:96:91:5f:af:c0"}'
tink workflow create -t 67a0b806-6315-4a98-a56d-99fb03b30573 -r '{"device_1": "b4:96:91:5f:af:c0"}'
[...]

PR #346, opened 2020-10-21
Fix concurrent workflow create bug by unique index

ryli17:
## Description

When multiple process concurrently create workflow, the code to access workflow_worker_map is not concurrent safe. This PR fix this issue. 

## Why is this needed

During workflow creation, new rows may be add to table `workflow_worker_map` in the same transaction. The original design use ` INSERT ... SELECT`. This doesn't work concurrently with SERRIELIZABLE isolation.

Fixes: #

- Create unique index on workflow_worker_map table
- Use `INSERT ... ON CONFLICT DO NOTHING;` to avoid table lock.

## How Has This Been Tested?

Manually as https://github.com/tinkerbell/tink/pull/342.

- No impact
- database migrate to 202010221010-add-unique-index.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/db/workflow.go
+++ b/db/workflow.go
@@ -73,15 +73,13 @@ func insertInWorkflow(ctx context.Context, db *sql.DB, wf Workflow, tx *sql.Tx) 
 }
 
 func insertIntoWfWorkerTable(ctx context.Context, db *sql.DB, wfID uuid.UUID, workerID uuid.UUID, tx *sql.Tx) error {
-	// TODO This command is not 100% reliable for concurrent write operations
 	_, err := tx.Exec(`
 	INSERT INTO
 		workflow_worker_map (workflow_id, worker_id)
-	SELECT $1, $2
-	WHERE
-		NOT EXISTS (
-			SELECT workflow_id FROM workflow_worker_map WHERE workflow_id = $1 AND worker_id = $2
-		);
+	VALUES
+	        ($1, $2)
+	ON CONFLICT (workflow_id, worker_id)
+	DO NOTHING;
 	`, wfID, workerID)
 	if err != nil {
 		return errors.Wrap(err, "INSERT in to workflow_worker_map")
--- a/db/migration/202010221010-add-unique-index.go
+++ b/db/migration/202010221010-add-unique-index.go
@@ -1,0 +1,12 @@
+package migration
+
+import migrate "github.com/rubenv/sql-migrate"
+
+func Get202010221010() *migrate.Migration {
+        return &migrate.Migration{
+                Id: "202010221010-add-unique-index",
+                Up: []string{`
+CREATE UNIQUE INDEX IF NOT EXISTS uidx_workflow_worker_map ON workflow_worker_map (workflow_id, worker_id);
+`},
+        }
+}
--- a/db/migration/migration.go
+++ b/db/migration/migration.go
@@ -6,6 +6,7 @@ func GetMigrations() *migrate.MemoryMigrationSource {
 	return &migrate.MemoryMigrationSource{
 		Migrations: []*migrate.Migration{
 			Get202009171251(),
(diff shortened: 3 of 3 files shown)


kdeng3849 (changes requested):
Can you squash the commits?

db/migration/202009171251-init-database.go
```diff
@@ -77,6 +77,9 @@ CREATE TABLE IF NOT EXISTS workflow_worker_map (
         , worker_id UUID NOT NULL
 );
 
+ALTER TABLE workflow_worker_map DROP CONSTRAINT IF EXISTS workflow_worker_pkey;
```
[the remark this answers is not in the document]
mmlb: This needs to be done in a different migration file/procedure so that it will be run automatically.

db/migration/202009171251-init-database.go
```diff
@@ -77,6 +77,9 @@ CREATE TABLE IF NOT EXISTS workflow_worker_map (
...
+ALTER TABLE workflow_worker_map ADD CONSTRAINT workflow_worker_pkey PRIMARY KEY (workflow_id, worker_id);
```
mmlb, line 81: This needs to be done in a different migration file/procedure so that it will be run automatically.

Would it be better to do a concurrent unique index and then attach as a table constraint? Migrations will most likely be run in parallel with existing running code and locking the table doesn't sound like a good idea.

db/workflow.go
```diff
@@ -73,15 +73,13 @@ func insertInWorkflow(ctx context.Context, db *sql.DB, wf Workflow, tx *sql.Tx)
...
 		workflow_worker_map (workflow_id, worker_id)
-	SELECT $1, $2
-	WHERE
-		NOT EXISTS (
-			SELECT workflow_id FROM workflow_worker_map WHERE workflow_id = $1 AND worker_id = $2
-		);
+	VALUES
+	        ($1, $2)
+	ON CONFLICT (workflow_id, worker_id)
+	DO NOTHING;
```
mmlb, line 82: this will still return an error right?

mmlb (changes requested):

db/workflow.go
ryli17: No, this will not return error on conflict.

db/migration/202009171251-init-database.go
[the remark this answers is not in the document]
ryli17: Previous consideration is psql do not support `ALTER TABLE table ADD CONSTRAINT constraint PRIMARY KEY IF NOT EXISTS`. New implementation just uses unique index.

db/migration/202009171251-init-database.go
ryli17: Done. Actually a unique index is sufficient in this case. I moved it to a separate migration 202010221010-add-unique-index.

ryli17:
squashed the commits.

gianarb (dismissed):

db/workflow.go
mmlb: This is not what we want right? We should be returning an error so that we can bubble it up to the caller.

mmlb (changes requested):

db/workflow.go
ryli17: @mmlb, This actually is what we want. If the row is already in the table, nothing need to do. This behavior is the same as the the original `SELECT ...` code try to achieve. The fix here is to remove `SELECT ...` block, which voilate serializable isolation policy.

mmlb: ah true that makes sense to keep the same behavior. Not sure the original behavior was the best idea, but that should be handled separately if necessary.

mmlb (approved):

kdeng3849 (approved):

PR #349, opened 2020-10-26
Support fetching templates by name

kdeng3849:
## Description

Adds the ability to query templates by name.   
The `ListTemplates` method has been changed to accept a `FilterRequest` instead of just `Empty`.   
The `FilterRequest` has a field `filter` that accepts a string used to match the names of templates. A wildcard `*` can be used to match a sequence of zero or more characters.  
Passing an empty `FilterRequest` will return all templates.

## Why is this needed

A user might have the following templates: `hello-world`, `ubuntu-01`, `ubuntu-02`, `ubuntu-03`, `goodbye-world` and wants to list only the linux templates. Before, they would only be able to either list all of the templates, or only one (and only if they already know the id).  
With this PR, they can just call `ListTemplates` with the `FilterRequest`'s filter field to `ubuntu-*` and it will return only `ubuntu-01`, `ubuntu-02`, and `ubuntu-03`.

Fixes: https://github.com/tinkerbell/tink/issues/352 / ENG-9455

## How Has This Been Tested?

* manually
* TBD

Users using an older version of the tink-cli may experience errors. Best solution is to update the tink-server, tink-cli, and tink-worker to use the images built from this PR

## Checklist:

I have:

- [ ] added unit or e2e tests


mmlb:
I opened up #351 to fix the whitespace issue and avoid it in the future by making sure CI catches.

gianarb:
Do we want somehow to formalize and make this filtering available for other resources? I am just trying to avoid an unfriendly proliferation of different filters that probably sounds useful short term but won't make usability nice in long term

kdeng3849:
@gianarb Can you elaborate? Are you suggesting we should also provide the option to filter by any attribute of a resource?

gianarb:
@kdeng3849 yeah something like that.

I am wondering if we should build this feature in a way that will allow us to get filtering done for all the resources across fields in a consistent way. Without building something ad hoc for one case. 

I think having the filtering done one by one will make the UX inconsistent and not great

gianarb (approved):

gianarb (changes requested):
We need a rebase, @mmlb 's fault :P

mmlb:
oh that merged already? :toot:

grpc-server/template_test.go
```diff
@@ -107,3 +110,76 @@ func TestCreateTemplate(t *testing.T) {
...
+
+func TestGetTemplateByName(t *testing.T) {
+	type (
+		args struct {
+			db   mock.DB
+			name string
+		}
+		want struct {
+			expectedError bool
+		}
```
mmlb, line 122: can we make this something like:

```suggestion
		err bool
```
instead? No need to define a struct for just one type and its easier to use too.

mmlb (changes requested):

mmlb:
> @kdeng3849 yeah something like that.
> 
> I am wondering if we should build this feature in a way that will allow us to get filtering done for all the resources across fields in a consistent way. Without building something ad hoc for one case.
> 
> I think having the filtering done one by one will make the UX inconsistent and not great

Yeah I think maybe something like
```proto
GetTemplate(GetRequest) returns (WorkflowTemplate)
...
message GetRequest {
  oneof get_oneof {
    string id = 1;
    string name = 2;
  }
}
```
we can leave the other oneof options for later if necessary. I think I'd rather see an pattern we can agree on that is extensible to avoid GetBy... for each proto type. Does this line up with what you're thinking @gianarb ?

mattcburns (dismissed):

gianarb:
Yeah I think it sounds good @mmlb

DavidHwu:
@mmlb & @gianarb this PR is blocking Central Eng (or @mattcburns ) from moving forward.
I've read your change requests... is this a must have or NTH (Nice To Have)?

gianarb (approved):
I am having some difficulties visualizing the grpc API right now but I don't want to hold more on this :)

db/db.go
```diff
@@ -35,7 +36,7 @@ type hardware interface {
 
 type template interface {
 	CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
-	GetTemplate(ctx context.Context, id string) (string, string, error)
+	GetTemplate(ctx context.Context, attributes map[string]string) (string, string, error)
```
mmlb: I think a struct will be better to use here.

protos/template/template.proto
```diff
@@ -51,4 +51,9 @@ message CreateResponse {
 
 message GetRequest {
   string id = 1;
+  string name = 2;
```
mmlb: this is in the wrong commit :D

cmd/tink-cli/cmd/template/list.go
```diff
@@ -43,7 +43,7 @@ var listCmd = &cobra.Command{
 }
 
 func listTemplates(cmd *cobra.Command, t table.Writer) {
-	list, err := client.TemplateClient.ListTemplates(context.Background(), &template.FilterRequest{Filter: "%"})
+	list, err := client.TemplateClient.ListTemplates(context.Background(), &template.FilterRequest{Filter: "*"})
```
mmlb, line 46: this commit should be squashed into the previous one

grpc-server/template_test.go
```diff
@@ -2,6 +2,7 @@ package grpcserver
```
mmlb, line 2: Please update the commit message to something like

> Add tests for Get filtering

grpc-server/workflow.go
```diff
@@ -270,15 +270,18 @@ func (s *server) ShowWorkflowEvents(req *workflow.GetRequest, stream workflow.Wo
 	return nil
```
mmlb, line 270: I think adding createYaml to the commit message would help when looking through git history:

> createYaml: use more descriptive argument names

Rename temp and tempData to templateID and templateData for better clarity.

db/db.go
```diff
@@ -36,7 +36,7 @@ type hardware interface {
 
 type template interface {
```
mmlb, line 37: agree with this, but should just be squashed in the commit that does it :D

protos/template/template.proto
```diff
@@ -51,4 +51,9 @@ message CreateResponse {
...
+}
+
+message FilterRequest {
+  string filter = 1;
 }
```
mmlb, line 61: I think it'd be great to have `FilterRequest` look like `GetRequest` so that we have one way to do filtering. This would also help to convey what field the filter is taking place on.

```proto
message FilterRequest {
  one_of filter_types {
    string name = 1;
  }
}
```

mmlb (changes requested):

mmlb (approved):

PR #355, opened 2020-10-29
Fix login to quay only in master

gianarb:
The login to Quay has to happen only for commits landing to master


thebsdbox:
`/lgtm` - my only question is, are we renaming master to something else?

thebsdbox (approved):

gianarb:
I have no idea @thebsdbox

PR #358, opened 2020-11-06
Remove hard coded role and db name from init db migration

splaspood:
## Description

The initial database migration currently has a hard coded database name and role.  This change removes both.

## Why is this needed

This prevents us from deploying tink against a database that happens to utilize a different role or db name.
## Checklist:

I have:

- [ ] added unit or e2e tests


mmlb (dismissed):

mmlb:
@splaspood you need to update the docker-compose.yml so it creates the db

mmlb (changes requested):
@splaspood you need to update the docker-compose.yml so it creates the db

mmlb (approved):

PR #362, opened 2020-11-10
fix getting template by name not returning the associated id

kdeng3849:
## Description

Fetching a template by name now also returns the associated ID.

## Why is this needed

A recent PR https://github.com/tinkerbell/tink/pull/349 added the functionality of fetching a template by name. The returned template is expected to include the ID, name, and template data. However, currently, it would only return the name and the template data. The returned ID is just an empty string.

## How Has This Been Tested?

Manually

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/grpc-server/workflow_test.go
+++ b/grpc-server/workflow_test.go
@@ -41,11 +41,11 @@ func TestCreateWorkflow(t *testing.T) {
 		args args
 		want want
 	}{
-		"FailedToGetTempalte": {
+		"FailedToGetTemplate": {
 			args: args{
 				db: mock.DB{
-					GetTemplateFunc: func(ctx context.Context, fields map[string]string) (string, string, error) {
-						return "", "", errors.New("failed to get template")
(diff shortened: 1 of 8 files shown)


mmlb (approved):

grpc-server/workflow_test.go
```diff
@@ -58,8 +58,8 @@ func TestCreateWorkflow(t *testing.T) {
 		"FailedCreatingWorkflow": {
 			args: args{
 				db: mock.DB{
-					GetTemplateFunc: func(ctx context.Context, fields map[string]string) (string, string, error) {
-						return "", templateData, nil
+					GetTemplateFunc: func(ctx context.Context, fields map[string]string) (string, string, string, error) {
```
shadowcolossus192, line 61: Instead of returning an empty string, maybe we can return a pre-defined name for mock (e.g., `MockedTemplate`?).

PR #364, opened 2020-11-12
Run in Kubernetes

mrchrd:
## Description

This adds an option to deploy Tinkerbell using Kubernetes (k3s) instead of Docker Compose.

## Why is this needed

This allows Tinkerbell to run in Kubernetes. This will enable further development like Helm Charts, Operators, etc.

## How Has This Been Tested?

Manually tested the Hello World workflow.

There's no impact. Users can use Kubernetes by setting up environment variable `USE_KUBERNETES=true` in `envrc`, or before launching `vagrant up` or `generate-envrc.sh`. Leaving the variable undefined or setting it to `false` will use Docker Compose as expected.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/deploy/kubernetes/boots.yaml
+++ b/deploy/kubernetes/boots.yaml
@@ -1,0 +1,111 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app: boots
+  name: boots
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: boots
+  strategy:
+    type: Recreate
+  template:
+    metadata:
+      labels:
+        app: boots
+    spec:
+      containers:
+      - args:
+        - -dhcp-addr
+        - 0.0.0.0:67
+        - -tftp-addr
+        - 0.0.0.0:69
+        - -http-addr
+        - 0.0.0.0:80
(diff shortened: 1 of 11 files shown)


detiber:
@mrchrd After speaking with @gianarb we thought it would be best if we moved the kubernetes work to a separate repository, similar to https://github.com/tinkerbell/sandbox.

I went ahead and created https://github.com/tinkerbell/k8s-sandbox. Do you mind porting the work in this PR over to that repository?

mrchrd:
Then, in order to avoid duplicating code, I suggest that:
- Deployment scripts and assets are removed from the tink repository; and
- Either:
  - This PR is moved to the existing sandbox repository; or
  - We move the content of the kindDev branch to the k8s-sandbox directory (and forget this PR).

detiber:
@mrchrd My apologies for not elaborating more, my suggestion was to start the k8s-sandbox with the work you have from this PR, and then I would port over my kindDev branch work to that repo, rebasing on top of the work you've done.

The goal is indeed to remove deployment scripts and assets from this repository completely.

mrchrd:
https://github.com/tinkerbell/k8s-sandbox/pull/3

PR #365, opened 2020-11-13
Fix hardware not found response

kdeng3849:
## Description

When getting hardware and hardware is not found, the error `unexpected end of JSON input` is no longer returned.
The error returned will be a `hardware not found: sql: no rows in result set` instead.

In addition, the error message received when trying to create a workflow with a worker/hardware not present in the database is now:
```
2020/11/18 21:24:01 rpc error: code = Unknown desc = failed to create workflow: unable to insert into action list: no worker found: mac: b4:96:91:5f:ac:00
```

## Why is this needed

The error message was vague and misleading. 

Error message before (as seen in boots):
```
boots_1        | {"level":"info","ts":1605204343.6437285,"caller":"src/dhcp.go:76","msg":"retrieved job is empty","service":"github.com/tinkerbell/boots","pkg":"main","type":"DHCPDISCOVER","mac":"f8:f2:1e:a6:cd:10","err":"discover from dhcp message: get hardware by mac from tink: rpc error: code = Unknown desc = unexpected end of JSON input","errVerbose":"rpc error: code = Unknown desc = unexpected end of JSON input\nget hardware by mac from tink\ngithub.com/tinkerbell/boots/packet.
...
```

Error message after:
```
boots_1        | {​​​​​​​​"level":"info","ts":1605289315.5323317,"caller":"myapp/dhcp.go:76","msg":"retrieved job is empty","service":"github.com/tinkerbell/boots","pkg":"main","type":"DHCPDISCOVER","mac":"b4:96:91:5f:ac:e0","err":"discover from dhcp message: get hardware by mac from tink: rpc error: code = Unknown desc = sql: no rows in result set","errVerbose":"rpc error: code = Unknown desc = sql: no rows in result set\nget hardware by mac from tink\ngithub.com/tinkerbell/boots/packet..
...
```
(I know, sorry for the bad formatting)

## How Has This Been Tested?

Manually

N/A

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -102,18 +102,11 @@ func get(ctx context.Context, db *sql.DB, query string, args ...interface{}) (st
 
 	buf := []byte{}
 	err := row.Scan(&buf)
-	if err == nil {
-		return string(buf), nil
-	}
-
-	if err != sql.ErrNoRows {
-		err = errors.Wrap(err, "SELECT")
+	if err != nil {
 		logger.Error(err)
-	} else {
-		err = nil
+		return "", errors.Wrap(err, "SELECT")
 	}
-
-	return "", err
+	return string(buf), nil
 }
 
 // buildGetCondition builds a where condition string in the format "column_name = 'field_value' AND"
--- a/db/template.go
+++ b/db/template.go
@@ -49,7 +49,7 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 func (d TinkDB) GetTemplate(ctx context.Context, fields map[string]string) (string, string, string, error) {
 	getCondition, err := buildGetCondition(fields)
 	if err != nil {
-		return "", "", "", errors.Wrap(err, "failed to build get condition")
+		return "", "", "", errors.Wrap(err, "failed to get template")
 	}
 
 	query := `
--- a/db/workflow.go
+++ b/db/workflow.go
@@ -40,12 +40,12 @@ func (d TinkDB) CreateWorkflow(ctx context.Context, wf Workflow, data string, id
 
 	err = insertActionList(ctx, d.instance, data, id, tx)
 	if err != nil {
-		return errors.Wrap(err, "failed to insert in workflow_state")
+		return errors.Wrap(err, "failed to create workflow")
 
 	}
 	err = insertInWorkflow(ctx, d.instance, wf, tx)
 	if err != nil {
-		return errors.Wrap(err, "Failed to workflow")
+		return errors.Wrap(err, "failed to create workflow")
 
 	}
 	err = tx.Commit()
@@ -109,10 +109,7 @@ func insertActionList(ctx context.Context, db *sql.DB, yamlData string, id uuid.
 
 		workerID, err := getWorkerID(ctx, db, task.WorkerAddr)
 		if err != nil {
-			if errors.Cause(err) == sql.ErrNoRows {
-				return errors.Wrapf(err, "hardware %s not found", task.WorkerAddr)
-			}
-			return err
+			return errors.WithMessage(err, "unable to insert into action list")
 		}
 		workerUID, err := uuid.Parse(workerID)
 		if err != nil {
@@ -690,7 +687,11 @@ func getWorkerIDbyMac(ctx context.Context, db *sql.DB, mac string) (string, erro
 		data @> $1
 	`
 
-	return get(ctx, db, query, arg)
+	id, err := get(ctx, db, query, arg)
+	if errors.Cause(err) == sql.ErrNoRows {
+		err = errors.WithMessage(errors.New(mac), "mac")
+	}
+	return id, err
 }
 
 func getWorkerIDbyIP(ctx context.Context, db *sql.DB, ip string) (string, error) {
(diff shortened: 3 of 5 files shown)


db/db.go
```diff
@@ -102,18 +102,14 @@ func get(ctx context.Context, db *sql.DB, query string, args ...interface{}) (st
...
 	err := row.Scan(&buf)
-	if err == nil {
-		return string(buf), nil
-	}
-
-	if err != sql.ErrNoRows {
-		err = errors.Wrap(err, "SELECT")
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return "", err // if sql.ErrNoRows, return error as is
```
mmlb, line 107: we should wrap this error

db/workflow.go
```diff
@@ -108,7 +108,7 @@ func insertActionList(ctx context.Context, db *sql.DB, yamlData string, id uuid.
 		}
 
 		workerID, err := getWorkerID(ctx, db, task.WorkerAddr)
-		if err != nil {
+		if err != nil && err != sql.ErrNoRows {
```
mmlb, line 111: Can you clear up why its ok to not return an error when err == sql.ErrNoRows, its not immediately obvious to me.
On second thought why don't we return an error in that case?

mmlb (changes requested):

db/workflow.go
kdeng3849: oh sorry it's because `workerID == ""` is basically the same as `sql.ErrNoRows` in this case. But  I realize it's a better idea to switch the order of the check so it checks for ErrNoRows first and then any other error.

mmlb: ah yes I see. Yeah lets handle ErrNoRows first&explicitly so its clearer :D :+1:

db/db.go
kdeng3849: oops i missed this comment. The reason I didn't wrap this error was because `sql.ErrNoRows` is checked here: https://github.com/tinkerbell/tink/blob/71a1c33b0753bec7f501e00cd64e882742f286e6/db/workflow.go#L112

If I wrap the error, it becomes harder to check (unless you have any suggestions).

mmlb: ah true

db/template.go
```diff
@@ -49,7 +49,7 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 func (d TinkDB) GetTemplate(ctx context.Context, fields map[string]string) (string, string, string, error) {
 	getCondition, err := buildGetCondition(fields)
 	if err != nil {
-		return "", "", "", errors.Wrap(err, "failed to build get condition")
+		return "", "", "", errors.Wrap(err, "failed to get template")
```
mmlb, line 52: this should actually be `errors.WithMessage` not wrap but we don't need to change that now. Using errors.Wrap on an already Wrapped error will show up with a new stack trace for each `Wrap` which isn't really what we want. `WithMessage` just adds more messages to the original stack trace.

db/workflow.go
```diff
@@ -109,9 +109,10 @@ func insertActionList(ctx context.Context, db *sql.DB, yamlData string, id uuid.
 
 		workerID, err := getWorkerID(ctx, db, task.WorkerAddr)
 		if err != nil {
+			if err == sql.ErrNoRows {
```
mmlb, line 112: this should be `if errors.Cause(err) == sql.ErrNoRows {` right?

db/db.go
```diff
@@ -102,18 +102,18 @@ func get(ctx context.Context, db *sql.DB, query string, args ...interface{}) (st
...
-
-	if err != sql.ErrNoRows {
-		err = errors.Wrap(err, "SELECT")
+	if err != nil {
+		if err == sql.ErrNoRows {
+			return "", errors.Wrap(func() error {
+				return func() error {
+					return err
+				}()
+			}(), "hardware not found")
```
mmlb, line 111: what is going on here? :D. Why isn't this just `return "", errors.Wrap(err, "hardware not found")`?

mmlb (changes requested):

db/db.go
kdeng3849: This is implementing the method for the `Cause` to work, basically making the `sql.ErrNoRows` error the "cause"

kdeng3849: Should I just create a global var with the value `HardwareNotFound` and then check for that instead?

db/workflow.go
kdeng3849: Right... I put the check for errors.Cause somewhere else 🤦‍♀️

db/workflow.go
```diff
@@ -109,9 +109,10 @@ func insertActionList(ctx context.Context, db *sql.DB, yamlData string, id uuid.
 
 		workerID, err := getWorkerID(ctx, db, task.WorkerAddr)
 		if err != nil {
+			if errors.Cause(err) == sql.ErrNoRows {
+				return fmt.Errorf("hardware mentioned with reference %s not found", task.WorkerAddr)
```
mmlb, line 113: why not `errors.WithMessage() here instead of dropping the stack frame and all the goodness from errors.Wrap?

db/workflow.go
```diff
@@ -689,7 +687,11 @@ func getWorkerIDbyMac(ctx context.Context, db *sql.DB, mac string) (string, erro
 		data @> $1
 	`
 
-	return get(ctx, db, query, arg)
+	id, err := get(ctx, db, query, arg)
+	if errors.Cause(err) == sql.ErrNoRows {
+		err = errors.WithMessagef(errors.New(mac), "mac")
```
mmlb, line 692: errors.WithMessage

mmlb (changes requested):
one *last* change and we're done!

db/workflow.go
kdeng3849: oops

mmlb (approved):

PR #369, opened 2020-11-17
Fix #318 : Starts the db transaction after validating the data

### LINKED ISSUE #318: Open transactions after validating data
## Expected Behaviour

As validation code do not access the database, it can be executed outside transaction to reduce transaction time.

## Current Behaviour

Currently transaction is opened before data is validated, which extends lock time on database, which increases the latency on multiple requests: https://github.com/tinkerbell/tink/blob/master/db/template.go#L17-L30.

## Possible Solution

Validate data first, then open transaction.

hellozee:
## Description
Moves the code for data validation before creating the db transaction.

## Why is this needed
Fixes: #318 

## How Has This Been Tested?
Ran `make test` with Linux host.

### DIFF
--- a/db/template.go
+++ b/db/template.go
@@ -14,14 +14,14 @@ import (
 
 // CreateTemplate creates a new workflow template
 func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error {
+	_, err := wflow.Parse([]byte(data))
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


gianarb:
Thank you for your contribution @hellozee ! Do you mind to rebase? @kdeng3849 noticed a fix in our CI system and it is not fixed https://github.com/tinkerbell/tink/pull/370 this should make CI pass for you as well

hellozee:
Done, :3

gianarb (approved):

gianarb (approved):

PR #371, opened 2020-11-23
Fix #363: Creation of template with the same name

### LINKED ISSUE #363: Creating a template sharing the same name as a deleted template causes an error
## Expected Behaviour

Templates can be successfully created even when it shares the same name as a deleted template.

## Current Behaviour

Due to the fact that templates are soft deleted and the recent change of name also being a unique field, creating a template sharing the same name as a deleted template returns this error:
```
rpc error: code = Unknown desc = INSERT: pq: duplicate key value violates unique constraint "template_name_key"
```

1. create a unique constraint on name (if db isn't using the [latest tink migrations](https://github.com/tinkerbell/tink/blob/master/db/migration/202009171251-init-database.go#L22))
```
alter table template add constraint template_name_key unique (name);
```
2. create a workflow with a name
3. delete the workflow
4. create another workflow with the same name as in step 1

## Context

At first, I thought the soft delete was to be able to get workflow data even after the associated template has been (soft) deleted, but this is not the case. Therefore, what exactly is the purpose of soft deleting?

Granted, there is a workaround for this (e.g. use a UUID as part of the name), but it seems unintuitive that new templates have to take deleted templates into consideration. Also, the only way to see what names have already been used is going into the db.

## Your Environment

Ubuntu 18.04

- Packet using Terraform (sandbox)
- running all the latest images as of 11/11
- added a unique constraint on template name in the db to match what's currently in the [tink migrations](https://github.com/tinkerbell/tink/blob/master/db/migration/202009171251-init-database.go#L22)
```
alter table template add constraint template_name_key unique (name);
```

* Link to your project or a code example to reproduce issue:
### Reproducing the error
```
/ # tink template create -n hello -p hello-world.yml
Created Template:  3119c312-2465-11eb-84af-0242ac130004
/ # tink template delete 3119c312-2465-11eb-84af-0242ac130004
/ # tink template create -n hello -p hello-world.yml
2020/11/11 21:31:18 rpc error: code = Unknown desc = INSERT: pq: duplicate key value violates unique constraint "template_name_key"
```

### Getting a workflow with a deleted associated template
```
/ # tink template list
+--------------------------------------+-----------------+-------------------------------+-------------------------------+
| TEMPLATE ID                          | TEMPLATE NAME   | CREATED AT                    | UPDATED AT                    |
+--------------------------------------+-----------------+-------------------------------+-------------------------------+
| f1e780de-2460-11eb-ad81-0242ac130004 | hello-new-world | 2020-11-11 21:00:35 +0000 UTC | 2020-11-11 21:00:35 +0000 UTC |
+--------------------------------------+-----------------+-------------------------------+-------------------------------+
/ # tink workflow list
+--------------------------------------+--------------------------------------+-----------------------------------+-------------------------------+-------------------------------+
| WORKFLOW ID                          | TEMPLATE ID                          | HARDWARE DEVICE                   | CREATED AT                    | UPDATED AT                    |
+--------------------------------------+--------------------------------------+-----------------------------------+-------------------------------+-------------------------------+
| d3de4f6a-246c-11eb-933f-0242ac130004 | f1e780de-2460-11eb-ad81-0242ac130004 | {"device_1": "b4:96:91:5f:a9:50"} | 2020-11-11 22:25:39 +0000 UTC | 2020-11-11 22:25:39 +0000 UTC |
+--------------------------------------+--------------------------------------+-----------------------------------+-------------------------------+-------------------------------+
/ # tink workflow get d3de4f6a-246c-11eb-933f-0242ac130004
version: 0.1
name: hello_world
global_timeout: 6000
tasks:
- name: "hello-world"
  worker: "b4:96:91:5f:a9:50"
  actions:
  - name: "hello-world"
    image: hello-world
    timeout: 90

/ # tink template delete f1e780de-2460-11eb-ad81-0242ac130004
/ # tink workflow get d3de4f6a-246c-11eb-933f-0242ac130004
2020/11/11 23:01:12 rpc error: code = Unknown desc = failed to get template with ID: f1e780de-2460-11eb-ad81-0242ac130004: sql: no rows in result set
```

Original reporter: Matt Burns

parauliya:

## Description

This PR will fix allow user to create a template with the same name after deleting it. Before this, it was not possible because of an issue also after deleting a template all the workflow which were created with that template before deleting it will be able to access the template since we template will be deleted softly. 

## Why is this needed

Fixes: #363 

## How Has This Been Tested?
Tested all the scenarios mentioned in the linked issue manually using vagrant setup.

Have written a migration script for db.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -37,6 +37,7 @@ type hardware interface {
 type template interface {
 	CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
 	GetTemplate(ctx context.Context, fields map[string]string) (string, string, string, error)
+	GetTemplateForWorkflow(ctx context.Context, fields map[string]string) (string, string, string, error)
 	DeleteTemplate(ctx context.Context, name string) error
 	ListTemplates(in string, fn func(id, n string, in, del *timestamp.Timestamp) error) error
 	UpdateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
--- a/db/db.go
+++ b/db/db.go
@@ -36,8 +36,7 @@ type hardware interface {
 
 type template interface {
 	CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
-	GetTemplate(ctx context.Context, fields map[string]string) (string, string, string, error)
-	GetTemplateForWorkflow(ctx context.Context, fields map[string]string) (string, string, string, error)
+	GetTemplate(ctx context.Context, fields map[string]string, deleted bool) (string, string, string, error)
 	DeleteTemplate(ctx context.Context, name string) error
 	ListTemplates(in string, fn func(id, n string, in, del *timestamp.Timestamp) error) error
 	UpdateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
@@ -115,7 +114,7 @@ func get(ctx context.Context, db *sql.DB, query string, args ...interface{}) (st
 func buildGetCondition(fields map[string]string) (string, error) {
 	for column, field := range fields {
 		if field != "" {
-			return fmt.Sprintf("%s = '%s' AND", column, field), nil
+			return fmt.Sprintf("%s = '%s'", column, field), nil
 		}
 	}
 	return "", errors.New("one GetBy field must be set to build a get condition")
--- a/db/migration/202012041103-remove-unique-constraints.go
+++ b/db/migration/202012041103-remove-unique-constraints.go
@@ -1,0 +1,12 @@
+package migration
+
+import migrate "github.com/rubenv/sql-migrate"
+
+func Get202012041103() *migrate.Migration {
+	return &migrate.Migration{
+		Id: "202010221010-remove-unique-index",
+		Up: []string{`
+                ALTER TABLE template DROP CONSTRAINT template_name_key;
+                `},
+	}
+}
--- a/db/template.go
+++ b/db/template.go
@@ -14,6 +14,12 @@ import (
 
 // CreateTemplate creates a new workflow template
 func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error {
+	fields := map[string]string{
+		"name": name,
+	}
+	if d.getTemplateByName(ctx, fields) {
+		return errors.New("Template with name '" + name + "' already exist")
+	}
 	_, err := wflow.Parse([]byte(data))
 	if err != nil {
 		return err
@@ -23,7 +29,6 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 	if err != nil {
 		return errors.Wrap(err, "BEGIN transaction")
 	}
-
 	_, err = tx.Exec(`
 	INSERT INTO
 		template (created_at, updated_at, name, data, id)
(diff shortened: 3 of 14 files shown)


deploy/db/tinkerbell-init.sql
```diff
@@ -13,7 +13,7 @@ CREATE INDEX IF NOT EXISTS idxgin_type ON hardware USING GIN (data JSONB_PATH_OP
 
 CREATE TABLE IF NOT EXISTS template (
         id UUID UNIQUE NOT NULL
-        , name VARCHAR(200) UNIQUE NOT NULL
+        , name VARCHAR(200) NOT NULL
```
gianarb, line 16: This is not used anymore. We use migrations now

https://github.com/tinkerbell/tink/tree/master/db/migration

You have to create a new migration file that ALTERS the current table

parauliya: Done

db/template.go
```diff
@@ -14,7 +14,14 @@ import (
 
 // CreateTemplate creates a new workflow template
 func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error {
-	_, err := wflow.Parse([]byte(data))
```
gianarb, line 17: Why are you doing this?! If the template is not valid we should not go further. Can you revert or explain why you moved this down there?

db/db.go
```diff
@@ -37,6 +37,7 @@ type hardware interface {
 type template interface {
 	CreateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
 	GetTemplate(ctx context.Context, fields map[string]string) (string, string, string, error)
+	GetTemplateForWorkflow(ctx context.Context, fields map[string]string) (string, string, string, error)
```
gianarb, line 40: I don't want to expose another API if not needed. Can you make `ListTemplates` smarter and remove this call?

gianarb (changes requested):
A few comments and tests have to be written for this PR

db/template.go
parauliya: Yes you are right. I moved it down by mistake. Will fix it.

db/template.go
```diff
@@ -74,6 +80,35 @@ func (d TinkDB) GetTemplate(ctx context.Context, fields map[string]string) (stri
...
+	if err != nil {
+		return "", "", "", errors.Wrap(err, "failed to get template")
+	}
+
+	query := `
+	SELECT id, name, data
+	FROM template
+	WHERE
+		` + getCondition + `
+		created_at IS NOT NULL
```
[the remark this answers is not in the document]
parauliya: This code has been discarded.

db/db.go
parauliya: I have made `GetTemplate` API smarter and removed this new API.

db/template.go
parauliya: Done

db/db.go
parauliya: Done

gianarb (dismissed):
It is good for me. Thank you for taking the time to add coverage.

I have to say, the more I look at the database package, and more I think we are doing something wrong in there. I would really like to write it in a more consistent way. But it is not part of this work for sure :D

kdeng3849 (approved):

db/migration/202012041103-remove-unique-constraints.go
```diff
@@ -0,0 +1,12 @@
+package migration
+
+import migrate "github.com/rubenv/sql-migrate"
+
+func Get202012041103() *migrate.Migration {
+	return &migrate.Migration{
+		Id: "202010221010-remove-unique-index",
```
ryli17, line 7: The id is wrong. Please use the same timestamp in the function name.

db/template.go
```diff
@@ -19,11 +19,17 @@ func (d TinkDB) CreateTemplate(ctx context.Context, name string, data string, id
 		return err
 	}
 
+	fields := map[string]string{
+		"name": name,
+	}
+	_, _, _, err = d.GetTemplate(ctx, fields, false)
```
ryli17, line 25: This query is not in the same transaction as the following insert DML, thus is not concurrent safe.

PR #383, opened 2020-12-10
Fix migration id for template name key

gianarb:
Introduced by https://github.com/tinkerbell/tink/pull/371 the migration
has a wrong ID.


parauliya (approved):
LGTM

ryli17:
DB migration from previous version to this version may panic, due to unknown migration "202010221010-remove-unique-index".
Workaround:
1. "UPDATE gorp_migrations SET id='202012041103-template-with-same-name-are-acceptable' WHERE id='202010221010-remove-unique-index'.
2. Or rebuild db.

Issue #374, opened 2020-12-03
Get a passing CII Best Practices Badge


caniszczyk:
https://bestpractices.coreinfrastructure.org/en

PR #384, opened 2020-12-11
CII Best Practices Badge

gauravgahlot:
Closes: #374


gianarb:
Screenshot 2020-12-11 at 13 01 11

Thank you for your contribution! Do you know why there are 2 tink projects submitted that look the same?
And there a few sticks that IMHO should be green, like the documentation for an example. Is it something that has to be submitted?

gauravgahlot:
@gianarb 
- I have removed the duplicate project. That was because I had submitted `tink` like a month ago and I forgot it. So I unknowingly resubmitted it.
 - Yeah, there are a few that should be green and can be marked manually. So, I will do that. Maybe we can get on a call and do it at once. 
- I don't see an option to add another owner.

gianarb (approved):

PR #385, opened 2020-12-11
Fix #377: Removing name flag from the tink template create/update command

### LINKED ISSUE #377: tink-cli should parse template name from the template itself
Why do we have a template name cli arg when `name` is a require template field? `tink-cli` should use the name in the template. Doing otherwise can lead to lots of confusion.

parauliya:
## Description

This PR will remove the `name` flag from the `tink template create/update` command in tink-cli.

## Why is this needed

Fixes: #377 

## How Has This Been Tested?

This has been tested manually over the vagrant setup.

## How to migrate?

**Current Behaviour**:
At present, the `tink template create -n <name>` allows users to provide a name to a workflow template. A user can also update this name using the `tink template update -n <new-name>` command.

The workflow template also has the `name` field. So, a workflow is getting itss name from two different and places and this can be confusing to users.
```
version: "0.1"
name: hello_world_workflow
global_timeout: 6
tasks:
  - name: "hello world"
    worker: "{{.device_1}}"
    actions:
      - name: "hello_world"
        image: hello-world
        timeout: 60
```

**New Behaviour**:
With this PR, `tink template create` and `tink template update` CLI will _not_ accept a name for the template to be created or updated. Instead, the template name will be picked from the `name` field in the template definition. 

The existing users need to run the `tink template update <id> --file <template-file>` command. This will update the name of each template as per the value of `name` field in the template definition.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/tink-cli/cmd/template/create.go
+++ b/cmd/tink-cli/cmd/template/create.go
@@ -2,6 +2,7 @@ package template
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -12,21 +13,25 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/tinkerbell/tink/client"
 	"github.com/tinkerbell/tink/protos/template"
+	"gopkg.in/yaml.v2"
 )
 
 var (
 	fPath        = "path"
-	fName        = "name"
 	filePath     string
 	templateName string
 )
+
+type TemplateName struct {
+	Name string `yaml:"name"`
+}
 
 // createCmd represents the create subcommand for template command
 var createCmd = &cobra.Command{
 	Use:   "create",
 	Short: "create a workflow template ",
 	Example: `tink template create [flags]
-cat /tmp/example.tmpl | tink template create -n example`,
+cat /tmp/example.tmpl | tink template create `,
 	PreRunE: func(c *cobra.Command, args []string) error {
 		if !isInputFromPipe() {
 			path, _ := c.Flags().GetString(fPath)
@@ -69,8 +74,6 @@ func readAll(reader io.Reader) []byte {
 func addFlags() {
 	flags := createCmd.PersistentFlags()
 	flags.StringVarP(&filePath, "path", "p", "", "path to the template file")
-	flags.StringVarP(&templateName, "name", "n", "", "unique name for the template (alphanumeric and case sensitive)")
-	_ = createCmd.MarkPersistentFlagRequired(fName)
 }
 
 func tryParseTemplate(data string) error {
@@ -78,7 +81,16 @@ func tryParseTemplate(data string) error {
 	if _, err := tmpl.Parse(data); err != nil {
 		return err
 	}
-	return nil
+	var templ TemplateName
+	err := yaml.Unmarshal([]byte(data), &templ)
+	if err != nil {
+		return err
+	}
+	if templ.Name != "" {
+		templateName = templ.Name
+		return nil
+	}
+	return errors.New("Template does not have `name` field which is mandatory")
 }
 
 func createTemplate(data []byte) {
--- a/cmd/tink-cli/cmd/template/create.go
+++ b/cmd/tink-cli/cmd/template/create.go
@@ -8,35 +8,29 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
-	tt "text/template"
 
 	"github.com/spf13/cobra"
 	"github.com/tinkerbell/tink/client"
 	"github.com/tinkerbell/tink/protos/template"
-	"gopkg.in/yaml.v2"
+	"github.com/tinkerbell/tink/workflow"
 )
 
 var (
-	fPath        = "path"
-	filePath     string
-	templateName string
+	file     = "file"
+	filePath string
 )
-
-type TemplateName struct {
-	Name string `yaml:"name"`
-}
 
 // createCmd represents the create subcommand for template command
 var createCmd = &cobra.Command{
 	Use:   "create",
 	Short: "create a workflow template ",
 	Example: `tink template create [flags]
-cat /tmp/example.tmpl | tink template create `,
+cat /tmp/example.tmpl | tink template create -n example`,
 	PreRunE: func(c *cobra.Command, args []string) error {
 		if !isInputFromPipe() {
-			path, _ := c.Flags().GetString(fPath)
+			path, _ := c.Flags().GetString(file)
 			if path == "" {
-				return fmt.Errorf("either pipe the template or provide the required '--path' flag")
+				return errors.New("either pipe the template or provide the required '--file' flag")
 			}
 		}
 		return nil
@@ -55,10 +49,11 @@ cat /tmp/example.tmpl | tink template create `,
 
 		data := readAll(reader)
 		if data != nil {
-			if err := tryParseTemplate(string(data)); err != nil {
+			wf, err := workflow.Parse(data)
+			if err != nil {
 				log.Fatal(err)
 			}
-			createTemplate(data)
+			createTemplate(wf.Name, data)
 		}
 	},
 }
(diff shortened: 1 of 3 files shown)


gianarb:
This is a BC break and I think we should take this opportunity to carefully review our UX.

This is what I mean in practice: `tink hardware push` uses `-f` to require the location for a file, here we use `-p` for the same thing. Should we move it to `-f` in gain some sort consistency?

gauravgahlot:
> This is a BC break and I think we should take this opportunity to carefully review our UX.
> 
> This is what I mean in practice: `tink hardware push` uses `-f` to require the location for a file, here we use `-p` for the same thing. Should we move it to `-f` to gain some sort of consistency?

`tink hardware push` does not have a shorthand for file path and uses `--file`. `-f` is used for the facility. 
```
Usage:
  tink hardware push [flags]

Examples:
cat /tmp/data.json | tink hardware push
tink hardware push --file /tmp/data.json

Flags:
      --file string   hardware data file
  -h, --help          help for push

Global Flags:
  -f, --facility string   used to build grpc and http urls

either pipe the data or provide the required '--file' flag
```

That's the reason why we had to use `-p` for the file path.

gianarb:
Thank you for clarifying @gauravgahlot but my point is the same:

hardware use `--file` when asking for the location for a file.
template uses `-p` for the same reason.

Should we use the same one? 👍 👎

gauravgahlot:
> Thank you for clarifying @gauravgahlot but my point is the same:
> 
> hardware use `--file` when asking for the location for a file.
> template uses `-p` for the same reason.
> 
> Should we use the same one?

Oh yes totally. Initially, we had thought of removing `facility` altogether and use `-f` for the file path. But I'm assuming the facility is required in the Packet environment. 🤔

gianarb:
> But I'm assuming the facility is required in the Packet environment

Never say to me that something is required for ONE single company.. :) but yeah for now I think we should use `--file` and leave the facility, probably when @mmlb will start removing Packet or dead code from all the repo `-f/--facility` will go away, but out of scope for now.

cmd/tink-cli/cmd/template/create.go
```diff
@@ -12,21 +13,25 @@ import (
...
 )
 
 var (
 	fPath        = "path"
-	fName        = "name"
 	filePath     string
 	templateName string
 )
 
+type TemplateName struct {
```
gauravgahlot, line 25: Should not be an exported type.

cmd/tink-cli/cmd/template/create.go
```diff
@@ -12,21 +13,25 @@ import (
...
+	Name string `yaml:"name"`
+}
+
 // createCmd represents the create subcommand for template command
 var createCmd = &cobra.Command{
 	Use:   "create",
 	Short: "create a workflow template ",
 	Example: `tink template create [flags]
-cat /tmp/example.tmpl | tink template create -n example`,
+cat /tmp/example.tmpl | tink template create `,
```
gauravgahlot, line 34: Can we please add more examples in the `Long` description?
For reference, you can check [`tink events watch`](https://github.com/tinkerbell/tink/blob/master/cmd/tink-cli/cmd/events/watch.go#L70).

cmd/tink-cli/cmd/template/create.go
```diff
@@ -69,16 +74,23 @@ func readAll(reader io.Reader) []byte {
...
+	var templ TemplateName
+	err := yaml.Unmarshal([]byte(data), &templ)
+	if err != nil {
+		return err
+	}
+	if templ.Name != "" {
+		templateName = templ.Name
+		return nil
+	}
+	return errors.New("Template does not have `name` field which is mandatory")
```
gauravgahlot, line 93: Please move all the error messages to a `const` and ensure they all start with a lower case.

gauravgahlot (changes requested):

mmlb:
> > But I'm assuming the facility is required in the Packet environment
> 
> Never say to me that something is required for ONE single company.. :) but yeah for now I think we should use `--file` and leave the facility, probably when @mmlb will start removing Packet or dead code from all the repo `-f/--facility` will go away, but out of scope for now.

Facility is only really used for the cert generation stuff, something I'd actually like to get rid of. Let me open up an issue for that (#389).

gianarb (changes requested):
Great, @parauliya do you mind changing `-p` with `--file` and we are good? Thanks

mmlb (approved):

cmd/tink-cli/cmd/template/create.go
mmlb, line 93: This should use `github.com/tinkerbell/tink/workflow.Parse` instead.

mmlb (changes requested):

cmd/tink-cli/cmd/template/create.go
```diff
@@ -69,16 +74,23 @@ func readAll(reader io.Reader) []byte {
 func addFlags() {
 	flags := createCmd.PersistentFlags()
 	flags.StringVarP(&filePath, "path", "p", "", "path to the template file")
```
gianarb, line 76: Please change this with `--file` to unify with `hardware push`

gianarb (changes requested):
This is a bc break, according with the proposal 0011 https://github.com/tinkerbell/proposals/blob/master/proposals/0011/README.md can you add a chapter in the PR description:

# How to migrate

https://github.com/tinkerbell/proposals/blob/master/proposals/0011/README.md#how-to-migrate

Explaining what is not working anymore and how to do the same in the new way, thanks

gauravgahlot (dismissed):

cmd/tink-cli/cmd/template/create.go
```diff
@@ -29,9 +28,9 @@ var createCmd = &cobra.Command{
 cat /tmp/example.tmpl | tink template create -n example`,
 	PreRunE: func(c *cobra.Command, args []string) error {
 		if !isInputFromPipe() {
-			path, _ := c.Flags().GetString(fPath)
+			path, _ := c.Flags().GetString(file)
 			if path == "" {
-				return fmt.Errorf("either pipe the template or provide the required '--path' flag")
+				return errors.New("either pipe the template or provide the required '--path' flag")
```
gianarb, line 33: I think this is `'--file' flag` now

gianarb (changes requested):

gianarb (approved):

gauravgahlot:
No code changes only updated the branch.

gianarb (approved):

gauravgahlot (approved):

gianarb (approved):

gauravgahlot:
@mmlb Please let me know once you are back, and I will update the branch (one last time). 😄

cmd/tink-cli/cmd/template/create.go
```diff
@@ -2,36 +2,41 @@ package template
...
+	"github.com/tinkerbell/tink/workflow"
 )
 
 var (
-	fPath        = "path"
-	fName        = "name"
-	filePath     string
-	templateName string
+	file     = "file"
+	filePath string
```
mmlb, line 20: I don't quite see the reason for `file` global var when its not used in the `flags.StringVarP` call. It should either be named a little more descriptively (filePathArgName?) and used there too or we just hard code use of `"file"` every where.

I'm not a fan of the "variable for the variable name" approach...

cmd/tink-cli/cmd/template/create.go
```diff
@@ -2,36 +2,41 @@ package template
...
+# Pipe the file to create a template:
+$ cat /tmp/example.tmpl | tink template create
+
+# Create template using the --file flag:
+$ tink template create --file /tmp/example.tmpl
+`,
 	PreRunE: func(c *cobra.Command, args []string) error {
 		if !isInputFromPipe() {
-			path, _ := c.Flags().GetString(fPath)
+			path, _ := c.Flags().GetString(file)
```
mmlb, line 37: path isn't ever being used right?

cmd/tink-cli/cmd/template_test.go
```diff
@@ -61,8 +61,8 @@ func Test_templateCmd(t *testing.T) {
 				if err := root.Execute(); err != nil {
 					t.Error(err)
 				}
-				if !strings.Contains(out.String(), "create a workflow template") {
-					t.Error("expected output should include create a workflow template")
+				if !strings.Contains(out.String(), "Create template using the --file flag") {
```
mmlb, line 64: can be DRY'd by doing:
```go
want := "Create template using the --file flag"
if !strings.Contains(out.String(), want) {
    t.Error("unexpected output, looking for %q as a substring in %q", want, out.String())
}
```
or similar

cmd/tink-cli/cmd/template_test.go
```diff
@@ -112,8 +112,8 @@ func Test_templateCmd(t *testing.T) {
 				if err := root.Execute(); err != nil {
 					t.Error(err)
 				}
-				if !strings.Contains(out.String(), "update a template") {
-					t.Error("expected output should include update a template")
+				if !strings.Contains(out.String(), "Update an existing template") {
```
mmlb, line 115: ditto

mmlb (changes requested):

gauravgahlot (dismissed):

gianarb (approved):

gauravgahlot (approved):

mmlb (approved):

PR #386, opened 2020-12-11
Remove global logger from db package

gianarb:
I am working on a integration test suite for the db package. And the
global logger is not an ideal approach in any case.

This PR moves the logger from package level to TinkDB


thebsdbox (approved):
Nice, is the plan here to share the tinkDB.Logger ?

/lgtm

PR #387, opened 2020-12-11
Setup integration test suite for db package

gianarb:
## Description

TestContainer is a library that spins up and down containers programmatically inside a test case. It is very comfortable to run integration tests and to set up requirements such as Postgres in our case.

This PR contains the first tests that validate the happy path for the CreateTemplate function.

## Question

Right now in order to separate the tests requiring docker from the one who can run standalone I created a separate package `./db/_integration` that by default is not picked up by `go test ./...` 

The thread off here is that code coverage does not count the new package, even if it will be nice to have it included. A possible solution is to add a flag like:

```
go test --with-docker ./...
```

if the flag is set the tests requiring docker will run. In this way, we should have code coverage counting integration tests.

## Why is this needed

Because it is better to refactor a package when it is covered by tests.


db/_integration/db_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+CHECK_DB:
+	for ii := 0; ii < 5; ii++ {
+		err = dbCon.Ping()
+		if err != nil {
+			t.Log(errors.Wrap(err, "db check"))
+			time.Sleep(1 * time.Second)
+			goto CHECK_DB
+		}
```
mmlb, line 64: this kind of doesn't make any sense right? If not ready go back to before the loop forever. If it is ready... make sure its ready 5 times?

db/_integration/db_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+// NewPostgresDatabaseClient returns a SQL client ready to be used. Behind the
+// scene it is starting a Docker container that will get cleaned up when the
+// test is over. Tests using this function are safe to run in parallel
+func NewPostgresDatabaseClient(t *testing.T, ctx context.Context, req NewPostgresDatabaseRequest) (*sql.DB, *db.TinkDB, func() error) {
+	postgresC, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: testcontainers.ContainerRequest{
+			Image:        "postgres:13.1",
+			ExposedPorts: []string{"5432/tcp"},
+			WaitingFor:   wait.ForLog("database system is ready to accept connections"),
```
mmlb, line 29: we should add support to wait for something like wait.Exec and then we can just call `pg_isready`.

workflow/template_validator.go
```diff
@@ -31,6 +33,26 @@ func Parse(yamlContent []byte) (*Workflow, error) {
 	return &workflow, nil
 }
 
+// MustParse parse a slice of bytes to a template. It an error occurs the
```
mmlb, line 36: can you add tests for these so codecov is happy?

mmlb (changes requested):
lgtm so far. Lets just set the tests up to run by default and call t.Skip() if it detects that docker is not available.

db/db_test.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	for ii := 0; ii < 5; ii++ {
+		err = dbCon.Ping()
+		if err != nil {
+			t.Log(errors.Wrap(err, "db check"))
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		break
```
mmlb, line 65: could have been:
```go
if err == nil {
	break
}
t.Log...
```
too :D

mmlb (approved):

mmlb (approved):

PR #393, opened 2020-12-15
tink template watch

gauravgahlot:
The user can now watch for events using the `tink template watch` function.

### DIFF
--- a/cmd/tink-cli/cmd/template/watch.go
+++ b/cmd/tink-cli/cmd/template/watch.go
@@ -26,7 +26,7 @@ import (
 // watchCmd represents the watch command
 var watchCmd = &cobra.Command{
 	Use:     "watch",
-	Short:   "register to watch an id for any changes",
+	Short:   "watch for events of given template(s)",
 	Example: "tink template watch 224ee6ab-ad62-4070-a900-ed816444cec0 cb76ae54-93e9-401c-a5b2-d455bb3800b1",
 	Args: func(_ *cobra.Command, args []string) error {
 		if len(args) < 1 {
--- a/cmd/tink-cli/cmd/template/watch.go
+++ b/cmd/tink-cli/cmd/template/watch.go
@@ -1,0 +1,105 @@
+// Copyright © 2018 packet.net
+
+package template
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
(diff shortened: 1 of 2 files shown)


cmd/tink-cli/cmd/template/watch.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/tinkerbell/tink/client/informers"
+	"github.com/tinkerbell/tink/protos/events"
+	"github.com/tinkerbell/tink/workflow"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+// watchCmd represents the watch command
+var watchCmd = &cobra.Command{
+	Use:     "watch",
+	Short:   "register to watch an id for any changes",
```
gianarb, line 29: What does it mean?!

workflow/types.go
```diff
@@ -29,3 +31,13 @@ type Action struct {
...
 }
+
+// Template represents a workflow template
+type Template struct {
+	ID        string    `json:"id,omitempty"`
+	Name      string    `json:"name,omitempty"`
+	CreatedAt time.Time `json:"created_at,omitempty"`
+	UpdatedAt time.Time `json:"updated_at,omitempty"`
+	DeletedAt time.Time `json:"deleted_at,omitempty"`
+	Data      string    `json:"data,omitempty"`
```
gianarb, line 42: Technically speaking we have a Workflow type in `workflow/types.go`, I think `Data` should be that type, can you double check? Thanks

gianarb:
I would like to see a test that coming from a mock of the raw event returns a formatted response

gianarb:
I am sorry if this comment comes probably a bit too late but I just realized it now.

We have a couple of "root" commands that describe the resources tink provides:

```
hardware
workflow
template
event
```

We can watch events already using:

```
tink event watch
```

And we can filter by resource type and event type.

Now we are adding `watch` command to any resource. This new set of commands we are opening will have to be maintained and kept consistent across all the resources. Is it something we want? Can we make the `tink event watch` if needed?

Or can we embed the `watch` feature as part of commands we already have like Kubernetes does for example for:

```
kubectl get pod --watch
```

I am not taking any decision but I want to get other people option.

I am OK with the current solution `having watch command for each resource` if we can write a reusable Watch command that can be added to the current resources. It will help to keep consistency in moving forward.

Issue #379, opened 2020-12-10
duplicate key value violates unique constraint "events_pkey" error while creating multiple workflows


invidian:
## Expected Behaviour
Workflows can be created in parallel without errors.

## Current Behaviour
Sometimes, when many workloads are created in parallel, following error is returned:
```
Error: creating workflow: rpc error: code = Unknown desc = failed to create workflow: INSERT in to workflow: pq: duplicate key value violates unique constraint "events_pkey"
```

Logs from server:
```
[...]

PR #395, opened 2020-12-16
Fixed: duplicate key value violates unique constraint while creating multiple workflows

gauravgahlot:
## Why is this needed

Fixes: #379 

## How Has This Been Tested?

1. `git clone https://github.com/kinvolk/terraform-provider-tinkerbell && cd terraform-provider-tinkerbell`
2. build new images with changes in place and update docker-compose file
3. `docker-compose -f test/docker-compose.yml up -d`
4. `TF_ACC=1 TINKERBELL_GRPC_AUTHORITY=127.0.0.1:42113 TINKERBELL_CERT_URL=http://127.0.0.1:42114/cert go test -v -run=^TestAccWorkflow_.*$ -count 100 ./tinkerbell/`

### DIFF
--- a/db/migration/2020121691335-update-events-primary-key.go
+++ b/db/migration/2020121691335-update-events-primary-key.go
@@ -1,0 +1,24 @@
+package migration
+
+import migrate "github.com/rubenv/sql-migrate"
+
+// Get2020121691335 updates the primary key on events table.
+//
+// Fixes: https://github.com/tinkerbell/tink/issues/379
+//
+// We can have multiple events generated at a given time, therefore the value of
+// 'created_at' field will be same for each event. This violates the unique constraint
+// on "events_pkey", when these events are add to the events table.
+//
+// The migration changes the primary key on events table from 'created_at' to `id`,
+// which will always be unique for each event generated at any point in time.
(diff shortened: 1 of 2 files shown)


gianarb:
@gauravgahlot I think you have to create a separate migration :) so if we have people running the previous version they will be able to move from the master to this one

gianarb (changes requested):
@gauravgahlot I think you have to create a separate migration :) so if we have people running the previous version they will be able to move from the master to this one

invidian (commented):
I didn't get time to test it yet, but from what I've seen, it should fix the issue. Scheme also looks sensible, though some explanation in commit message would be nice, so one don't have to figure it out on their own.

gianarb:
Reading the comment @invidian said I think it will be nice to actually have that documentation as a comment in the migration function itself! ❤️

mmlb:
> Reading the comment @invidian said I think it will be nice to actually have that documentation as a comment in the migration function itself! heart

This is probably a good idea in general for all migrations :D

invidian (approved):
LGTM

gianarb (approved):

Issue #405, opened 2020-12-28
Tink CLI current state and the future is called consistency


gianarb:
Hello

The CLI is important because it is the entry point many people will use to onboard Tinkerbell. If the CLI is not great, they will even stop using Tinkerbell. Or at least that's how I approach an ecosystem where there is a new cool project to use every day.

# Consistency

I vote tink CLI with a 3 out of 10 currently. There is very low consistency in what you can do with a resource (template, event, hardware...). Some of them have `get,` others have `list` others have `all` as a subcommand. This isn't good already, but I can even go further. We use `--file` to pass a file when pushing new hardware; we use `-p` to create a template #270. Same thing, different flag. Same action (create) different name push or create.

Even when the behavior is consistent, it is just luck or a memory exercise left to the contributor or the reviewer that has to remember or figure out if flags and commands are in some way consistent with the current state of CLI. This isn't good. Consistency has to be built in the code itself in the form of structure and abstraction.

--- 

The tink CLI uses Cobra, and it is structured with resources as root level: `events,` `hardware,` `template,` `workflow`. Each of those resources has its own commands. I think it is an overall good structure. Still, I expect to quickly move between those resources having a consistent structure for common operation (per resource specialization is fine, but it should be an advanced requirement).

Common operations are:

* `get` -> Retrieve a list or a single resource printing important information with the ability to do `filtering` based on what the resource allows me to filter on. It will be nice to be able to specify what to show as well (if you want only the ID #403 , for example)
* `describe` -> Shows a single resource with all its details
* `create` -> Create a new resource
* `delete` -> Delete a resource
* `update` -> Update a resource

Every command that outputs data should use the same rendering components with the ability to:

* Specify what we want to see
* And the output format: JSON, CSV, ps, or table

The right way to enforce consistency is at the code level, which means that create, update, delete, get, describe have to be reusable commands across resources.

---

## Example: Hardware

Current situation

```terminal
Available Commands:
  delete      delete hardware by id
  get
Display one or many resources
  id          get hardware by id
  ip          get hardware by any associated ip
  list        list all known hardware
  mac         get hardware by any associated mac
  push        push new hardware to tink
  watch       register to watch an id for any changes
```

What I would like to get:

* delete
* get (with filtering made in the right way we can deprecate mac, list, id, IP)
* create (in favor of push that can be deprecated)
* update

---

Any thought?

gauravgahlot:
[...]

PR #406, opened 2020-12-28
Bootstrap Get command for tink-cli

gianarb:
## Description

Create a reusable Get command.

TODO:

* [ ] ~Implement `field-selector` to filter by some fields `--filder-selector mac==aa:bb:cc:dd`, `--field-selector template==6b1ad3d5-3dd7-43be-9c08-8cc566cb0ae5`. The goal here is to deprecate the hardware IP and Mac command.~ (we can't do it at this stage because the grpc API is not flexible enough, it will be a nightmare to implement)
* [x] Create a structure in the client pkg that contains all the grcp clients and pass it to `get.CmdOpt` instead of using the global clients
* [x] Write tests for the get command in hardware, template, and workflow
* [x] Have a look at how to use Get for events
* [x] Implement populate with args

## Why is this needed

Related to #405

A good way to gain consistency across a CLI is to write code that enforces that. Right now we rewrite commands across resources and this practice is wrong.

This PR bootstraps a Get command that can be used across all the resources.

## How Has This Been Tested?

Unit tests

## How to migrate?

This PR breaks two commands:

```
$ tink template get
$ tink workflow get
```

You can get the same behavior setting the environment variable `TINK_CLI_VERSION=0.0.0`. Please, we will remove this workaround very soon, let us know what you have missed with the new command and we will try to give you an alternative that works with the new CLI we have in our mind!

There are a couple of deprecated commands that print to stderr the alternative you can use. Please let us know if some of your use cases are not covered yet


thebsdbox:
Nice I love this !

cmd/tink-cli/cmd/workflow/list.go
```diff
@@ -27,6 +27,17 @@ var listCmd = &cobra.Command{
 	Use:     "list",
 	Short:   "list all workflows",
 	Example: "tink workflow list",
+	Deprecated: `This command is deprecated and it will change at some
```
thebsdbox, line 30: I don't know about the `Deprecated:` option in cobra.. is it outputted on STDOUT? My main concern is breaking people scripts?

thebsdbox (approved):
`/lgtm` - I think the idea to sanitise the UX now is a great idea.

client/main.go
```diff
@@ -24,6 +24,47 @@ var (
...
+	TemplateClient template.TemplateServiceClient
+	WorkflowClient workflow.WorkflowServiceClient
+	HardwareClient hardware.HardwareServiceClient
+	EventsClient   events.EventsServiceClient
+}
+
+// NewMetaClientFromGlobal is a dirty hack that returns a MetaClient using the
+// global variables exposed by the client package. Globals should be avoided
+// and we will deprecated them at some point replacing this function with
+// NewMetaClient. If you are strating a new project please use the last one
```
stolsma, line 38: "...are strating..." should probably be starting... ;-)

stolsma:
@gianarb A generic Tink grpc client abstraction is also needed for the web Portal project (and maybe later on also for other projects) as that project doesn't work anymore because of changed Tink grpc code. Is it possible to carve the generic parts of Tink grpc client interactions out to a specific package? I'm referring here to the grpc code in client.go... 

What do you think??

gianarb:
Thank you for your review @thebsdbox ! @stolsma thank you as well I will address your comment next week! 

About the grpc client. The package is already isolated https://github.com/tinkerbell/tink/tree/master/client but I don't like that it uses global functions. This PR makes the first refactoring. I will integrate, as part of this PR a few mocks that I hope will set a common way to use the client in a test suite.

cmd/tink-cli/cmd/workflow/list.go
gianarb: It prints to stdout. Personally, it is a bc break announced in this PR and in Sandbox when we will release it. I would like to remove old-style commands very soon. So, yes it breaks, mitigation is easy, do not update and I hope the `Deprecated` note and documentation will help to move fast to new commands.

stolsma:
@gianarb Yep, I know the Tinkerbell Portal doesn't work at this moment because of the OLD API use and I already created a Issue for that ([https://github.com/tinkerbell/portal/issues/16](https://github.com/tinkerbell/portal/issues/16)) before I found out that you where working on improving the Tink client API ;-)

Love your ideas about the client API direction, that's the direction I was hinting at in my previous comment before I scanned your code so you where already on it :-)   

I would like to take a stab into implementing your improved client API into the Portal code to get that running again. When do you think the code of this pull request will be merged so I can use it there?

gianarb:
Hey @stolsma ! Great, I am very happy to read that you want to help us and make a contribution. You have to give me a couple of days to figure out when @mmlb and a few other people will be back from vacation because as you imagine this is a good chunk of work coming from me directly, I want to discuss it with other contributors first.

Worst case I think we can open a PR with only the Client related work! That said I hope to have the client part merged this week.

client/main.go
```diff
@@ -24,6 +24,47 @@ var (
...
+		TemplateClient: template.NewTemplateServiceClient(conn),
+		WorkflowClient: workflow.NewWorkflowServiceClient(conn),
+		HardwareClient: hardware.NewHardwareServiceClient(conn),
+		EventsClient:   events.NewEventsServiceClient(conn),
+	}
+}
+
 // GetConnection returns a gRPC client connection
 func GetConnection() (*grpc.ClientConn, error) {
 	certURL := os.Getenv("TINKERBELL_CERT_URL")
```
stolsma, line 70: Is it possible to make the client pkg independent of environment variables and hoist them to code using the client pkg code and insert them as arguments in the GetConnection function?? It will make the client pkg more independent of type of use (imo)...

gianarb: It is already like that if you use the function `NewMetaClient` https://github.com/tinkerbell/tink/pull/406/files#diff-39e4af22263d62c01d979fa1f6ec1770d58fc1b03b0aeec2f17ba84e7136c0dbR59 you have to pass a `grpc.ClientConnInterface` to it, the way you make it is totally up to you

gianarb:
This CLI has currently one problem, the grpc connection gets created as part of the `cmd/tink-cli.Main`. it means that no matter what a tink-server has to be up and running. We have to fix it someway, probably with a Lazy client or creating the meta client later

client/main.go
stolsma: @gianarb I know, but then the user of the pkg needs to implement the CERT stuff himself and more or less is duplicating 90% of the code in this function.

To be honest, I'm lazy and spoiled by the k8s client pkg api's... ;-)

gianarb: I am even not understanding or I am doing some wrong assumption. Let me try to explain.

We can discuss if we like how GetConneciton works #174 does it. if you want to get a GRPC connection up without an environment variable you CAN do it with this PR. Look at `GetConnection` as a utility function you can use if you want to have a consistent UX with other Tinkerbell project. If you are speaking about something else or it looks mandatory to you can you elaborate? because maybe I am looking at this problem in the wrong direction

stolsma: Ok, didn't want to confuse you! 

>  Look at GetConnection as a utility function you can use if you want to have a consistent UX with other Tinkerbell project. If you are speaking about something else or it looks mandatory to you can you elaborate?

I see GetConnection as a utility function too but I like to have a more generic kind that doesn't force ENV vars but enables the use of the certificate config. If I don't want to use the specific ENV vars (but with the CERT loading part) I need to implement the whole function, including CERT loading etc myself.

> We can better encapsulate the GetConnection to accept a struct that gets populated via env var, that's probably the right solution but I will probably do it later on in its own chunk of work

Yep, thats the ultimate implementation I was looking for: struct argument for a 'GetConnectionWithConfig' function that also uses the CERT stuff. But I do understand if that is maybe implemented later in its own chunk of work.

Thanks for taking the time to answer me, great job!! 👍

displague:
@gianarb have a look at https://github.com/packethost/packet-cli/pull/74 which made a similar change to a decedent project.

cmd/tink-cli/cmd/get/get.go
```diff
@@ -0,0 +1,100 @@
...
+	PopulateTable func([]interface{}, table.Writer) error
+
+	// Output specifies the output you want the list of resources printed
+	// out. By default it is table but it can be JSON ar CSV.
+	Output string
+	// NoHeaders does not print the header line
+	NoHeaders bool
+}
+
+const shortDescr = `
```
mmlb, line 27: this initial new line makes it so that the short description messes with the help output:
```
./tink-cli hardware --help
tink hardware client
...
Available Commands:
  delete      delete hardware by id
  get         
Display one or many resources
  id          get hardware by id
  ip          get hardware by any associated ip
...
```
If you remove this new line then it looks correct:
```
./tink-cli hardware --help
tink hardware client
...
Available Commands:
  delete      delete hardware by id
  get         Display one or many resources
  id          get hardware by id
  ip          get hardware by any associated ip
...
```

cmd/tink-cli/cmd/get/get.go
```diff
@@ -0,0 +1,100 @@
...
+tink hardware get
+
+# List all workflow in csv output format.
+tink template get -o csv
+
+# List a single template in json output format.
+tink workflow get -o json [id]
+`
+
+func NewGetCommand(opt CmdOpt) *cobra.Command {
```
mmlb, line 47: I kind of would like to have `NewGetCommand` take in a string that it uses to populate the parent in the example.
```go
func NewGetCommand(opt CmdOpt, parent string)
...
   Example: strings.Replace(exampleDescr, "$parent", parent, -1),
...
)
```
 
This way all the examples make the most amount of sense and don't have example for things we aren't trying to do.

cmd/tink-cli/cmd/hardware/get.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/tinkerbell/tink/cmd/tink-cli/cmd/get"
+	"github.com/tinkerbell/tink/protos/hardware"
+)
+
+type GetHardware struct {
+	get.CmdOpt
+	metaClient *client.MetaClient
+}
+
+func NewGetHardwareOpt(metaClient *client.MetaClient) GetHardware {
```
mmlb, line 18:
```suggestion
func NewGetHardware(metaClient *client.MetaClient) GetHardware {
```
So it matches `New$type(...) $type {` pattern

cmd/tink-cli/cmd/get/get.go
```diff
@@ -0,0 +1,100 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/jedib0t/go-pretty/table"
+	"github.com/spf13/cobra"
+)
+
+type CmdOpt struct {
```
mmlb, line 12:
```suggestion
type Options struct {
```
so it looks like `get.Options` and reads better imo.

cmd/tink-cli/cmd/hardware/get.go
```diff
@@ -0,0 +1,62 @@
...
+	return data, nil
+}
+
+func (h *GetHardware) PopulateTable(data []interface{}, t table.Writer) error {
+	for _, v := range data {
+		if hw, ok := v.(*hardware.Hardware); ok {
+			// TODO(gianarb): I think we should
+			// print it better. The hardware is one
+			// even if if has more than one
+			// interface.
```
mmlb, line 52: agree, maybe we can skip the hw.Id for subsequent nics so it looks like a continuation?

cmd/tink-cli/cmd/hardware/ip.go
```diff
@@ -16,38 +16,33 @@ import (
...
-		}
-	},
-}
-
-func addIPFlags() {
-	flags := ipCmd.Flags()
+		},
+	}
+	flags := cmd.Flags()
 	flags.BoolVarP(&data, "details", "d", false, "provide the complete hardware details in json format")
```
mmlb, line 46: can we rename `data` -> `details` while here?

cmd/tink-cli/cmd/hardware/mac.go
```diff
@@ -13,39 +13,33 @@ import (
...
-		}
-	},
-}
-
-func addMacFlags() {
-	flags := macCmd.Flags()
+		},
+	}
+	flags := cmd.Flags()
 	flags.BoolVarP(&data, "details", "d", false, "provide the complete hardware details in json format")
```
mmlb, line 43: ditto `data` -> `details`

cmd/tink-cli/cmd/workflow/list.go
mmlb: @thebsdbox I share your concern over stdout, happily though it seems cobra does the right thing and prints to stderr not stdout.
```
[~/go/src/github.com/tinkerbell/tink/cmd/tink-cli]─[manny@dellnix]> 
./tink-cli  workflow list 2>/dev/null
Exception: ./tink-cli exited with 1
[tty 24], line 1: ./tink-cli  workflow list 2>/dev/null
                                                                                                                                                            
[~/go/src/github.com/tinkerbell/tink/cmd/tink-cli]─[manny@dellnix]> 
./tink-cli  workflow list >/dev/null
Command "list" is deprecated, This command is deprecated and it will change at some
	point. Please try what follows:

	# If you want to retrieve a single workflow you know by ID
	tink workflow get [id]
	# You can print it in JSON and CSV as well
	tink workflow get -o json [id]

	# Get a list of available workflows
	tink workflow get [id]

Error: fetch cert: Get "http://127.0.0.1:42114/cert": dial tcp 127.0.0.1:42114: connect: connection refused
Usage:
  tink workflow list [flags]

Examples:
tink workflow list

Flags:
  -h, --help    help for list
  -q, --quiet   only display workflow IDs

Global Flags:
  -f, --facility string   used to build grpc and http urls

fetch cert: Get "http://127.0.0.1:42114/cert": dial tcp 127.0.0.1:42114: connect: connection refused
Exception: ./tink-cli exited with 1
[tty 25], line 1: ./tink-cli  workflow list >/dev/null
```

cmd/tink-cli/cmd/workflow/list.go
```diff
@@ -27,6 +27,17 @@ var listCmd = &cobra.Command{
...
+	point. Please try what follows:
+
+	# If you want to retrieve a single workflow you know by ID
+	tink workflow get [id]
+	# You can print it in JSON and CSV as well
+	tink workflow get -o json [id]
+
+	# Get a list of available workflows
+	tink workflow get [id]
```
mmlb, line 39: This should just point users to run `tink workflow get` instead or maybe we can print the example somehow?

mmlb (changes requested):

mmlb:
Well that was a doozy :D

cmd/tink-cli/cmd/get/get.go
mmlb: and I just noticed that its got a capital letter when it should be lower case actually

mmlb:
>     * [ ]  Implement `field-selector` to filter by some fields `--filder-selector mac==aa:bb:cc:dd`, --field-selector template==6b1ad3d5-3dd7-43be-9c08-8cc566cb0ae5`. The goal here is to deprecate the hardware IP and Mac command.

I don't think this would be very nice UX, why does Get need to support that at all? The Get implementation can just parse the rest of the args itself to figure out what to filter off of. And if you do the change to `NewGetCommand we can probably also add an extra arg to add filter examples too.

cmd/tink-cli/cmd/hardware/get.go
gianarb: Not sure if I understand it but having 2 lines that belong to the same hardware where the first one has an empty ID makes `awk` or `grep` kind of hard. (overall as discussed elsewhere) I think we should not use the table as a format by default but something like `ps`.

gianarb:
> I don't think this would be very nice UX, why does Get need to support that at all? The Get implementation can just parse the rest of the args itself to figure out what to filter off of. And if you do the change to `NewGetCommand we can probably also add an extra arg to add filter examples too.

@mmlb I think I don't understand what you are saying. Can you help me?

> why does Get need to support that at all?

I would like to combine the various requests we have:

* Get by ID
* Get All
* Get by IP
* Get by MAC

All under the same `get` command. Same as `kubectl` does. The first two are easy:

Get By ID: `tink <resource> get id1 id2 id3`
Get All: `tink <resource> get`

For the other two? (or more if needed)

cmd/tink-cli/cmd/get/get.go
```diff
@@ -0,0 +1,100 @@
...
+resource. You can specify the kind of output you want to receive. It can be
+table or csv or json.
+`
+
+const exampleDescr = `
+# List all hardware in table output format.
+tink hardware get
+
+# List all workflow in csv output format.
+tink template get -o csv
```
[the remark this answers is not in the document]
gianarb: `-f` is file in create or whatever I don't remember. I used `-o` because kubectl uses it. Docker uses `--format`, I am happy to use it but without the short version.

cmd/tink-cli/cmd/get/get.go
```diff
@@ -0,0 +1,100 @@
...
+	// Headers is the list of headers you want to print as part of the list
+	Headers []string
+	// RetrieveData reaches out to Tinkerbell and it gets the required data
+	RetrieveData func(context.Context) ([]interface{}, error)
+	// PopulateTable populates a table with the data retrieved with the RetrieveData function.
+	PopulateTable func([]interface{}, table.Writer) error
+
+	// Output specifies the output you want the list of resources printed
+	// out. By default it is table but it can be JSON ar CSV.
+	Output string
```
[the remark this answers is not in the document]
gianarb: I used `-o/--output` because kubectl uses it. Docker uses `--format` without a short version.

Now, `create` uses or will require a way to pass a file or a path as we are discussing here #270. if you want `--format/-f` we won't be able to use `--file/-f` in creating/update. Any opinion? My suggestions is:

* `--format` without a short version
* `-f/--file` to pass a filepath during create/updater

cmd/tink-cli/cmd/get/get.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+
+			switch opt.Output {
+			case "json":
+				// TODO(gianarb): the table library we use do
+				// not support JSON right now. I am not even
+				// sure I like tables! So complicated...
+				b, err := json.Marshal(struct {
+					Data interface{} `json:"data"`
+				}{Data: data})
```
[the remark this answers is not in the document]
gianarb: I don't know, I am thinking about a future where we will have pagination?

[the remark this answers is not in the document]
mmlb: hmm yeah I see that case. I tend to favor jsonnl and forego pagination and leave that to either filtering things or head/tail/grep. But this is fine too.

cmd/tink-cli/cmd/hardware/get.go
mmlb: yeah I don't think we need to support table for programmatic cli use imo. If you want to throw into a pipeline then just use a better machine parse-able format imo.

cmd/tink-cli/cmd/get/get.go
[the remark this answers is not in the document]
mmlb: Yep I like --format & -f/--file that sounds ok to me. We can also do -F for format too.

client/main.go
mmlb: can we skip that in CI? This feels weird to let CI dictate removal of deprecated functions which removes the utility of marking things Deprecated ... I'd expect CI to ignore deprecated functions from within this module :thinking:

client/main.go
mmlb: /minor note: I *really* want to get rid of all the odd cert handling stuff :D.

mmlb:
> > I don't think this would be very nice UX, why does Get need to support that at all? The Get implementation can just parse the rest of the args itself to figure out what to filter off of. And if you do the change to `NewGetCommand we can probably also add an extra arg to add filter examples too.
> 
> @mmlb I think I don't understand what you are saying. Can you help me?
> 
> > why does Get need to support that at all?
> 
> I would like to combine the various requests we have:
> 
>     * Get by ID
> 
>     * Get All
> 
>     * Get by IP
> 
>     * Get by MAC
> 
> 
> All under the same `get` command. Same as `kubectl` does. The first two are easy:
> 
> Get By ID: `tink <resource> get id1 id2 id3`
> Get All: `tink <resource> get`
> 
> For the other two? (or more if needed)

So I mean to make `Get` pluggable and/or just be the common "interface" of get commands. So in that case `get uuid1 uuid1` or just `get` can be handled by `Get` but anything else that `Get` doesn't understand it passes to a function (like `RetrieveData` does, `ParseGet` or something?)

Does that clear it up?

cmd/tink-cli/cmd/get/get.go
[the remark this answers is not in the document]
displague: I'm also swayed by `kubectl` usage. `-o/--output` or `--format` works for me, I also like `-f/--file`.  Config files are the only other use-case that comes to mind and `--config-something` can work for that.

I can never get `-O` or `-o` straight between `curl` and `wget`, while redirection works in most cases, sometimes it is handy to have an output file argument in cases where you don't control the environment.  I think we can wait for that request to come along before introducing the feature.

The `packet-cli` went with `--json` and `--yaml`, which would have been fine but `-y` and `-j` were added as short names (🤦  mea culpa) which limits our ability to use those later, my advise is to not introduce a short name until you are sure about it.

cmd/tink-cli/cmd/get/get.go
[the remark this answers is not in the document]
displague: I don't expect pagination added by the cli, I think I would expect the raw API response. The `packet-cli` confuses me in this way because it outputs the `packngo` Go structure of the API responses rather than the raw API responses.

Thinking about `kubectl` again, I think they also prefer the raw API responses (which is part of the reason we get managed-fields noise).

client/main.go
gianarb: I was trying to be lazy! I appreciate your attempt to make me a better person!

cmd/tink-cli/cmd/get/get.go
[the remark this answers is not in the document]
displague: I suppose that's not possible when the API does not speak json natively.

cmd/tink-cli/cmd/get/get.go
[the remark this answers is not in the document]
mmlb: I'm ok with -o but its not my preference, if we do go with it I think tacking "format" to it in all its references makes sense (cli -h output, OutputFormat struct field...) it just seems more descriptive to use format over output imo :D.

I don't think we'd want to replace -f/--file with --config... as its not meant for config and is actually scope pretty well by the <resource> parent anyway. -i/--input is probably a better option name.

cmd/tink-cli/cmd/hardware/ip.go
gianarb: not here, will do in its own pr

gianarb:
@displague about this:

https://github.com/tinkerbell/tink/pull/406#issuecomment-754076861

I am not sure I like it unfortunately and but tried passing the client in a lot of different ways delaying as long as possible but I can't find a good way yet. I am confused atm, will get back on that problem later but all the other comments are addressed

cmd/tink-cli/cmd/hardware/mac.go
gianarb: not here, will do in its own pr

client/main.go
```diff
@@ -24,6 +25,83 @@ var (
...
+}
+
+type ConnOptions struct {
+	CertURL       string
+	GRPCAuthority string
+}
+
+func (o *ConnOptions) SetFlags(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(&o.CertURL, "tinkerbell-cert-url", "http://127.0.0.1:42114/cert", "The URL where the certificate is located")
+	flagSet.StringVar(&o.GRPCAuthority, "tinkerbell-grpc-authority", "127.0.0.1:42113", "Link to tink-server grcp api")
```
mmlb, line 76:
```suggestion
	flagSet.StringVar(&o.GRPCAuthority, "tinkerbell-grpc-authority", "127.0.0.1:42113", "Link to tink-server grpc api")
```
Sigh I do this all the time

stolsma:
@gianarb I like the client api changes! It is a good basis to extend and unifi the api later on! For me this is a lgtm, but that's not on me 😉

protos/hardware/gen.go
```diff
@@ -0,0 +1,3 @@
+package hardware
+
+//go:generate moq -out mock.go . HardwareServiceClient HardwareService_AllClient
```
mmlb, line 3: boots uses https://github.com/golang/mock/gomock we should probably pick one for both projects.

Makefile
```diff
@@ -31,3 +31,7 @@ test:
 verify:
 	goimports -d .
 	golint ./...
+
+protos/gen_mock:
+	go generate ./protos/**/*
+	goimports -w ./protos/**/mock.go
```
mmlb, line 37: Would be great if these were proper deps for the necessary binaries so that they would always be up to date. Doesn't have to be now, I really want to see this PR done with :D. I can open up an issue when this PR is merged though

mmlb (dismissed):

mmlb (approved):

Issue #396, opened 2020-12-16
Events are not cleaned up when workflow is removed


invidian:
## Expected Behaviour
Events associated with a workflow should be removed when workflow itself is removed.

## Current Behaviour
`tink workflow events` returns events for workflows, which are not listed by `tink workflow list`.
## Possible Solution
Remove workflow and all associated events inside single transaction to ensure everything is cleaned up.

## Steps to Reproduce (for bugs)

1. Run sample workflow.
2. Run `tink workflow delete` to remove your sample workflow.
3. Run `tink workflow events` to list events for your workflow.

## Context

Orphan events may accumulate over time if one runs large amount of workflows, which will lead to extensive disk usage, slower seek of events etc.

One argument against this might be, that events may be used as sort of audit feature. In such case a different issue would be needed, to periodically remove events older than X days to keep number of rows reasonable.

gianarb:
That's not a concern I have, there is (or if there is not we will have one) a GC cleaning up old events no matters if the workflow is still there or not (they have a TTL almost like the Kubernetes one)

invidian:
Would it make sense to hide events for work flows which had been removed maybe? I was surprised that I removed the workflow and I was still able to find events for it.

gianarb:
> Would it make sense to hide events for workflows which had been removed maybe?

I better understand your perspective here. Now that you know how it works and that there is a TTL attached to the events do you feel confused? Because I prefer to solve this by documenting the event lifecycle more than adding more code

invidian:
>I better understand your perspective here. Now that you know how it works and that there is a TTL attached to the events do you feel confused? Because I prefer to solve this by documenting the event lifecycle more than adding more code

The TTL is time based?

Generally speaking, if this is like this _by design_, then yeah, having it documented would be great and enough for me :+1:

gianarb:
Yes the TTL is time-based and it can be changed in tink-server with an env var, you can find more about this in the actual proposal. https://github.com/tinkerbell/proposals/blob/master/proposals/0012/README.md

Documentation for events is ongoing (I am writing it, so I will be sure to include this conversation!)

gianarb:
I think this was related to event system we dropped #464 . Closing

PR #407, opened 2020-12-30
Events are cleaned up if the resource is deleted

gauravgahlot:
## Description

Going forward if a resource (template, hardware, workflow, ...) is deleted, all the events for that particular resource will also be cleaned up (irrespective of configured TTL).

## Why is this needed

Fixes: #396 

## How Has This Been Tested?

- Manually over vagrant setup
- Unit Tests

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/db/hardware.go
+++ b/db/hardware.go
@@ -27,6 +27,15 @@ func (d TinkDB) DeleteFromDB(ctx context.Context, id string) error {
 		return errors.Wrap(err, "DELETE")
 	}
 
+	_, err = tx.Exec(`
+	DELETE FROM events
+	WHERE
+		resource_id = $1;
+	`, id)
+	if err != nil {
+		return errors.Wrap(err, "DELETE EVENTS")
+	}
+
 	err = tx.Commit()
 	if err != nil {
 		return errors.Wrap(err, "COMMIT")
--- a/db/template.go
+++ b/db/template.go
@@ -99,7 +99,16 @@ func (d TinkDB) DeleteTemplate(ctx context.Context, id string) error {
 		id = $1;
 	`, id)
 	if err != nil {
-		return errors.Wrap(err, "UPDATE")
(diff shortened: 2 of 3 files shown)


gianarb:
I am still not 100% sure that this is a valuable solution. Or the solution we want. And I want to ask @invidian to help me with this.

@gauravgahlot this issue has a problem, it is deleting all the events for a particular resource when it gets deleted. It also deletes the event that notifies that a resource is deleted I presume, it means that users and applications that rely on events to synchronize an external system or to build automation will have issues dealing with a deletion. 

Excluding the event kind `DELETED` will be enough?

If a user wrote something like an "audit logger" that tracks all the actions made via Tinkerbell with an event TTL of 1h. The extension for some reason stays offline for 30 minutes and in that our hardware gets created, updated, and deleted it will be impossible to reconstruct that part of the history. IMPOSSIBLE because events are not stored anywhere anymore.

I still think events should be immutable with the only exception of the garbage collector part. How Kubernetes handle this problem for example? Do we know? 

Thanks

gauravgahlot:
AFAIK, if we delete a resource in Kubernetes we can still get its events using `kubectl get event`(including `DELETED` event). And after a TTL (default is 1hr I think) those events are gone. 

I'm glad most of the work in the PR is around tests and not the fix itself. 😄 
I was anyway going to extract the test commits to a separate PR and only leave the fix and its unit tests here.

gianarb:
well done @gauravgahlot but now that you said how Kubernetes works I will take it as a confirmation that this is a feature we should not implement 👍

invidian:
IMO Documenting the TTL for Tinkerbell events (whatever it is or whatever is planned) should be sufficient to resolve #396 :+1:

gianarb:
Great, I will close this for now, about doc, atm I don't know where to write about it! I have to figure it out

PR #411, opened 2021-01-04
while setting up provisioner, check if osie file path is valid

Cbkhare:
## Description
While creating provisioner, if `TB_OSIE_TAR` is set then osie file is extracted from the file whose location/absolute path is the value of this variable. 

## Why is this needed

if `TB_OSIE_TAR` is set and the file is not present at the given path(value of TB_OSIE_TAR) workflow fails stating the Linux image is not found. Which is valid as osie directory under `deploy/state` would be empty. 

- check has been added to verify if the file is present in the given path or not. 

## How Has This Been Tested?
By creating the provisioner.
Provisioner is not created if a tar file is present at the location defined in `TB_OSIE_TAR`. 

**on success** 
Screenshot from 2021-01-04 18-50-51

**On failure** 
Screenshot from 2021-01-04 18-48-44

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -260,6 +260,12 @@ setup_osie() (
 			curl -fsSL 'https://tinkerbell-oss.s3.amazonaws.com/osie-uploads/latest.tar.gz' -o ./osie.tar.gz
 			tar -zxf osie.tar.gz
 		else
+			if [ ! -f "$TB_OSIE_TAR" ]; then
+				echo "$ERR osie tar not found in the given location $TB_OSIE_TAR"
+				exit 1
+			else
+				echo "$INFO extracting osie tar"
+			fi
 			tar -zxf "$TB_OSIE_TAR"
 		fi
 
--- a/setup.sh
+++ b/setup.sh
@@ -263,9 +263,8 @@ setup_osie() (
 			if [ ! -f "$TB_OSIE_TAR" ]; then
 				echo "$ERR osie tar not found in the given location $TB_OSIE_TAR"
 				exit 1
-			else
-				echo "$INFO extracting osie tar"
 			fi
+			echo "$INFO extracting osie tar"
 			tar -zxf "$TB_OSIE_TAR"
 		fi
 


mmlb:
Shouldn't this already be caught by `tar` exiting with an error and `set-e` being set?

Cbkhare:
> Shouldn't this already be caught by `tar` exiting with an error and `set-e` being set?

@mmlb , right, I think the script is not bailing out in that case. With this PR it goes for a graceful exit.

setup.sh
```diff
@@ -260,6 +260,12 @@ setup_osie() (
 			curl -fsSL 'https://tinkerbell-oss.s3.amazonaws.com/osie-uploads/latest.tar.gz' -o ./osie.tar.gz
 			tar -zxf osie.tar.gz
 		else
+			if [ ! -f "$TB_OSIE_TAR" ]; then
+				echo "$ERR osie tar not found in the given location $TB_OSIE_TAR"
+				exit 1
+			else
+				echo "$INFO extracting osie tar"
```
tstromberg, line 267: Please remove the "else" and move this echo statement to appear just before the `tar`. This will make the code easier to follow.

tstromberg (changes requested):

tstromberg:
This PR looks great! The DCO checker noticed that one of your git commits isn't signed properly - can you fix that so that this PR may be merged?

https://github.com/tinkerbell/tink/pull/411/checks?check_run_id=2999422937 has instructions on how to fix this PR.

gauravgahlot:
@tstromberg I tried those steps already but didn't work. Will it be okay if I close this PR and create a new one. I know the original contributor personally and I don't think he is looking into Tinkerbell anymore.

nshalman:
> @tstromberg I tried those steps already but didn't work. Will it be okay if I close this PR and create a new one?
> I know the original contributor personally and I don't think he is looking into Tinkerbell anymore.

@gauravgahlot That would be great. I'll close out this one.

Issue #415, opened 2021-01-12
Evaluating template in workflow replaced << with &lt;<


invidian:
## Expected Behaviour
Workflow properly renders template without replacing any characters.

## Current Behaviour

When template includes sequence like `<<`, it gets replaced into `&lt;<.

## Possible Solution

Use `text/template` instead of `html/template` when rendering workflows.

## Steps to Reproduce (for bugs)

Unit test to reproduce:
```go
func TestRenderTemplateNotHTML(t *testing.T) {
        validTemplate := `
version: "0.1"
name: hello_world_workflow
global_timeout: 600
tasks:
  - name: "hello world<<"
    worker: "{{.device_1}}"
    actions:
    - name: "hello_world"
      image: hello-world
      timeout: 60
`
        templateID := uuid.New().String()
        hwAddress := []byte("{\"device_1\":\"08:00:27:00:00:01\"}")

        result, err := RenderTemplate(templateID, validTemplate, hwAddress)
        if err != nil {
                t.Fatalf("rendering should not fail")
        }

        expected := "hello world<<"

        if !strings.Contains(result, expected) {
                t.Fatalf("Result should contain %q, got \n\n%s", expected, result)
        }
}
```

## Context
https://twitter.com/_invidian/status/1339504259340091392

Found while working on https://github.com/tinkerbell/cluster-api-provider-tink/pull/17.

## Your Environment

Tinkerbell sandbox from version https://github.com/tinkerbell/sandbox/commit/be40a7b371fdab947a413e190cfe99b692651d8e.

gianarb:
I agree with John Arundel https://twitter.com/bitfield/status/1339506837230632960 is not a bug it is a security-conscious decision. 

I am joking thank you for opening this issue

invidian:
Added sample unit test to reproduce. Perhaps one in the code could be more comprehensive.

PR #416, opened 2021-01-12
RenderTemplate using the text/template package, not the html one

gianarb:
## Description

Apparently, we imported the wrong package, we need to use text no HTML rules.

## Why is this needed

Wrong import, we do not need to escape following HTML rules.
Fixes: #415 

## How Has This Been Tested?
 Unit test

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/workflow/template_validator_test.go
+++ b/workflow/template_validator_test.go
@@ -237,6 +237,7 @@ func TestRenderTemplate(t *testing.T) {
 		hwAddress        []byte
 		templateID       string
 		templateData     string
+		skip             string
 		expectedError    func(t *testing.T, err error)
 		expectedTemplate string
 	}{
@@ -261,6 +262,7 @@ tasks:
 		{
 			name:         "invalid-hardware-address",
 			templateData: validTemplate,
+			skip:         "Not sure if this is an expected error or not but this test is not well written",
 			hwAddress:    []byte("{\"invalid_device\":\"08:00:27:00:00:01\"}"),
 			expectedError: func(t *testing.T, err error) {
 				if err == nil {
@@ -304,9 +306,13 @@ tasks:
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
+			if test.skip != "" {
+				t.Skip(test.skip)
+			}
 			temp, err := RenderTemplate(test.templateID, test.templateData, test.hwAddress)
-			if err != nil {
+			if test.expectedError != nil {
 				test.expectedError(t, err)
+				return
 			}
 			if diff := cmp.Diff(test.expectedTemplate, temp); diff != "" {
 				t.Error(diff)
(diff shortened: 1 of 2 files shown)


invidian (approved):
I guess it's sufficient to fix #415 :+1:

workflow/template_validator_test.go
```diff
@@ -233,17 +233,37 @@ func TestValidateTemplate(t *testing.T) {
...
+      image: hello-world
+      timeout: 60
+`,
 		},
 		{
-			name:      "invalid-hardware-address",
-			hwAddress: []byte("{\"invalid_device\":\"08:00:27:00:00:01\"}"),
+			name:         "invalid-hardware-address",
+			templateData: validTemplate,
+			skip:         "Not sure if this is an expected error or not but this test is not well written",
```
gauravgahlot, line 265: Why do we want to skip this test?

gianarb: the error does not happen with the current code

gauravgahlot: That's odd because I do get an error:

```
/ # cat tmp/hw.json | tink hardware push
2021/01/13 08:16:06 Hardware data pushed successfully

/ # cat tmp/hello.tmpl | tink template create -n hello
Created Template:  a6aff6d5-5577-11eb-aef3-0242ac120004
 
/ # tink workflow create -t a6aff6d5-5577-11eb-aef3-0242ac120004 -r '{"invalid_device": "08:00:27:00:00:01"}'
2021/01/13 08:24:04 rpc error: code = Unknown desc = failed to render template, invalid hardware address: {"invalid_device": "08:00:27:00:00:01"}

/ # cat tmp/hello.tmpl 
version: "0.1"
name: hello_world_template
global_timeout: 6
tasks:
  - name: "hello world"
    worker: "{{.device_1}}"
    actions:
      - name: "hello_world"
        image: hello-world
        timeout: 60
/ # 

```

gianarb: The error is not caused by a particular validation on the field, but it is caused by the `html/template` package. That is wrong

gianarb: Fixed! Thanks

invidian (approved):

workflow/template_validator_test.go
```diff
@@ -233,34 +233,88 @@ func TestValidateTemplate(t *testing.T) {
...
+`,
+		},
 	}
 
-	templateID := uuid.New().String()
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			_, err := RenderTemplate(templateID, validTemplate, test.hwAddress)
-			if err != nil {
+			if test.skip != "" {
```
gauravgahlot, line 308: @gianarb We don't need this anymore, I presume.

gauravgahlot (changes requested):

workflow/template_validator_test.go
gianarb: Removed, if we will need to temporarily skip tests, we will have to add it back 👍

gauravgahlot (approved):
Thanks.

invidian (approved):

Issue #417, opened 2021-01-20
Update new contributor experience


jacobweinstock:
Getting started with the Tinkerbell stack and through this Tink repo the local development path, for me, had a few hurdles. 

- I found it difficult to understand and spin up Tink without diving deep into the code.
- I haven't been able to generate TLS certificates locally using the scripts found in this repo. They were all very coupled to vagrant with virtualbox.
- There were a lot of environment variables that were difficult to understand what to set them to without, again, diving deep into the code.
- `make run` did not work for me.
```bash
ERROR: 
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
```
- Documentation was difficult to find

## Expected Behaviour
lower barrier to use, contribute and understand.

## Current Behaviour
see above

## Possible Solution
I have a PR here: https://github.com/tinkerbell/tink/pull/418 with proposed changes to add an environment variable to disable TLS in the server and cli. Documentation for all environment variables. New top-level docker compose file. And updated `make run` target. See PR 

## Your Environment

* Operating System and version:
MacOS Catalina

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:
Running Tink locally with docker-compose.

* Link to your project or a code example to reproduce issue:

tstromberg:
@jacobweinstock - IMHO, this issue would be a lot more actionable if split into multiple issues. Would you mind doing so, so that they can be attacked one at a time?

jacobweinstock:
I feel that all of these except one have actually been resolved.

_I found it difficult to understand and spin up Tink without diving deep into the code._

- `make run` now easily stands up the Tink stack.

_I haven't been able to generate TLS certificates locally using the scripts found in this repo. They were all very coupled to vagrant with virtualbox._

- `make run` and the docker-compose.yaml now handle TLS certs more clearly.

_There were a lot of environment variables that were difficult to understand what to set them to without, again, diving deep into the code._

- This is the one I feel is still unresolved. I have a closed PR (#418) that documents the environment variables. I will reopen it.

_`make run` did not work for me._

- This is working well for me now.

nshalman:
#545 is a very good start. If anyone wants to pick up from there and complete the work it would be much appreciated.

PR #418, opened 2021-01-20
update contributor experience

jacobweinstock:
## Description

Getting started with the Tinkerbell stack and through this Tink repo the local development path, for me, had a few hurdles. 

- I found it difficult to understand and spin up Tink without diving deep into the code.
- I haven't been able to generate TLS certificates locally using the scripts found in this repo. They were all very coupled to vagrant with virtualbox.
- There were a lot of environment variables that were difficult to understand what to set them to without, again, diving deep into the code.
- `make run` did not work for me.
```bash
ERROR: 
        Can't find a suitable configuration file in this directory or any
        parent. Are you in the right directory?

        Supported filenames: docker-compose.yml, docker-compose.yaml
```
- Documentation was difficult to find

This PR adds `make run` and `make run-clean` to allow spin up and down a docker-compose tink, tink-cli, db stack. PR updates the server and cli dockerfiles and adds a top level docker-compose.yaml. 

## Why is this needed

Fixes: https://github.com/tinkerbell/tink/issues/417

## Checklist:

I have:

- [x] added unit or e2e tests


gianarb:
Great initiative. The development experience in Tinkerbell is not something that I am personally really involved with but I know @mmlb has opinions as well.

Overall the `directory` folder as we know it can be away or change drastically. I would like to avoid more files or ways to developer I think we should just have one that works for a good amount of people. 

Now sure which one, overall what I don't like is the fact that it adds another docker-compose file without removing anything

mmlb:
Hey @jacobweinstock I agree with @gianarb re multiple docker-compose files. Can you try to merge them? I think yours is the one to keep. For example, I think we can drop the tls cert stuff and registry container. I'm also not sure we need the connections to tink-servr via 127.0.0.1 and can just use docker-compose's dns.

jacobweinstock:
@gianarb @mmlb , just to clarify are we talking about the `deploy` directory and `deploy/docker-compose.yml`? 

Also, would the prefered location for the new updated `docker-compose.yml` file in this PR be top-level or in the `deploy` directory or somewhere else?

mmlb:
@jacobweinstock yep getting rid of the deploy/docker-compose.yml is what I'd like to see. deploy was started from before we had the sandbox so I think having the files here be tailored for development is what makes the most sense.

jacobweinstock:
I will close this PR. I'll create new PRs. one to be just docs and create another PR for code changes.

Issue #414, opened 2021-01-12
Create a testsuite for migration


gianarb:
https://github.com/tinkerbell/tink/pull/383#issuecomment-748398976

![Screenshot 2021-01-12 at 13 50 36](https://user-images.githubusercontent.com/1630267/104316696-275f6180-54dd-11eb-9dc1-7a12f92b2646.png)

Those are two reports caused by a not that defined workflow around migration. We have to improve safeguards and migration validation, even via testcontainers if needed and with static checks.

- [x] Write an integration test that runs all the migration and validate that they work ( this is in sandbox)
- [x] Write a unit test that validates that new migration have the right ID and the right function name
- [x] maybe we can use `go generate` for something here (I don't think it applies here)

PR #421, opened 2021-01-23
Simple tests for db/migration

mmlb:
## Description

Some quick/shallow test for migration definition.

## Why is this needed

https://github.com/tinkerbell/tink/pull/371 introduced a migration where none of the names matched :D, the Id & FuncName was fixed in https://github.com/tinkerbell/tink/pull/383 but that one missed the file name. Why make humans do things they're bad at when instead we can make a computer do something its good at? So this is that :D.

## How Has This Been Tested?

Ran `go test` and it found the bad file name. I artificially changed an Id and the test caught the FuncName|Id mismatch.

N/A

## Checklist:

I have:

- [x] added unit or e2e tests


mmlb:
According to codecov our migrations have 100% code coverage so that means no bugs right? :D

gianarb:
Thanks, I wrote this issue back then when the various bugs you referenced here showed up #414.
Really appreciate seeing this PR coming!

gianarb (approved):

mmlb:
Yep I don't think this'll get us a fix for #414 but sets us up to work on it :D

PR #422, opened 2021-01-26
Remove deploy directory

gianarb:
As discussed on Slack and briefly in #418 we are not sure we need the
deploy directory now that we have sandbox.

The initial idea about leaving it in the tink repository was to provide
an environment that developers can use to contribute to Tinkerbell.

I am not sure if it evolved in that way and it is kind of making
maintenance hard because it diverges from what we have in sandbox.

Overall the proposal is to remove it from tink.


fransvanberckel:
How did you solve the use of deploy/db/tinkerbell-init.sql? The db directory is not in sandbox.

jacobweinstock (dismissed):

gianarb:
Hey @fransvanberckel this is one of the reasons why I think the way `deploy` works today is not effective. It is a leftover, it does not work. You have to remove that line because tink-server does not need that anymore, migration is doing the same thing.

mrchrd:
I looked at it last weekend and I postponed the project. I realized that I needed to understand better the testing strategy. The line between testing the microservice vs tinkerbell as a whole looks blurry so I thought it needed to be clarified first.

I guess that `setup.sh`, `generate-envrc.sh`, `test/_vagrant` and other files should be removed to, but I don't know what would be the impact.

gianarb:
All those files are in sandbox as well (slightly differently probably)@mrchrd

gianarb:
No impact @mrchrd you are right! they should be removed! let me update my PR!

jacobweinstock (approved):

gianarb:
I wrote a guide about how I develop tink locally (only CLI and Server), I hope it will work as a replacement at least when you don't need all the stack up and running (I usually don't need it)

https://github.com/tinkerbell/tinkerbell-docs/pull/71

fransvanberckel:
Please verify the leftover files, .dockerignore, .envrc, .gitignore, Makefile and test-docker-compose.yml.

PR #423, opened 2021-01-26
Remove gRPC TLS cert

gianarb:
The way we do TLS in Tinkerbell is just a complication.
It does not increase security because everybody that knows how to
download the certificate from the tink-server can connect.

It can't be used to identify the requester and it means that
is not a carrier that we can use for authorisation moving forward.

With a couple of other maintainers we would like to remove TLS
completely for now until we can figure out a better design for it.

### DIFF
--- a/client/main.go
+++ b/client/main.go
@@ -1,10 +1,7 @@
 package client
 
 import (
-	"crypto/x509"
-	"io/ioutil"
 	"log"
-	"net/http"
 	"os"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


mmlb (approved):

fransvanberckel:
What about removing the generate_certificates function in setup.sh from the sandbox repo?

gianarb:
Hey @fransvanberckel will do as a follow-up. I want to get the vibe of this #422 first :)

jacobweinstock (approved):

Issue #359, opened 2020-11-07
Metadata (protos/packet) should be typed in Hardware


displague:
## Expected Behaviour

Metadata is typed within the Hardware type used by the Hardware client.

```go
	hwReq := &hardware.Hardware{
		Network: &hardware.Hardware_Network{},
		Id:       "uuid",
		Version:  0,
		Metadata: &packet.Metadata{
		        Instance: &packet.Metadata_Instance{},
	        }
	}
	pushReq := &hardware.PushRequest{data: hwReq}
```

## Current Behaviour

Metadata is a string within the Hardware type used by the Hardware client.

```golang
        metadata := &packet.Metadata{
		Instance: &packet.Metadata_Instance{
			
		}
	}
	hwReq := &hardware.Hardware{
		Network: &hardware.Hardware_Network{},
		Id:        "uuid",
		Version:  0,
		Metadata: metadata.String(),
	}
	pushReq := &hardware.PushRequest{data: hwReq}
```

As a Tinkerbell client implementor, creating a Hardware object, it is unexpected that the client interface should require a string interpretation of the `Metadata` type from the `protos/packet` package. 

## Possible Solution

* Rename protos/packet to protos/metadata.
* Define the Hardware type's Metadata as a Metadata type (not a string) using protobuf imports (https://developers.google.com/protocol-buffers/docs/proto3) and optional/repeated elements (https://developers.google.com/protocol-buffers/docs/encoding#optional)

## Context

I'm investigating a Crossplane client for Tinkerbell (https://github.com/displague/crossplane-provider-tinkerbell).

This would also benefit the Tinkerbell Terraform provider, which currently passes json and yaml blobs around.

## Your Environment

Go importing tink at 384fe30ce2a4

displague:
Is `metadata` intentionally a string to stay unbiased about the metadata format?
Shouldn't Tinkerbell have a bias on the metadata format (a bias to follow hegel)?

This would be helpful for adding `cloud-init` support for the Tinkerbell metadata service.

gianarb:
@mmlb is it part of the proposal you raised around instance? Can you help us to figure how if @displague has a reasonable request or if it is like it is for good reasons?

jacobweinstock:
`packetMetadata.String()` does not exist anymore. @displague if you feel this is still needed, please reopen.

PR #428, opened 2021-02-02
Remove packet protobuf

gianarb:
This should not be required and I can't see we use it


gianarb:
Hegel uses it

mmlb:
You know I think we should probably promote it as the metadata value and then later replace it with Instance proposal. I don't think we want all the things here in Instance. But actually, the problem with promoting this to hardware.metadata is that it then becomes hard for other people to extend how ever they want ...

gianarb:
What do you mean expanded as they want? Metadata are the contract used from
tink worker and boots it can't be whatever we want.

I understand it is to be flexible but a well designed structure can allow
that I suppose

gianarb:
Ok I suppose we are not ready for this, if you want to discuss about this topic I think the right issue is https://github.com/tinkerbell/tink/issues/359

PR #429, opened 2021-02-03
Fix permissions on downloaded artifacts:

jacobweinstock:
## Description

The Github Action `actions/upload-artifact@v2` does not preserve file permissions on upload. https://github.com/actions/upload-artifact/tree/v2#permission-loss
This means almost certainly all docker images, before this commit, are not working.
I have validated that with the latest tags, as shown here:

image

It's difficult to test this change. I think this will work but will need it to run in Github actions to actually know.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -135,6 +135,8 @@ jobs:
       with:
           name: ${{ matrix.binary}}
           path: cmd/${{ matrix.binary }}
+    - name: Fix Permissions
+      run: chmod +x cmd/${{ matrix.binary }}*
     - name: ${{ matrix.repository }}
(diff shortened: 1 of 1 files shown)


markyjackson-taulia (dismissed):

markyjackson-taulia (approved):

gianarb:
We use the images in sandbox and theu works :)

gianarb:
If I remember well you use this images internally as well

gianarb:
Not sure when this happened! But we need better CI for sure! I am scared!!
Thanks

mmlb:
Hmm agree we should probably do simple `--help` of the container image probably.

PR #436, opened 2021-02-11
Pid namespace support

thebsdbox:
## Description

This PR updates the action protos with a pid option of type string, following the `--pid` flag for the docker engine. The most common use-case is to pass the string `host` to the pid option which will allow the action to run in the same process ID namespace as the host itself. 

## Why is this needed

This adds the capability to both cleanly `kexec` and `reboot` as an action. 

Fixes: 👉  (an already merged `kexec` action on the tinkerbell action hub)

## How Has This Been Tested?
This has been tested by running a workflow with the `kexec` action

If this isn't used then it defaults to pid: container (or the runtime default)

## Checklist:

I have:

- [ ] added unit or e2e tests


markyjackson-taulia (approved):
/lgtm
Also tested this yesterday

gianarb (approved):

Issue #413, opened 2021-01-11
Can't create workflow with template bigger than 8000 characters


invidian:
It seems workflow/template combination cannot be bigger than 8000 characters due to use of `pg_notify`.

## Expected Behaviour

Workflow with template bigger than 8000 characters can be created.

## Current Behaviour

Tink returns the following error:
```
rpc error: code = Unknown desc = INSERT: pq: payload string too long
```

## Possible Solution

Probably don't use `pg_notify` (as partially suggested in https://github.com/tinkerbell/tink/issues/379#issuecomment-742488865).

## Steps to Reproduce (for bugs)

1.
2.
3.
4.

## Context

While implementing https://github.com/tinkerbell/cluster-api-provider-tink#17 with updates sandbox to include events code, I noticed, that I can no longer create workflows.

See also https://github.com/meteor/postgres-packages/issues/56

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

invidian:
Hmm, it seems once hit, I'm not able to remove old templates either :/

gianarb:
I still regret the decision of using triggers

invidian:
>@invidian every action creates an event, delete does not make an exception here

What I experienced it is I tried creating big workflow, then it blocked removing other templates for some reason.

mmlb:
:( I was not aware of that limitation of pg_notify, it seems we'll need to drop that then.

gianarb:
I would like to drop a 💣 . This looks like the right time.

I asked @gauravgahlot to help me with this issue. I realized that triggers are not that bad, but we have a problem with the overall concept of the template I think. Or at least, the way we implemented them is too simple. Let me explain, later you will understand the connection with this issue.

First, Kubernetes relays on ETCD watchers that are in some way similar to what we do with Postgres triggers. It is a database feature in both cases. Removing triggers is a bit too complicated because it requires coordination in a HA environment as @mmlb tried to say to me multi times, but I was too busy thinking about my own idea (now I get it sorry ✌️ ).

Template looks like this:

```proto
message WorkflowTemplate {
  string id = 1;
  string name = 2;
  reserved 3; // obsolete data
  google.protobuf.Timestamp created_at = 4;
  google.protobuf.Timestamp updated_at = 5;
[...]

PR #438, opened 2021-02-17
Template revision

gauravgahlot:
## Description

The PR introduces template revisions. The `template.Data` field will not contain the actual template but a reference to a particular revision. Revisions are immutable, when you update a template you create a new revision.

## Why is this needed

Fixes: #413 

## How Has This Been Tested?

- Unit Tests
- Manual Testing 

## Checklist:

I have:

- [x] added unit or e2e tests


db/db.go
```diff
@@ -42,6 +42,7 @@ type template interface {
 	DeleteTemplate(ctx context.Context, name string) error
 	ListTemplates(in string, fn func(id, n string, in, del *timestamp.Timestamp) error) error
 	UpdateTemplate(ctx context.Context, name string, data string, id uuid.UUID) error
+	ListTemplateRevisions(id string, fn func(id string, revision int, tCr *timestamp.Timestamp) error) error
```
gianarb, line 45: Let's try to write something a bit more readable:

```go
ListRevisionsByTemplateID(templateID string, fn func(revisionID string, revision int, tCr *timestamp.Timestamp) error) error
```

I am not sure what `revisiononID string` and `revision int` are

gianarb (changes requested):

db/db.go
```diff
@@ -81,6 +82,19 @@ func (t *TinkDB) Migrate() (int, error) {
 	return migrate.Exec(t.instance, "postgres", migration.GetMigrations(), migrate.Up)
 }
 
+// WithMigrations applies only the given migrations.
+// It is required to unit test migrations.
+func (t *TinkDB) WithMigrations(migrations []func() *migrate.Migration) error {
```
gianarb, line 87: I don't think we need this function for mainly one reason. This is a new API that should not be used outside of `tests`. So it is in the wrong place, it should be exposed via `TinkDB` because it allows developers to use it outside of tests, and we do not want that.

Another reason is that it does not add much value or simplification you use it this way:

```go
	err := tinkDB.WithMigrations([]func() *migrate.Migration{
		migration.Get202009171251,
		migration.Get202010071530,
		migration.Get202010221010,
		migration.Get202012041103,
		migration.Get202012091055,
		migration.Get2020121691335,
	})
```

But you can get to the same point writing:

```go
	_, err := migrate.Exec(db, "postgres", migrate.MemoryMigrationSource{
		Migrations: []*migrate.Migration{
			migration.Get202009171251(),
			migration.Get202010071530(),
			migration.Get202010221010(),
			migration.Get202012041103(),
			migration.Get202012091055(),
			migration.Get2020121691335(),
		},
	}, migrate.Up)
```

Almost same line of code. If you are wondering where `db` is coming from:

```go
db, tinkDB, cl := NewPostgresDatabaseClient(t, ctx, NewPostgresDatabaseRequest{})
```

db/template_test.go
```diff
@@ -364,6 +431,137 @@ func TestGetTemplateWithInvalidID(t *testing.T) {
...
+		t.Error(err)
+	}
+
+	// create a few templates
+	w := workflow.MustParseFromFile("./testdata/template_happy_path_1.yaml")
+	for i := 1; i <= 5; i++ {
+		id := uuid.New().String()
+		w.ID = id
+		w.Name = fmt.Sprintf("id_%d", rand.Int())
+		err := createTemplateFromWorkflowType(ctx, tinkDB, w)
```
gianarb, line 523: This test fails because the function `CreateTemplate` already relies on revision. Even if you didn't apply the migration yet, the Go code assumes revision to be in place.

In order to fix that you have to add a template not via code but with a SQL query.

gianarb (changes requested):

db/template_test.go
```diff
@@ -364,6 +433,173 @@ func TestGetTemplateWithInvalidID(t *testing.T) {
...
+	})
+	if err != nil {
+		t.Error(err)
+	}
+	if revisionCount != 5 {
+		t.Errorf("expected %d template revisions stored in the database but we got %d", 5, revisionCount)
+	}
+}
+
+func createTemplatesFromSQL(ctx context.Context, db *sql.DB) error {
```
gianarb, line 574: Can you add a comment? I am sure the next person or even myself in a couple of days will not remember why we have this function and the name is so generic that it won't help

gianarb:
@gauravgahlot got from Dan and EM two databases containing different templates. We will use them as data sources for testing the migration against more data ❤️

gauravgahlot:
> @gauravgahlot got from Dan and EM two databases containing different templates. We will use them as data sources for testing the migration against more data ❤️

Tested with the databases and it's working as expected. 🚀

gianarb:
When a workflow starts we have to keep track of the revision as well. To be able to figure out what it actually ran even after a couple of days, when it is likely possible that the template changed to a different revision.

In preparation of exposing the actual revision do you mind creating two GRPC requests as part of the TemplateService:

* ListRevisionsByTemplateID: This request takes a template ID and returns all the revision we have
* GetRevision: this request takes templateID and a revisionID and it returns a single revision

Please write tests for this as well and document the new API as part of the proto file

gauravgahlot:
> Is this PR ready to be re-reviewed? I noticed there were still some unresolved comments.

The PR aimed to solve the issue with tink events. Now that we don't have events, we don't really need this PR (unless we are planning to reintroduce events). The issue itself can be closed as well. However, it would be great if @thebsdbox and @gianarb can confirm.

Raj-Dharwadkar:
@gianarb Please can you confirm ?

gianarb:
I can confirm. Let's close it for now. Thanks a lot for your work I am sure if will be reused in the future

Issue #435, opened 2021-02-09
Create cli command


gianarb:
Currently, the CLI does not have a common way to create a new resource. Hardware gets `pushed`, template and workflows get created for example.

Even between templates and workflows, the features are different. You can create a template via a file with `--path` but not a workflow.

It is time to have a consistently create command as we have for delete #425 and get.

---

## Input channel

The create command should accept a `--path`:

```
tink hardware create --path ./hardware.json
```

`--path` needs to be a file. If it is a directory for now you can fail to say that directory is not supported for now.

Or it can get input from stdin:

```
cat ./template.yaml | tink template create - 
```

## Input format

Things can get messy (because they are messy already). Hardware is the main issue I presume and we know why #359. Personally and when looking at examples I visualize in my mind hardware like JSON. Template and workflows as YAML.

My current suggestion is to give up, for now, let's derive the input format from the `--path`, if the file is `.json` let's unmarshal with JSON. If it is `.yaml` let's use YAML. 

In case of input via stdin let's do cascade, let's start with JSON, fallback to YAML. if nothing works it is a not valid input format.

We only support YAML and JSON.

## Output

The output is the same as the `get` command but with only the created resources (reuse the same code please).
It means that create should support:

```
--format
```

As a way to select the output format (and this makes clear why kubectl uses `--output` and not `--format`, because it makes clear that `--output` manipulate the output format... `--format` here opens the question: "is it for input or for output?" @mmlb )

mmlb:
I'd rather not see 2 different forms of specifying paths, normal `-` is a unix short hand for stdin where ever a filepath is expected. For example:

```
man cat
...
EXAMPLES
       cat f - g
              Output f's contents, then standard input, then g's contents.
```
also see `man curl` for `"-"`, there are many others :D.


So I'd rather see `tink template create path, if path is `-` read from stdin. If path is not given read from stdin (if stdin is a tty print a "reading from stdin" message to stderr). Otherwise if path is given then read that.

mmlb:
> Even between templates and workflows, the features are different. You can create a template via a file with `--path` but not a workflow.

This is currently working as designed right? Workflows are only ever meant to be created from a preexisting template so it doesn't make sense create a workflow based off of a file.

> ## Input format
> 
> Things can get messy (because they are messy already). Hardware is the main issue I presume and we know why #359. Personally and when looking at examples I visualize in my mind hardware like JSON. Template and workflows as YAML.
> 
[...]

PR #449, opened 2021-03-04
Fix for #435, Create cli for tink

parauliya:

## Description

Reusable tink create cli

## Why is this needed
Fixes: #435 

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/tink-cli/cmd/create/create.go
+++ b/cmd/tink-cli/cmd/create/create.go
@@ -1,0 +1,123 @@
+package create
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/tinkerbell/tink/client"
+	"google.golang.org/grpc"
+)
+
+type ResourceType string
+
+var (
+	Resources = map[ResourceType]int32{
+		"HARDWARE": 0,
+		"TEMPLATE": 1,
+		"WORKFLOW": 2,
+	}
+)
+
+// Options struct
+type Options struct {
+
+	// Based on this the Create function will be called
+	Resource ResourceType
+	// CreateByStdin is used for creation of Hardware and Template resources
+	CreateByStdin func(context.Context, *client.FullClient, []byte) (interface{}, error)
+
+	// CreateByFlags is used for Workflow resource creation
+	CreateByFlags func(context.Context, *client.FullClient) (interface{}, error)
+
+	clientConnOpt *client.ConnOptions
+	fullClient    *client.FullClient
+}
+
+// SetClientConnOpt set client options
+func (o *Options) SetClientConnOpt(co *client.ConnOptions) {
+	o.clientConnOpt = co
+}
+
+// SetFullClient set complete clim
+func (o *Options) SetFullClient(cl *client.FullClient) {
+	o.fullClient = cl
+}
+
+const shortDescr = "Create one or more resources"
+
+const longDescr = `Creates one or more resources and prints the status of
+the deleted resource.
+
+# Create template resource (success)
+tink template create < <path to template file>
+Created Template: 8ae1cc24-6a9c-11eb-a0fc-0242ac120005
+
+# Create template resource (not found)
+tink template create < <path to template file>
+Error	file not found or 
+`
+
+const exampleDescr = `# Create template resource
+tink template create < <path to template file>
+
+# Create hardware resource
+tink hardware create < <path to hardware data file>
+
+# Create workflow resource
+tink workflow create --template-id <template id> --hardware '{"device_1":"00:99:88:77:66"}'
+
+`
+
+// NewCreateCommand creates a new `tink create` command
+func NewCreateCommand(opt Options) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:                   "create",
+		Short:                 shortDescr,
+		Long:                  longDescr,
+		Example:               exampleDescr,
+		DisableFlagsInUseLine: true,
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			if opt.fullClient != nil {
+				return nil
+			}
+			return nil
+		},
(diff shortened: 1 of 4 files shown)


gianarb:
Hey @parauliya as we spoke yesterday I have implemented the common feature and the structure of the create command.

You should not deference NEVER any resource inside the `create` pkg. This is the only way to stay resource agnostic. When you have doubt look at `delete`, `get`.

have a look at the `doc.go` file for the `create` package to have an idea about where I am going with it.

I realized too late that I introduced the `--file` that is something we probably don't want. I am not sure and I don't know how cobra supports what @mmlb suggests like `- <path_file>`. @mmlb do you know? Also, I think it opens the door for more questions:

* can you create multiple resources in one command if you pass multiple files? `- <path_file1> `- <path_file2>` or should we fail? 
* If the multiple resource creation is something we want I think we have to reevaluate the strategy to reconcile `flags` with the content passed via `stdin`, or `file` because if it is not a single resource is harder to reconcile (right now flags override what gets passed via `stdin`, `file`.

mmlb:
> I realized too late that I introduced the `--file` that is something we probably don't want. I am not sure and I don't know how cobra supports what @mmlb suggests like `- <path_file>`. @mmlb do you know? 

No I don't know, but I would _assume_ cobra will have 2 left over args after processing: `-` and `<path_file>`.

Also, I think it opens the door for more questions:
> 
>     * can you create multiple resources in one command if you pass multiple files? `- <path_file1> - <path_file2>` or should we fail?

Would be nice to process them all, imo.

> 
>     * If the multiple resource creation is something we want I think we have to reevaluate the strategy to reconcile `flags` with the content passed via `stdin`, or `file` because if it is not a single resource is harder to reconcile (right now flags override what gets passed via `stdin`, `file`.

Have an example of this?

gianarb:
> Have an example of this?

I want to get to a point where you can even pass a file or a stdin or if available for a resource you can use flags. Workflows are the ones I have in my mind right now.

You can use flags:

```
$ tink workflow create \
  --template-id CA321BBF-2AE0-4B45-AAE3-076C894FA282 \
  --data device_1=AF:E8:60:85:A6:CF \
  --data address=192.168.1.12
```

You can get it from stdin:

```
cat ./workflows.yaml | tink workflow create
```

Probably workflow.yaml will look like:

```yaml
templateID: CA321BBF-2AE0-4B45-AAE3-076C894FA282
data:
  - device_1: AF:E8:60:85:A6:CF
  - address: 192.168.1.12
```

In theory, if we assume you can create only one resource at a time we can merge flags over a stdin/file. But if we take more files, I think we should error because we won't know how to merge it:

```
cat ./workflows.yaml | tink workflow create --data device_1=5D:8B:43:28:66:21
```

If we go with:

```
 tink workflow create -- <file1>
```

The UX enables the user to think that they can add more paths (just as you do with git for example), so if we decide that we don't want to support multiple resources (because we don't know how to merge flags, or to avoid explaining it to end-users), I think we should avoid a UX that is not self-explanatory.   

Those are the options I see:

* Declare that only one resource can be created at the time
* Error if stdin/file AND flags are passed at the same time (this will allow us to accept stdin or files that have multiple resources)
* Accept everything and explain all the rules we onboarded, but this is error-prone and not that user friendly

mmlb:
Ah yeah `workflow create` is a weird beast here. It would be cool if viper did something like `--template id1 --device foo --device bar --template id2 --device baz`, but I think it probably makes more sense to rethink differently here. I think either we take in one/many files like you show as path or stdin and/or allow flags to define only 1 workflow and everything comes in from flags. I would not really miss the flags going away.

gianarb:
> I would not really miss the flags going away.

This is a hard sentence. Currently, today workflows for example only get created via flags... So in theory you are saying you won't miss the requirement of writing your workflows to a file in order to create one. This makes scripting a bit more complicated.

I think flags play a role in the update command, not really related to create but I expect the update command to supports the same flags a create has to do something like a PATCH. If we remove flags all together it means we will update only via file (extract the current version, modify and "apply")

mmlb:
@gianarb 

> > I would not really miss the flags going away.
> 
> This is a hard sentence. Currently, today workflows for example only get created via flags... So in theory you are saying you won't miss the requirement of writing your workflows to a file in order to create one. This makes scripting a bit more complicated.

Not really:
```sh
for mac in mac1 mac2 mac3; do
  tink workflow create <<-EOF
    templateID: CA321BBF-2AE0-4B45-AAE3-076C894FA282
    data:
      - device_1: $mac
  EOF
```
or any other way to build up stdin which isn't too bad and why I say I wouldn't miss flags going away.

If you do want to keep them then it would be ok to do just one workflow at a time and only via flags.

I don't really feel strongly either way. I'd be fine with making `workflow create` look like all the other commands in that they don't take flags.

> I think flags play a role in the update command, not really related to create but I expect the update command to supports the same flags a create has to do something like a PATCH. If we remove flags all together it means we will update only via file (extract the current version, modify and "apply")
> 
> NB: Weird to speak about update for workflows that technically should be immutable (I think)

Yeah, I'm still like -0 on updating a template/workflow. I'd rather it just be a new template instead. Leaving "update" logic out to a different service/program/user. If someone _really_ wants it then fine but I don't really see the use case.

mmlb:
> Yeah, I'm still like -0 on updating a template/workflow. I'd rather it just be a new template instead. Leaving "update" logic out to a different service/program/user. If someone _really_ wants it then fine but I don't really see the use case.

Ooops I was thinking of updating a template not a workflow, which is what I'm -0 for. I'm very much -2/NAK for updating a workflow. That should be immutable. You didn't like the workflow your're running then cancel it and start a new one.

splaspood:
Looks like there have been some good points made by a number of people back in March. What would it take for us to come to consensus around these changes?

tstromberg:
Closing this PR as stale. Please reopen this PR when you are ready to retake a look at it. Thank you for your contribution!

Issue #382, opened 2020-12-10
Worker tries to execute the workflow even if pulling image from registry fails


invidian:
## Expected Behaviour

Workflow execution should stop if image can be pulled.

## Current Behaviour

Even if pulling image from registry fails, worker still tries to execute the workflow and then complains that the image cannot be found.

## Possible Solution

Make sure pulling errors are handled properly.

## Steps to Reproduce (for bugs)

1. Create a worker machine with 4GB of RAM to make `tmpfs` in OSIE of size 2GB.
2. Create a workflow with Docker image bigger than 1GB (e.g. `ubuntu` with `qemu` and `libguestfs-tools` packages installed).
3. Try running a workflow.

Cbkhare:
hey @invidian ,  nice catch! 

I tried reproducing the issue and below are my findings. 

- The error which is visible to us is actually not the root cause. Workflow execution has failed way before while pulling the image. It can be seen in the tink-worker logs with message `no space left on the device`. 
- What we are observing is that the exit should have been before creating the container, which didn't really happened. 
- Reason for the problem here is that [io.copy](https://github.com/tinkerbell/tink/blob/a41538da8b478b4d46aa5ab88d7dccfb54f9273d/cmd/tink-worker/internal/registry.go#L67) never really received the syscall of failure and eventually didn't returned the same. 
- And lately, our workflow execution went ahead for [creating](https://github.com/tinkerbell/tink/blob/master/cmd/tink-worker/internal/worker.go#L104) the container and failed. 


I tried to pinpoint the error by debugging it with delve. What seems to cause the failure is during the [Copy](https://github.com/golang/go/blob/master/src/io/io.go#L429). We can notice in the below logs after docker interrupt, Read returned `EOF` which for copy is graceful exit instead of an error. Usually, io.Copy returns the error when failed for `no space left on the device` but with docker pull I am expecting the interrupt is not received by the go compiler. 


```{"status":"Extracting","progressDetail":{"current":375681641,"total":375681641},"progress":"[==================================================\u003e]  375.7MB/375.7MB","id":"7899dcd9f509"}
[...]

PR #450, opened 2021-03-08
docker graceful exits when image pull fails #382

Cbkhare:
## Description

Docker image pull will do a graceful exit when an error is encountered.

## Why is this needed

- currently, with the go library `io.Copy`, an error is generated if the image pull fails, primarily when space is not left on the device.
- More details: https://github.com/tinkerbell/tink/issues/382#issuecomment-789672128 

Fixes: #

- Have removed the `io.Copy` 
- Added pull of an image and Error check via a struct. 

## How Has This Been Tested?
- Tested on Vagrant setup. 
- Test with the below template. 
```
version: "0.1"
name: helloworkflow
global_timeout: 600
tasks:
  - name: "hello world 1"
    worker: "{{.device_1}}"
    actions:
      - name: "hello_world_1"
        image: hello-world
        timeout: 60
      - name: "hello_world_2"
        image: test
        timeout: 200

```

- Test logs covers both +ve and -ve scearios.
- success with `hello-world` image and failure with a very large size dummy image of around 1 GB. 
```
{"level":"info","ts":1615207865.9741204,"caller":"cmd/root.go:132","msg":"no config file found","service":"github.com/tinkerbell/tink"}
{"level":"info","ts":1615207865.9748633,"caller":"cmd/root.go:53","msg":"starting","service":"github.com/tinkerbell/tink","version":"d0780ce"}
{"level":"info","ts":1615207865.992383,"caller":"internal/worker.go:270","msg":"starting with action","service":"github.com/tinkerbell/tink","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","actionName":"hello_world_1","taskName":"hello world 1"}
{"level":"info","ts":1615207865.9968224,"caller":"internal/worker.go:293","msg":"sent action status","service":"github.com/tinkerbell/tink","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","actionName":"hello_world_1","taskName":"hello world 1","duration":"0"}
{"level":"info","ts":1615207866.195519,"caller":"internal/action.go:54","msg":"creating container","service":"github.com/tinkerbell/tink","command":[]}
{"level":"info","ts":1615207866.2004075,"caller":"internal/worker.go:108","msg":"container created","service":"github.com/tinkerbell/tink","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","actionName":"hello_world_1","actionImage":"hello-world","containerID":"42482e6c52c6d13a44b8fc2456979409c9a4051ddf78b45c72ca7d3c9802daa7","command":[]}

Hello from Docker!
This message shows that your installation appears to be working correctly.

To generate this message, Docker took the following steps:
 1. The Docker client contacted the Docker daemon.
 2. The Docker daemon pulled the "hello-world" image from the Docker Hub.
    (amd64)
 3. The Docker daemon created a new container from that image which runs the
    executable that produces the output you are currently reading.
 4. The Docker daemon streamed that output to the Docker client, which sent it
    to your terminal.

To try something more ambitious, you can run an Ubuntu container with:
 $ docker run -it ubuntu bash

Share images, automate workflows, and more with a free Docker ID:
 https://hub.docker.com/

For more examples and ideas, visit:
 https://docs.docker.com/get-started/

{"level":"info","ts":1615207866.688619,"caller":"internal/worker.go:143","msg":"container removed","service":"github.com/tinkerbell/tink","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","actionName":"hello_world_1","actionImage":"hello-world","status":"STATE_SUCCESS"}
{"level":"info","ts":1615207866.6886694,"caller":"internal/worker.go:186","msg":"action container exited","service":"github.com/tinkerbell/tink","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","actionName":"hello_world_1","actionImage":"hello-world","status":"STATE_SUCCESS"}
{"level":"info","ts":1615207866.688689,"caller":"internal/action.go:115","msg":"removing container","service":"github.com/tinkerbell/tink","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","actionName":"hello_world_1","actionImage":"hello-world","containerID":"42482e6c52c6d13a44b8fc2456979409c9a4051ddf78b45c72ca7d3c9802daa7"}
{"level":"info","ts":1615207866.6984775,"caller":"internal/worker.go:334","msg":"sent action status","service":"github.com/tinkerbell/tink","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","actionName":"hello_world_1","taskName":"hello world 1"}
{"level":"info","ts":1615207866.712187,"caller":"internal/worker.go:293","msg":"sent action status","service":"github.com/tinkerbell/tink","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","actionName":"hello_world_2","taskName":"hello world 1","duration":"0"}
{"level":"error","ts":1615207891.0433881,"caller":"internal/worker.go:319","msg":"DOCKER PULL: DOCKER PULL: failed to register layer: Error processing tar file(exit status 1): write /root/test123/570bbc7011b5f6155f4631fc33753f02f7e8288bd8f2fac20d766e437e8f1bbe/layer.tar: no space left on device","service":"github.com/tinkerbell/tink","workerID":"0eba0bf8-3772-4b4a-ab9f-6ebe93b90a94","workflowID":"b19960b8-800c-11eb-ba2c-0242ac120005","actionName":"hello_world_2","taskName":"hello world 1","error":"DOCKER PULL: DOCKER PULL: failed to register layer: Error processing tar file(exit status 1): write /root/test123/570bbc7011b5f6155f4631fc33753f02f7e8288bd8f2fac20d766e437e8f1bbe/layer.tar: no space left on device","errorVerbose":"failed to register layer: Error processing tar file(exit status 1): write /root/test123/570bbc7011b5f6155f4631fc33753f02f7e8288bd8f2fac20d766e437e8f1bbe/layer.tar: no space left on device\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*RegistryConnDetails).pullImage\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/registry.go:86\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*Worker).execute\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/worker.go:101\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*Worker).ProcessWorkflowActions\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/worker.go:301\ngithub.com/tinkerbell/tink/cmd/tink-worker/cmd.NewRootCommand.func2\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/cmd/root.go:74\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:842\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:950\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:887\nmain.main\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/main.go:29\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1373\nDOCKER PULL\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*RegistryConnDetails).pullImage\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/registry.go:86\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*Worker).execute\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/worker.go:101\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*Worker).ProcessWorkflowActions\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/worker.go:301\ngithub.com/tinkerbell/tink/cmd/tink-worker/cmd.NewRootCommand.func2\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/cmd/root.go:74\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:842\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:950\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:887\nmain.main\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/main.go:29\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1373\nDOCKER PULL\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*Worker).execute\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/worker.go:102\ngithub.com/tinkerbell/tink/cmd/tink-worker/internal.(*Worker).ProcessWorkflowActions\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/internal/worker.go:301\ngithub.com/tinkerbell/tink/cmd/tink-worker/cmd.NewRootCommand.func2\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/cmd/root.go:74\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:842\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:950\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/Chitrabasu/Downloads/work/gocode/pkg/mod/github.com/spf13/cobra@v1.0.1-0.20200713175500-884edc58ad08/command.go:887\nmain.main\n\t/home/Chitrabasu/Downloads/work/repo/tink/cmd/tink-worker/main.go:29\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:203\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1373"}
Error: worker Finished with error: DOCKER PULL: DOCKER PULL: failed to register layer: Error processing tar file(exit status 1): write /root/test123/570bbc7011b5f6155f4631fc33753f02f7e8288bd8f2fac20d766e437e8f1bbe/layer.tar: no space left on device
Usage:
  tink-worker [flags]

Flags:
      --capture-action-logs        Capture action container output as part of worker logs (default true)
  -r, --docker-registry string     Sets the Docker registry (DOCKER_REGISTRY)
  -h, --help                       help for tink-worker
  -i, --id string                  Sets the worker id (ID)
      --max-file-size int          Maximum file size in bytes (MAX_FILE_SIZE) (default 10485760)
      --max-retry int              Maximum number of retries to attempt (MAX_RETRY) (default 3)
  -p, --registry-password string   Sets the registry-password (REGISTRY_PASSWORD)
  -u, --registry-username string   Sets the registry username (REGISTRY_USERNAME)
      --retry-interval duration    Retry interval in seconds (RETRY_INTERVAL) (default 3ns)
      --timeout duration           Max duration to wait for worker to complete (TIMEOUT) (default 1h0m0s)
  -v, --version                    version for tink-worker
```

Please note:
This change, additionally, alters the details of logs collection when an image is pulled. This is to avoid unnecessary details in the logs. Same has been concluded to remove via #451

### DIFF
--- a/cmd/tink-worker/internal/registry.go
+++ b/cmd/tink-worker/internal/registry.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"io"
-	"os"
 
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
@@ -20,6 +19,17 @@ type RegistryConnDetails struct {
 	user,
 	pwd string
 	logger log.Logger
+}
+
+// ImagePullStatus is the status of the downloaded Image chunk
+type ImagePullStatus struct {
+	Status         string `json:"status"`
+	Error          string `json:"error"`
+	Progress       string `json:"progress"`
+	ProgressDetail struct {
+		Current int `json:"current"`
+		Total   int `json:"total"`
+	} `json:"progressDetail"`
 }
 
 // NewRegistryConnDetails creates a new RegistryConnDetails
@@ -46,8 +56,12 @@ func (r *RegistryConnDetails) NewClient() (*client.Client, error) {
 	return c, nil
 }
 
+type imagePuller interface {
+	ImagePull(context.Context, string, types.ImagePullOptions) (io.ReadCloser, error)
+}
+
 // pullImage outputs to stdout the contents of the requested image (relative to the registry)
-func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client, image string) error {
+func (r *RegistryConnDetails) pullImage(ctx context.Context, cli imagePuller, image string) error {
 	authConfig := types.AuthConfig{
 		Username:      r.user,
 		Password:      r.pwd,
@@ -64,8 +78,18 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client,
 		return errors.Wrap(err, "DOCKER PULL")
 	}
 	defer out.Close()
-	if _, err := io.Copy(os.Stdout, out); err != nil {
-		return err
+	fd := json.NewDecoder(out)
+	var status *ImagePullStatus
+	for {
+		if err := fd.Decode(&status); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return errors.Wrap(err, "DOCKER PULL")
+		}
+		if status.Error != "" {
+			return errors.Wrap(errors.New(status.Error), "DOCKER PULL")
+		}
 	}
 	return nil
 }
--- a/cmd/tink-worker/internal/registry_test.go
+++ b/cmd/tink-worker/internal/registry_test.go
@@ -1,0 +1,75 @@
+package internal
+
+import (
+	"context"
+	"errors"
+	"io"
+	"io/ioutil"
+	"strings"
+	"testing"
+
+	"github.com/docker/docker/api/types"
+	"github.com/packethost/pkg/log"
+	//"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+func setupTestLogger(t *testing.T) log.Logger {
+	service := "github.com/tinkerbell/tink"
+	logger, err := log.Init(service)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return logger
+}
+
+type imagePullerMock struct {
+	stringReadCloser io.ReadCloser
+	imagePullErr     error
+}
+
+func (d *imagePullerMock) ImagePull(ctx context.Context, str string, op types.ImagePullOptions) (io.ReadCloser, error) {
+	return d.stringReadCloser, d.imagePullErr
+}
+
+func TestPullImageAnyFailure(t *testing.T) {
+	for _, test := range []struct {
+		testName         string
+		testString       string
+		testImagePullErr error
+		testErr          error
+	}{
+		{
+			testName:         "success",
+			testString:       "{\"status\": \"hello\",\"error\":\"\"}{\"status\":\"world\",\"error\":\"\"}",
+			testImagePullErr: nil,
+			testErr:          nil,
+		},
+		{
+			testName:         "fail",
+			testString:       "{\"error\": \"\"}",
+			testImagePullErr: errors.New("Tested, failure of the image pull"),
+			testErr:          errors.New("DOCKER PULL: Tested, failure of the image pull"),
+		},
+		{
+			testName:         "fail_partial",
+			testString:       "{\"status\": \"hello\",\"error\":\"\"}{\"status\":\"world\",\"error\":\"Tested, failure of No space left on device\"}",
+			testImagePullErr: nil,
+			testErr:          errors.New("DOCKER PULL: Tested, failure of No space left on device"),
+		},
+	} {
+		t.Run(test.testName, func(t *testing.T) {
+			ctx := context.Background()
+			rcon := NewRegistryConnDetails("test", "testUser", "testPwd", setupTestLogger(t))
+			stringReader := strings.NewReader(test.testString)
+			cli := &imagePullerMock{
+				stringReadCloser: ioutil.NopCloser(stringReader),
+				imagePullErr:     test.testImagePullErr,
+			}
+			err := rcon.pullImage(ctx, cli, test.testName)
+			if test.testErr != nil {
+				assert.Equal(t, err.Error(), test.testErr.Error())
+			}
+		})
+	}
+}
--- a/cmd/tink-worker/internal/registry_test.go
+++ b/cmd/tink-worker/internal/registry_test.go
@@ -10,7 +10,6 @@ import (
 
 	"github.com/docker/docker/api/types"
 	"github.com/packethost/pkg/log"
-	//"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -69,6 +68,8 @@ func TestPullImageAnyFailure(t *testing.T) {
 			err := rcon.pullImage(ctx, cli, test.testName)
 			if test.testErr != nil {
 				assert.Equal(t, err.Error(), test.testErr.Error())
+			} else {
+				assert.Equal(t, err, test.testErr)
 			}
 		})
 	}


cmd/tink-worker/internal/registry.go
```diff
@@ -64,8 +74,17 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client,
...
+	var status *ImagePullStatus
+	for {
+		if err := fd.Decode(&status); err != nil {
+			if err == io.EOF {
+				break
+			}
+		}
+		if status.Error != "" {
+			return errors.Wrap(errors.New(status.Error), "DOCKER PULL")
+		}
```
invidian, line 87: This changes the behavior and now hides the pulling logs, which I guess might be helpful while pulling large image.

cmd/tink-worker/internal/registry.go
```diff
@@ -64,8 +74,17 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client,
 		return errors.Wrap(err, "DOCKER PULL")
 	}
 	defer out.Close()
-	if _, err := io.Copy(os.Stdout, out); err != nil {
```
invidian, line 67: Hm, I think I don't understand. so `cli.ImagePull` does not return error when you run out of disk space? This seems like a bug to me.

cmd/tink-worker/internal/registry.go
Cbkhare: yes, that would be only possible if we print the `status`.

cmd/tink-worker/internal/registry.go
Cbkhare: yes, it is a bug. I think it is probably with docker. 

As of now, this change gives us leverage to bypass the error gracefully. But eventually, a proper fix on the go-client or docker would be required.

Cbkhare: IMO, i would have personally preferred the `io.Copy` catching that error. 

On our last syncup call me and @thebsdbox discussed this in details, so we thought to have a workaround ready at least for now. 

I also, think an alternative way could be to document this and ask the user to increase the disk space when this error is encountered.

invidian: Has this been reported upstream? I'm also the user of `cli.ImagePull`, so I'm curious under what conditions it may fail, but not return error.

cmd/tink-worker/internal/registry.go
invidian: Maybe we can use [TeeReader](https://golang.org/pkg/io/#TeeReader) to both print the messages and decode them to retain the old behavior?

cmd/tink-worker/internal/registry.go
Cbkhare: I think probably the issue is with `io.Copy` writing stream to stdout instead of `cli.ImagePull`.

cmd/tink-worker/internal/registry.go
Cbkhare: so TeeReader will return a Reader object and data will be pushed to stdout when we read it. Which will fail but go won't catch the error. I tested it, while downloading the image it failed in between, it didn't return the error.

invidian: I think we can defer io.Copy on tee reader to keep the output, then read and decode it as done in this PR. What do you think?

cmd/tink-worker/internal/registry.go
invidian: `io.Copy` happens after `cli.ImagePull` error checking. If Image pulling fail, I'd expect no logs will be printed.

cmd/tink-worker/internal/registry.go
Cbkhare: I don't think defer will make a difference. Both of them eventually will call the `read`. Secondly, reader object can not be read twice. 
Even if we mae it work, IMHO that would be an overhead for the `image pull` method. 
Do you have any specific use case where logs are required ? 

I think, one of the below three an be an approach:
1) keep `io.Copy` and document the error. 
2) keep the code change as in PR. For logs, we can add a logger from a separate issue. 
3) File a issue in repo of docker. Actually, you will find the issue on docker repo with the same context, and suggestions are there are to increase the host size. I didn't got lucky to find some other suggestions.

cmd/tink-worker/internal/registry.go
Cbkhare: The image is pulled in streams and read by `io.Copy` in chunks. Once one of the chunk is read EOF is generated. When read of second chunk fails, the interrupt is either not generated or go compiler doesn't receives it, so in a way for go error never occurred and that is the reason why go is not sending the error back. I have pasted the logs in the issue #382 .

cmd/tink-worker/internal/registry.go
invidian: >Do you have any specific use case where logs are required ?

It's the progress indicator, it's always nice to see in logs that something is happening. But if maintainers agrees this is not very important, I'm fine with it.

>Secondly, reader object can not be read twice.

This is why you could use TeeReader, to duplicate the stream.

I also found https://github.com/moby/moby/issues/36253 now, it seems `ImagePull` has even more bugs :|

cmd/tink-worker/internal/registry.go
invidian: I'm not sure what are you referring to as "go compiler", but I don't think the compiler is involved in here. Maybe you meant _Go runtime_? 

So also looking at https://github.com/moby/moby/issues/36253, it seems `cli.ImagePull` does not really return error when pulling image fails, but streams the possible errors via the reader. This is hell-confusing API :smile:

cmd/tink-worker/internal/registry.go
Cbkhare: > I also found moby/moby#36253 now, it seems ImagePull has even more bugs :|

Woh! this is surprising, I didn't notice this. 

> This is why you could use TeeReader, to duplicate the stream.

I tried to replicate this scenario, as in below code snippet, but I didn't observed that object can be streamed/read again. Can you share me an example on what you are suggesting, 

```	reader, err := cli.ImagePull(ctx, "docker.io/cbkhare/test", types.ImagePullOptions{}) //RegistryAuth: authStr})
	if err != nil {
		fmt.Println(reader)
		return
	}
	r := io.TeeReader(reader, os.Stdout)

        // streaming the object
	_, err= io.ReadAll(r)
        if err != nil {
                return err
        }

        // re streaming the object. 
        fd := json.NewDecoder(r)

        type Status struct {
        Status         string `json:"status"`
        Error          string `json:"error"`
        Progress       string `json:"progress"`
        ProgressDetail struct {
            Current int `json:"current"`
            Total   int `json:"total"`
            } `json:"progressDetail"`
        }
	var status *Status

        for {
        if err := fd.Decode(&status); err != nil {
	    if err == io.EOF {
		    break
            }
            fmt.Println("caught ", err, status)
        }
        if status.Error != "" {
		fmt.Println("Failed ", status.Error)
	}
	fmt.Printf("Status: %+v\n", status)
        }
        fmt.Println("done")
```

if the error is caught correctly, done shouldn't be printed.

invidian: I was thinking about something like:
```go
package main

import (
        "bytes"
        "encoding/json"
        "fmt"
        "io"
        "io/ioutil"
        "os"
        "strings"
)

// ImagePullStatus is the status of the downloaded Image chunk
type ImagePullStatus struct {
        Status         string `json:"status"`
        Error          string `json:"error"`
        Progress       string `json:"progress"`
        ProgressDetail struct {
                Current int `json:"current"`
                Total   int `json:"total"`
        } `json:"progressDetail"`
}

func main() {
        r := ioutil.NopCloser(strings.NewReader("hello world")) // r type is io.ReadCloser

        // Ensure ReadCloser will be closed and error will be handled.
        //
        // Error handling is not critical for execution, so only inform user
        // if error occurs, but proceed with execution.
        defer func() {
                fmt.Println("Closing reader")
                if err := r.Close(); err != nil {
                        fmt.Printf("Closing reader: %v\n", err)
                }
        }()

        // Make a copy of logs produced by reader into buffer, so it can be copied into stdout.
        var bufferRead bytes.Buffer

        reader := io.TeeReader(r, &bufferRead)

        // Stream copied logs.
        //
        // Error handling is not critical for execution, so only inform user
        // if error occurs, but proceed with execution.
        go func() {
                fmt.Println("Copying logs to stdout")

                if _, err := io.Copy(&bufferRead, os.Stdout); err != nil {
                        fmt.Printf("Writing logs to     stdout: %v\n", err)
                }
        }()

        fd := json.NewDecoder(reader)

        var status *ImagePullStatus

        fmt.Println("Entering decode loop")
        for {
                if err := fd.Decode(&status); err != nil {
                        if err == io.EOF {
                                break
                        }

                        fmt.Println("Decoding error occurred: %v", err)
                        continue
                }

                if status.Error != "" {
                        fmt.Printf("Error occurred: %v", status.Error)
                        return
                }
        }
}
```

This loops infinitely, it shouldn't, but I hope you get the idea.

cmd/tink-worker/internal/registry.go
```diff
@@ -64,8 +74,17 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client,
...
-		return err
+	fd := json.NewDecoder(out)
+	var status *ImagePullStatus
+	for {
+		if err := fd.Decode(&status); err != nil {
+			if err == io.EOF {
+				break
+			}
+		}
+		if status.Error != "" {
```
invidian, line 85: If error occurs but it's not EOF, deferencing `status` for `Error` field will panic.

Cbkhare: okay, I have added a return statement if `Decode` fails.

cmd/tink-worker/internal/registry.go
```diff
@@ -64,8 +74,18 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client,
...
+	for {
+		if err := fd.Decode(&status); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return errors.Wrap(err, "DOCKER PULL")
+		}
+		if status.Error != "" {
+			return errors.Wrap(errors.New(status.Error), "DOCKER PULL")
+		}
```
invidian, line 88: As we discussed on Slack, if we add `fmt.Printf("%+v\n", status)` below this line, we retain logging functionality and everything should work nicely then.

cmd/tink-worker/internal/registry.go
invidian: I created https://github.com/moby/moby/issues/42126, maybe they improve it somehow...

cmd/tink-worker/internal/registry.go
gianarb: Hey! A couple of things! @invidian thank you for catching this. @Cbkhare you should limit the scope of a PR to a minimum. Changing the logging behavior is probably out of scope. 

A few days ago something funny happened. Another member of the community asked for the opposite of what @invidian wants. Let's discuss "image pull logging" there because as I said, this change is out of scope for this PR.

https://github.com/tinkerbell/tink/issues/451

gianarb:
Ok! As we discussed here https://github.com/tinkerbell/tink/issues/451 we can remove the progress bar because it is not really a good way when to communicate what a program is doing in this context. It is good for humans, but not ideal in this scenario. 

So, even if it was not part of this work let's do it here in a separate commit @Cbkhare 

Thanks a lot

Cbkhare:
@gianarb Separate commit wont be required. Added change itself covers that. We can just edit the details in the PR, add context of 451 and reference all the details. 

@invidian would you also like to add your any other input.

invidian (dismissed):
I think commits can be squashed, but it looks good to me :+1:

cmd/tink-worker/internal/registry.go
invidian: Not relevant anymore. See #451.

cmd/tink-worker/internal/registry.go
```diff
@@ -64,8 +74,18 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli *client.Client,
 		return errors.Wrap(err, "DOCKER PULL")
 	}
 	defer out.Close()
-	if _, err := io.Copy(os.Stdout, out); err != nil {
-		return err
+	fd := json.NewDecoder(out)
+	var status *ImagePullStatus
+	for {
```
invidian, line 79: I think we could add some units tests for this. They shouldn't be too hard to write and I think it would be better to have some internal tests, rather than no tests.

mmlb:
Will a docker pull error cause tink-worker to exit?

Cbkhare:
> Will a docker pull error cause tink-worker to exit?

yes, currently if there is a docker pull error then tink-worker exits

mmlb:
> > Will a docker pull error cause tink-worker to exit?
> 
> yes, currently if there is a docker pull error then tink-worker exits

We should not do this, this will be a hindrance to EM adoption and we'd have to patch that out. tink-worker should always be up to hopefully try and recover.

Cbkhare:
> > > Will a docker pull error cause tink-worker to exit?
> > 
> > 
> > yes, currently if there is a docker pull error then tink-worker exits
> 
> We should not do this, this will be a hindrance to EM adoption and we'd have to patch that out. tink-worker should always be up to hopefully try and recover.

This change does not alter the tink-worker existing functionality. This just does a gracefully exit and returns a proper error, instead of bailing out with a random error. 

IMHO, If we wish to remove the exit of tink-worker when docker failure occurs, then we can create a separate issue and fix it..

mmlb:
> > > > Will a docker pull error cause tink-worker to exit?
> > > 
> > > 
> > > yes, currently if there is a docker pull error then tink-worker exits
> > 
> > 
> > We should not do this, this will be a hindrance to EM adoption and we'd have to patch that out. tink-worker should always be up to hopefully try and recover.
> 
> This change does not alter the tink-worker existing functionality. This just does a gracefully exit and returns a proper error, instead of bailing out with a random error.
> 
> IMHO, If we wish to remove the exit of tink-worker when docker failure occurs, then we can create a separate issue and fix it..

Yep that sounds good, I was likely to do the same :D

mmlb:
#463

cmd/tink-worker/internal/registry.go
```diff
@@ -46,8 +56,12 @@ func (r *RegistryConnDetails) NewClient() (*client.Client, error) {
 	return c, nil
 }
 
+type ImagePuller interface {
```
invidian, line 59: Does this interface need to be exported?

cmd/tink-worker/internal/registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+	} {
+		t.Run(test.testName, func(t *testing.T) {
+			ctx := context.Background()
+			rcon := NewRegistryConnDetails("test", "testUser", "testPwd", setupTestLogger())
+			cli := &imagePullerMock{}
+			err := rcon.pullImage(ctx, cli, test.testName)
+			if test.err != nil {
+				assert.Error(t, err)
+				return
+			}
```
invidian, line 69: I think test should also do some assertions when error is _not_ expected, but it was returned.

cmd/tink-worker/internal/registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+func (d *imagePullerMock) ImagePull(ctx context.Context, str string, op types.ImagePullOptions) (io.ReadCloser, error) {
+	var s string
+	var err error
+	if str == "test/success" {
+		s = "{\"status\": \"hello\",\"error\":\"\"}{\"status\":\"world\",\"error\":\"\"}"
+	} else if str == "test/fail" {
+		s = "{\"error\": \"\"}"
+		err = errors.New("Tested, failure of the image pull")
+	} else if str == "test/fail_partial" {
+		s = "{\"status\": \"hello\",\"error\":\"\"}{\"status\":\"world\",\"error\":\"Tested, failure of No space left on device\"}"
```
invidian, line 36: It would be nice to have this data defined directly in the test case, to keep things simple.

cmd/tink-worker/internal/registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+	return logger
+}
+
+type imagePullerMock struct{}
+
+func (d *imagePullerMock) ImagePull(ctx context.Context, str string, op types.ImagePullOptions) (io.ReadCloser, error) {
+	var s string
+	var err error
+	if str == "test/success" {
+		s = "{\"status\": \"hello\",\"error\":\"\"}{\"status\":\"world\",\"error\":\"\"}"
```
invidian, line 31: There should only be a test case when decoding fails, as IMO it's important part of the spec, to decide on what to do if Docker throws some unexpected message.

cmd/tink-worker/internal/registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+	} else if str == "test/fail" {
+		s = "{\"error\": \"\"}"
+		err = errors.New("Tested, failure of the image pull")
+	} else if str == "test/fail_partial" {
+		s = "{\"status\": \"hello\",\"error\":\"\"}{\"status\":\"world\",\"error\":\"Tested, failure of No space left on device\"}"
+	}
+	stringReader := strings.NewReader(s)
+	stringReadCloser := ioutil.NopCloser(stringReader)
+	return stringReadCloser, err
```
invidian, line 40: It would be nice to also test case where `ImagePull` returns error directly, not via decoder. This mock method could return an error when given string is empty.

cmd/tink-worker/internal/registry_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/packethost/pkg/log"
+	"github.com/stretchr/testify/assert"
+)
+
+func setupTestLogger() log.Logger {
+	service := "github.com/tinkerbell/tink"
+	logger, err := log.Init(service)
+	if err != nil {
+		panic(err)
```
invidian, line 20: https://github.com/golang/go/wiki/CodeReviewComments#dont-panic

I'd use `t.Fatalf` instead and pass `*testing.T` as a parameter to this function.

cmd/tink-worker/internal/registry.go
Cbkhare: done.

cmd/tink-worker/internal/registry_test.go
Cbkhare: second test case does that. `Tested, failure of the image pull`

Cbkhare: third case does that. In that scenario, while reading the stream docker will bail out with `No space left on device`

cmd/tink-worker/internal/registry_test.go
Cbkhare: second case that `Tested, failure of the image pull`

cmd/tink-worker/internal/registry_test.go
Cbkhare: done

cmd/tink-worker/internal/registry_test.go
Cbkhare: Initially, I thought it could be better, but then I have made changes as per your suggestion.

cmd/tink-worker/internal/registry_test.go
Cbkhare: I think, for success case, only possible scenario is that err would be nil. Which we are covering implicitly in the assertion.

invidian (commented):
Mostly LGTM. We could test one more scenario:
```diff
diff --git cmd/tink-worker/internal/registry.go cmd/tink-worker/internal/registry.go
index 97bc13a..2f177ba 100644
--- cmd/tink-worker/internal/registry.go
+++ cmd/tink-worker/internal/registry.go
@@ -85,7 +85,7 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli imagePuller, im
                        if err == io.EOF {
                                break
                        }
-                       return errors.Wrap(err, "DOCKER PULL")
+                       //return errors.Wrap(err, "DOCKER PULL")
                }
                if status.Error != "" {
                        return errors.Wrap(errors.New(status.Error), "DOCKER PULL")
```

Right now with this patch, tests are still passing.

Cbkhare:
> Mostly LGTM. We could test one more scenario:
> 
> ```diff
> diff --git cmd/tink-worker/internal/registry.go cmd/tink-worker/internal/registry.go
> index 97bc13a..2f177ba 100644
> --- cmd/tink-worker/internal/registry.go
> +++ cmd/tink-worker/internal/registry.go
> @@ -85,7 +85,7 @@ func (r *RegistryConnDetails) pullImage(ctx context.Context, cli imagePuller, im
>                         if err == io.EOF {
>                                 break
>                         }
> -                       return errors.Wrap(err, "DOCKER PULL")
> +                       //return errors.Wrap(err, "DOCKER PULL")
>                 }
>                 if status.Error != "" {
>                         return errors.Wrap(errors.New(status.Error), "DOCKER PULL")
> ```
> 
> Right now with this patch, tests are still passing.

yes, that return of error is specific to `Decode`. Decode failure is not added in test cases and I think it is not required.

gianarb (approved):

thebsdbox (approved):
Excellent, thanks @Cbkhare

gauravgahlot:
@Cbkhare Could you update the branch, please?

PR #466, opened 2021-03-26
Fix migration ID

gauravgahlot:
## Why is this needed

The PR changes the migration ID:
`2020121691335-update-events-primary-key` to `202012169135-update-events-primary-key`.

### DIFF
--- a/db/migration/202012169135-update-events-primary-key.go
+++ b/db/migration/202012169135-update-events-primary-key.go
@@ -12,9 +12,9 @@ import migrate "github.com/rubenv/sql-migrate"
 //
 // The migration changes the primary key on events table from 'created_at' to `id`,
 // which will always be unique for each event generated at any point in time.
-func Get2020121691335() *migrate.Migration {
+func Get202012169135() *migrate.Migration {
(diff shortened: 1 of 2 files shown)


gianarb:
we should not change old migration because it can be already been applied to some environments. Changing the ID will break the migration process for those envs. Having an ID that is not formatted as it should be is wrong but we didn't catch it on time and the cost of changing it is too high.

gauravgahlot:
> we should not change old migration because it can be already been applied to some environments. Changing the ID will break the migration process for those envs. Having an ID that is not formatted as it should be is wrong but we didn't catch it on time and the cost of changing it is too high.
> 
> I am not sure what this is `Get202103261030`. I can't see this function anywhere in the codebase

`Get202103261030` was pushed by mistake. 

Due to incorrect ID, this migration is not applied in the correct order. This results in errors when remove `events` table in a separate migration.

gianarb:
Are you sure?! Technically the migrations are applied in the order they are appended to the slice! let's zoom so we can double check!

gianarb:
This is not solved by #467

PR #473, opened 2021-03-31
Improve empty string test

rayandas:
Signed-off-by: Rayan Das <rayandas91@gmail.com>

## Description

A string can be tested for its emptiness either by treating it as a slice and calculating the length of the slice, or by treating it as a string and directly comparing the value. While both produce identical code when compiled, it makes more sense to treat a string as itself, than a slice, for the sake of comparison of values.

For example:
```
len(xyz) == 0
```
can be rewritten as:
```
xyz == ""
```

## Why is this needed
The second method is considered more idiomatic.


gauravgahlot:
@rayandas Could you please update the branch?

rayandas:
@gauravgahlot updated.

gauravgahlot (approved):

Issue #475, opened 2021-04-06
Get command in tink-cli provides incorrect timestamp during get by Id


parauliya:
The output of `tink workflow get 6502e8ba-96a7-11eb-a69c-0242ac120005` command is as follows:

```
+--------------------------------------+--------------------------------------+---------------+----------------------+----------------------+
| ID                                   | TEMPLATE ID                          | STATE         | CREATED AT           | UPDATED AT           |
+--------------------------------------+--------------------------------------+---------------+----------------------+----------------------+
| 6502e8ba-96a7-11eb-a69c-0242ac120005 | 44548ced-96a7-11eb-a69c-0242ac120005 | STATE_PENDING | 1970-01-01T00:00:00Z | 1970-01-01T00:00:00Z |
+--------------------------------------+--------------------------------------+---------------+----------------------+----------------------+
```
And the output for `tink template get 44548ced-96a7-11eb-a69c-0242ac120005` command is as follows:
```
+--------------------------------------+----------------------+----------------------+----------------------+
| ID                                   | NAME                 | CREATED AT           | UPDATED AT           |
+--------------------------------------+----------------------+----------------------+----------------------+
| 44548ced-96a7-11eb-a69c-0242ac120005 | hello_world_workflow | 1970-01-01T00:00:00Z | 1970-01-01T00:00:00Z |
+--------------------------------------+----------------------+----------------------+----------------------+
```

And both of them exist because the output of `tink workflow get` and `tink template get` command is as follows:
```
+--------------------------------------+--------------------------------------+---------------+----------------------+----------------------+
| ID                                   | TEMPLATE ID                          | STATE         | CREATED AT           | UPDATED AT           |
+--------------------------------------+--------------------------------------+---------------+----------------------+----------------------+
| 6502e8ba-96a7-11eb-a69c-0242ac120005 | 44548ced-96a7-11eb-a69c-0242ac120005 | STATE_PENDING | 2021-04-06T07:12:06Z | 2021-04-06T07:12:06Z |
+--------------------------------------+--------------------------------------+---------------+----------------------+----------------------+
```

```
+--------------------------------------+----------------------+----------------------+----------------------+
| ID                                   | NAME                 | CREATED AT           | UPDATED AT           |
+--------------------------------------+----------------------+----------------------+----------------------+
| 44548ced-96a7-11eb-a69c-0242ac120005 | hello_world_workflow | 2021-04-06T07:11:11Z | 2021-04-06T07:11:11Z |
[...]

PR #476, opened 2021-04-08
Fix_475: Correct timestamps in get by Id commands for template and workflow

parauliya:

## Description

After this change the value of `createdAt` and `updatedAt` will be correct in the output of `tink template/workflow get <id>` command.

## Why is this needed
Fixes: #475 

## How Has This Been Tested?
This has been tested manually throgh vagrant setup.

No Impact

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/db/mock/mock.go
+++ b/db/mock/mock.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/tinkerbell/tink/db"
+	tb "github.com/tinkerbell/tink/protos/template"
 	pb "github.com/tinkerbell/tink/protos/workflow"
 )
 
@@ -25,5 +26,5 @@ type DB struct {
 	InsertIntoWorkflowEventTableFunc func(ctx context.Context, wfEvent *pb.WorkflowActionStatus, time time.Time) error
 	// template
 	TemplateDB      map[string]interface{}
-	GetTemplateFunc func(ctx context.Context, fields map[string]string, deleted bool) (string, string, string, error)
+	GetTemplateFunc func(ctx context.Context, fields map[string]string, deleted bool) (*tb.WorkflowTemplate, error)
 }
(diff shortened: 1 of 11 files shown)


db/mock/mock.go
```diff
@@ -25,5 +26,5 @@ type DB struct {
 	InsertIntoWorkflowEventTableFunc func(ctx context.Context, wfEvent *pb.WorkflowActionStatus, time time.Time) error
 	// template
 	TemplateDB      map[string]interface{}
-	GetTemplateFunc func(ctx context.Context, fields map[string]string, deleted bool) (string, string, string, error)
+	GetTemplateFunc func(ctx context.Context, fields map[string]string, deleted bool) (*tb.WorkflowTemplate, error)
```
Cbkhare, line 29: Replacing, set of strings with this struct `tb.WorkflowTemplate` is really good.

grpc-server/template_test.go
```diff
@@ -7,7 +7,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/tinkerbell/tink/db/mock"
-	pb "github.com/tinkerbell/tink/protos/template"
+	tb "github.com/tinkerbell/tink/protos/template"
```
gauravgahlot, line 10: If you don't rename the package referer, some changes in the file would go away.

gauravgahlot (changes requested):

grpc-server/template_test.go
parauliya: Yeah, I thought that earlier, but since we are using `pb` as a package referrer for `workflow` and `tb` for package referrer for `template` at many other places so if I use `pb` for `template` package referrer it can cause confusions around the code readers.

parauliya: I have replaced `tb` with `pb` since there was only template package in that file.

gauravgahlot (approved):
LTGM

PR #478, opened 2021-04-08
Add docker image building targets to Makefile

mmlb:
## Description

Adds make targets to build the docker images.

## Why is this needed

Its not very obvious that a user needs to run `make crosscompile` before they can try to build the docker images (as came up in community slack yesterday).
By having a proper target in the Makefile we might avoid this situation entirely.

## How Has This Been Tested?

Ran `make images`

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -140,8 +140,7 @@ jobs:
     - name: ${{ matrix.repository }}
       uses: docker/build-push-action@v2
       with:
-        context: .
-        file: cmd/${{ matrix.binary }}/Dockerfile
+        context: cmd/${{ matrix.binary }}/
         cache-from: type=registry,ref=${{ matrix.repository }}:latest
         push: ${{ startsWith(github.ref, 'refs/heads/master') }}
         tags: ${{ steps.docker-image-tag.outputs.tags }}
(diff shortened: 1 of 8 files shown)


gianarb:
Thanks, when I look at Makefile I am not able to use it. I tried weeks ago
but it always requires too much time and I end up writing my own thing...
Do you think we can write something as part of the README.md about how to
do common tasks like genering all the binaries? single binaries? Does cross
compile? docker images and maybe clean the environment?

Thanks a lot!! This is mainly a request I personally have, but probably
many others have the same issue :D

mmlb:
> Thanks, when I look at Makefile I am not able to use it. I tried weeks ago but it always requires too much time and I end up writing my own thing... Do you think we can write something as part of the README.md about how to do common tasks like genering all the binaries? single binaries? Does cross compile? docker images and maybe clean the environment? Thanks a lot!! This is mainly a request I personally have, but probably many others have the same issue :D

I added I think the most relevant bits of that to README.md already. I also reworked the Makefile so only the "important" targets are there and also added `make help` to it (which is also in the README). I think that should be good enough. Can you take a look at those and see if you want something tweaked/added?

mmlb:
@gianarb I was going to add the `cli`,`server`,`worker` pseudo targets to the Makefile instead of just as virtual deps but meh go compile is fast enough that I don't think it really matters to just build worker or cli or server, especially when make can do it in parallel.

Also clean is just better off with `git clean -fxd` imo.

gianarb (approved):
Thanks a lot!

PR #482, opened 2021-04-14
Fixed #477 : Corrected state for get and get by id command.

### LINKED ISSUE #477: "tink workflow get/get by id" command returns the incorrect state for workflows
"tink workflow get/get by id" command returns the incorrect state for workflows.
Right now it always returns state as STATE_PENDING for all the workflows which is incorrect because if a workflow is running it should have state as STATE_RUNNING and if finished successfully, it should have state as STATE_SUCCESSFUL.

## Expected Behaviour
"tink workflow get/get by id" command should return the current state for workflows.

## Current Behaviour
Right now it always returns state as STATE_PENDING for all the workflows which is incorrect.

## Possible Solution
Update the workflow state as soon as it started/finished/failed.

* Operating System Linux (Ubuntu 1804):

* How are you running Tinkerbell: Using Vagrant & VirtualBox

parauliya:

## Description
After this Fix user will get the correct state in `tink workflow get` and `tink workflow get <id>` 

## Why is this needed

Fixes: #477 

## How Has This Been Tested?
Tested Manually with Vagrant Setup.

No Impact.

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/grpc-server/workflow.go
+++ b/grpc-server/workflow.go
@@ -13,14 +13,6 @@ import (
 	wkf "github.com/tinkerbell/tink/workflow"
 )
 
-var state = map[int32]workflow.State{
-	0: workflow.State_STATE_PENDING,
-	1: workflow.State_STATE_RUNNING,
-	2: workflow.State_STATE_FAILED,
-	3: workflow.State_STATE_TIMEOUT,
-	4: workflow.State_STATE_SUCCESS,
-}
-
 const errFailedToGetTemplate = "failed to get template with ID %s"
 
 // CreateWorkflow implements workflow.CreateWorkflow
@@ -115,11 +107,12 @@ func (s *server) GetWorkflow(ctx context.Context, in *workflow.GetRequest) (*wor
 	if err != nil {
 		return &workflow.Workflow{}, err
 	}
+
 	wf := &workflow.Workflow{
 		Id:       w.ID,
 		Template: w.Template,
 		Hardware: w.Hardware,
-		State:    state[w.State],
+		State:    getWorkflowState(s.db, ctx, in.Id),
 		Data:     data,
 	}
 	l := s.logger.With("workflowID", w.ID)
@@ -181,6 +174,7 @@ func (s *server) ListWorkflows(_ *workflow.Empty, stream workflow.WorkflowServic
 			Hardware:  w.Hardware,
 			CreatedAt: w.CreatedAt,
 			UpdatedAt: w.UpdatedAt,
+			State:     getWorkflowState(s.db, stream.Context(), w.ID),
 		}
 		return stream.Send(wf)
 	})
@@ -278,4 +272,18 @@ func (s *server) ShowWorkflowEvents(req *workflow.GetRequest, stream workflow.Wo
 	s.logger.Info("done listing workflows events")
 	metrics.CacheHits.With(labels).Inc()
 	return nil
+}
+
+func getWorkflowState(db db.Database, ctx context.Context, id string) workflow.State {
+	wfctx, _ := db.GetWorkflowContexts(ctx, id)
+	if wfctx.CurrentActionState != workflow.State_STATE_SUCCESS {
+		return wfctx.CurrentActionState
+	} else {
+		wfacts, _ := db.GetWorkflowActions(ctx, id)
+		if isLastAction(wfctx, wfacts) {
+			return workflow.State_STATE_SUCCESS
+		} else {
+			return workflow.State_STATE_RUNNING
+		}
+	}
 }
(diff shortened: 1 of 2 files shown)


grpc-server/workflow.go
```diff
@@ -279,3 +273,17 @@ func (s *server) ShowWorkflowEvents(req *workflow.GetRequest, stream workflow.Wo
 	metrics.CacheHits.With(labels).Inc()
 	return nil
 }
+
+func getWorkflowState(db db.Database, ctx context.Context, id string) workflow.State {
```
jacobweinstock, line 277: any chance you could add code comments on why this logic is needed?

grpc-server/workflow.go
```diff
@@ -279,3 +273,17 @@ func (s *server) ShowWorkflowEvents(req *workflow.GetRequest, stream workflow.Wo
...
+	wfctx, _ := db.GetWorkflowContexts(ctx, id)
+	if wfctx.CurrentActionState != workflow.State_STATE_SUCCESS {
+		return wfctx.CurrentActionState
+	} else {
+		wfacts, _ := db.GetWorkflowActions(ctx, id)
```
jacobweinstock, line 282: doesnt `GetWorkflowContexts` already give `total_number_of_actions`? you could use that number and `current_action_index` to determine last action. It would save a database call.

jacobweinstock (commented):
any chance you could add some tests?

grpc-server/workflow.go
parauliya: Yeah, that is a good observation. Actually I did it beacuse the function `isLastAction()` was implemented in that way and asks for action lists as parameter. But yeah it is not required since we have total number of actions in `wfcontext` itself. I will change that. Thanks

grpc-server/workflow.go
parauliya: Yeah sure I will add reason behind this logic as comment over this function.

thebsdbox (approved):

grpc-server/workflow.go
```diff
@@ -279,3 +273,22 @@ func (s *server) ShowWorkflowEvents(req *workflow.GetRequest, stream workflow.Wo
...
+// For e.g. : If an action has Tailed or Timeout then the workflow state will also be
+// considered as Failed/Timeout. And If an action is successful then the workflow state
+// will be considered as Running until the last action of the workflow is executed successfully.
+
+func getWorkflowState(db db.Database, ctx context.Context, id string) workflow.State {
+	wfCtx, _ := db.GetWorkflowContexts(ctx, id)
+	if wfCtx.CurrentActionState != workflow.State_STATE_SUCCESS {
+		return wfCtx.CurrentActionState
+	} else {
+		//wfacts, _ := db.GetWorkflowActions(ctx, id)
```
jacobweinstock, line 287: should probably removed commented code

grpc-server/workflow.go
```diff
@@ -279,3 +273,22 @@ func (s *server) ShowWorkflowEvents(req *workflow.GetRequest, stream workflow.Wo
 	metrics.CacheHits.With(labels).Inc()
 	return nil
 }
+
+// This function will provide the workflow state on the basis of the state of the actions
+// For e.g. : If an action has Tailed or Timeout then the workflow state will also be
```
jacobweinstock, line 278: NIT: misspelled word "Tailed" -> "Failed"

grpc-server/workflow.go
```diff
@@ -181,6 +174,7 @@ func (s *server) ListWorkflows(_ *workflow.Empty, stream workflow.WorkflowServic
 			Hardware:  w.Hardware,
 			CreatedAt: w.CreatedAt,
 			UpdatedAt: w.UpdatedAt,
+			State:     getWorkflowState(s.db, stream.Context(), w.ID),
```
jacobweinstock, line 177: you're making 2 calls to `getWorkflowState`, this means 2 DB calls. You could call `getWorkflowState` just once (1 DB call) and store it to a variable for use in both locations.

jacobweinstock (commented):
thanks for adding tests! so important to making the tink codebase less fragile!
made a few comments for your review.

grpc-server/workflow.go
parauliya: I don't think it is possible to call `getWorkflowState` once and store it because the `workflow state` will change during the execution of a workflow which makes it dynamic. Therefore we have to fetch the `latest` value of a `workflow state`.

grpc-server/workflow.go
parauliya: Removed.

grpc-server/workflow.go
parauliya: Corrected

grpc-server/workflow.go
jacobweinstock: Ah, apologies. You are right. For some reason I thought I saw two calls in a single function.

jacobweinstock (approved):

Issue #481, opened 2021-04-11
generate-envrc.sh writes to "envrc" insted of ".env"


lukaskalvenas:
As asked by @gauravgahlot I'm writing an issue/reminder that generate-envrc.sh should write to ".env" and "setup.sh" should also look for ".env" instead of "envrc" :-)

lukaskalvenas:
@gauravgahlot I fixed the issue on my forked repo, along with another issue where template containers would not load on targets (issue https://github.com/tinkerbell/tink/issues/480)

Feel free to use if needed: https://github.com/lukaskalvenas/tink

PR #483, opened 2021-04-16
Change envrc to .env

gauravgahlot:
## Description

The setup scripts in `tinkerbell/tink` are different from that of `tinkerbell/sandbox`.

## Why is this needed

Fixes: #481

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,4 +14,5 @@ terraform.tfstate.backup
 # Vagrant
 **/.vagrant
 envrc
+.env
 deploy/state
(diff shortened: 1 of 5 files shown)


thebsdbox (dismissed):
`/LGTM` although I guess it's confusing that we have a `generate-envrc.sh` that doesn't generate an`envrc` 😂

gauravgahlot:
> `/LGTM` although I guess it's confusing that we have a `generate-envrc.sh` that doesn't generate an`envrc`

Surely 😄 I will update it.

thebsdbox (approved):
Thanks for sanitising these across the board.

Issue #485, opened 2021-04-22
tink-server fails to compare mac addresses


moadqassem:
## Expected Behaviour

When I create a new hardware, template and workflow, I expect to tink-server to match the worker mac address in pg with the provided mac address in the workflow request regardless the case sensitivity, as mac addresses are not case sensitive.
## Current Behaviour

tink-server fails to fetch the worker even though it has the same mac address but it is upper case.
## Possible Solution

Due to the merge of this PR https://github.com/tinkerbell/tink/pull/279, the mac address of the hardware data will be lowered and then persisted to avoid comparisons problems. However, if the you are creating a workflow based on the worker id, instead of the mac address(as the grpc client would do) it will fail to create the workflow. I opened a PR which should solve the issue:  

## Steps to Reproduce (for bugs)

1.create hardware data
2.create template data(add the worker mac address and with upper case letters)
3.create a workflow using the template and hardware id  
4. Workflow won't be created as this method here will fail to match and fetch mac addresses, because of the case sensitivity: https://github.com/tinkerbell/tink/blob/89f7a80d7ece500d2146fdab2a730ae0b19a5ab6/db/workflow.go#L686

## Your Environment

* Operating System and version (e.g. Linux, Windows, MacOS):

* How are you running Tinkerbell? Using Vagrant & VirtualBox, Vagrant & Libvirt, on Packet using Terraform, or give details:

* Link to your project or a code example to reproduce issue:

PR #486, opened 2021-04-22
fix getWorkerID when mac addresses have captalized letters

moadqassem:
## Why is this needed
To solve the worker not found errors, when the mac address of the worker is in upper case. 

Fixes: #485

## Checklist:

I have:

- [ ] added unit or e2e tests


tstromberg:
This PR looks great - is there any way we could have a unit test added to ensure that we don't break this again in the future?

moadqassem:
@tstromberg you can extend this case here https://github.com/tinkerbell/tink/blob/master/db/workflow_test.go#L42-L68 and and then do a fetch request or workflow get request with a device id that has the characters lower cases

gauravgahlot:
Hey @moadqassem, do you mind updating the branch and extend the test as well? It would be great to merge the PR then.

moadqassem:
@gauravgahlot I have added a test case to cover that https://github.com/tinkerbell/tink/pull/486/commits/d502edecbbad8a9144aa587b9967bc1bd454e5a2 

cc @tstromberg

gauravgahlot (approved):

gauravgahlot:
Thank you @moadqassem :rocket:

PR #501, opened 2021-05-23
remove the ports settings from the boots container docker-compose settings

rgl:
they are not supported when the container uses the host network_mode

this is the error given by docker-compose 1.29.2:

docker.errors.InvalidArgument: "host" network_mode is incompatible with port_binding


tstromberg:
@jacobweinstock @mmlb - can you take a look at this PR?

jacobweinstock (approved):
@rgl, thanks for this contribution!

mmlb:
lgtm

mmlb (approved):

PR #530, opened 2021-09-01
Adjust codespell invocation to only run codespell on .sh, .go, .md

### LINKED ISSUE #529: codespell is trying to fix checksums in go.mod
From https://github.com/tinkerbell/tink/pull/514/checks?check_run_id=3484813791

```
+ xargs codespell -q 3 -I .codespell-whitelist
+ git ls-files
go.sum:1420: Od ==> Of
+ failed=1
+ git ls-files '*.yml' '*.json' '*.md'
+ xargs prettier --check
Checking formatting...
All matched files use Prettier code style!
+ git ls-files '*.sh'
+ xargs shfmt -l -d
+ git ls-files '*.sh'
+ xargs shellcheck
+ nixfmt shell.nix
+ git diff
+ grep .
+ exit 1
```

We should either exclude `go.mod` or replace this check with the built-in golangci-lint equivalent.

tstromberg:
Fixes #529

### DIFF
--- a/ci-checks.sh
+++ b/ci-checks.sh
@@ -6,7 +6,8 @@ set -eux
 
 failed=0
 
-if ! git ls-files | xargs codespell -q 3 -I .codespell-whitelist; then
+# spell-checks only language files to avoid spell-checking checksums
+if ! git ls-files '*.sh' '*.go' '*.md' | xargs codespell -q 3 -I .codespell-whitelist; then
 	failed=1
 fi
 


ci-checks.sh
```diff
@@ -6,7 +6,8 @@ set -eux
 
 failed=0
 
-if ! git ls-files | xargs codespell -q 3 -I .codespell-whitelist; then
+# spell-checks only language files to avoid spell-checking checksums
+if ! git ls-files '*.sh' '*.go' '*.md' | xargs codespell -q 3 -I .codespell-whitelist; then
```
mmlb, line 10: what about the opposite, exclude `go.sum` and `go.mod`? This way any new files introduced are checked by default (which may or may not be a good thing ;) )

tstromberg: This was a conscious decision: The number of file extensions where it is valuable to spell-check code for is finite, whereas the number of file extensions that contain other data is unlimited:
 
* dockerignore
* editorconfig
* envrc
* gitignore
* idx
* json
* lock
* mk
* mod
* nix
* pack
* proto
* sample
* sum
* tmpl
* toml
* txt
* yaml
* yml

One could argue for adding some of these extensions - but a whitelist of code extensions will be far easier to maintain than a blacklist of data files.

mmlb: :+1: makes sense

mmlb (approved):

PR #533, opened 2021-09-01
setup_osie: Add --retry to curl, exit if it fails

### LINKED ISSUE #306: the file pull via curl is not robust and sometimes fail with out error / abort the script
the file pull via curl is not robust and sometimes fail with out error / abort the script

## Expected Behaviour
curl must catch the errors and abort the script.

## Current Behaviour
curl will fail silently

## Possible Solution
see the PR #305 

have a unstabel network

## Your Environment
local Environment

linux 
Vagrant & VirtualBox

tstromberg:
Fixes #306

Includes some minor refactoring to decrease the indentation level and adds some informational messages.

(PS: On that note, I really miss having a progress bar for curl, given that the download takes ~10 minutes here)


setup.sh
```diff
@@ -247,39 +247,50 @@ EOF
...
-			tar -zxf osie.tar.gz
-		else
-			if [ ! -f "$TB_OSIE_TAR" ]; then
-				echo "$ERR osie tar not found in the given location $TB_OSIE_TAR"
-				exit 1
-			fi
-			echo "$INFO extracting osie tar"
-			tar -zxf "$TB_OSIE_TAR"
+
+	if [ -d "$osie_current" ] && [ -d "$tink_workflow" ]; then
```
mmlb, line 253: this should probably check for 

```suggestion
	if [ -f "$osie_current/vmlinuz-x86_64" ] && [ -f "$tink_workflow/workflow-helper.sh" ]; then
```

instead. Otherwise if curl fails then a re-run will skip curl right? Because you're `mkdir -p` these dirs right after this branch

setup.sh
```diff
@@ -247,39 +247,50 @@ EOF
...
+		if mv workflow-helper.sh workflow-helper-rc "$tink_workflow"; then
+			cp -r ./* "$osie_current"
+		else
+			echo "$ERR failed to move 'workflow-helper.sh' and 'workflow-helper-rc'"
+			exit 1
 		fi
-	else
-		echo "$INFO found existing osie files, skipping osie setup"
+		popd
 	fi
```
mmlb, line 294: maybe this should actually be:

```suggestion
	if ! tar -zxf "$TB_OSIE_TAR" -C "$osie_current"; then
		echo "$ERR failed to extract osie tarball";
		exit 1;
	fi
	if ! mv/cp "$osie_current/workflow-helper.sh" "$osie_current/workflow-helper-rc" "$tink_workflow"; then
		echo "$ERR failed to move 'workflow-helper.sh' and 'workflow-helper-rc'";
		exit 1;
	fi
```
no need to change dir, ..., mv .* when `tar` can do what we want right? This way we don't have to deal with pushd errors (currently ignore, which tbf shouldn't ever really happen) but also doesn't keep 3 copies of osie around (the tarball, one in $cwd, and the one in $osie_current)... just 2 copies.

jacobweinstock:
Hey @tstromberg @mmlb. If I'm not mistaken, I don't think `setup.sh` is used anywhere anymore in this repo.

Issue #535, opened 2021-09-03
Tink server panic: `The unary server interceptor was already set and may not be reset.`


jacobweinstock:
It appears that the introduction of the otel interceptors has caused a runtime panic.
tink/grpc-server/grpc_server.go:57-58
```go
grpc.UnaryInterceptor(otelgrpc.UnaryServerInterceptor()),
grpc.StreamInterceptor(otelgrpc.StreamServerInterceptor()),
```

stack trace:

```bash
panic: The unary server interceptor was already set and may not be reset.

goroutine 1 [running]:
google.golang.org/grpc.UnaryInterceptor.func1(0xc00048e000)
        /Users/.../go/pkg/mod/google.golang.org/grpc@v1.39.0/server.go:382 +0x6b
google.golang.org/grpc.(*funcServerOption).apply(0xc0000ba200, 0xc00048e000)
        /Users/.../go/pkg/mod/google.golang.org/grpc@v1.39.0/server.go:201 +0x33
google.golang.org/grpc.NewServer(0xc000592280, 0x5, 0x8, 0x4)
        /Users/.../go/pkg/mod/google.golang.org/grpc@v1.39.0/server.go:564 +0xeb
github.com/tinkerbell/tink/grpc-server.SetupGRPC(0xe209b0, 0xc00064a840, 0xd21547, 0x1a, 0xc000012010, 0xc000568150, 0xc00064aac0, 0xc00006ca20, 0x0, 0x0, ...)
        /Users/.../repos/tinkerbell/tink/grpc-server/grpc_server.go:77 +0x319
main.NewRootCommand.func2(0xc0003f0000, 0x12f82f8, 0x0, 0x0, 0x0, 0x0)
        /Users/.../repos/tinkerbell/tink/cmd/tink-server/main.go:183 +0x812
github.com/spf13/cobra.(*Command).execute(0xc0003f0000, 0xc00009e200, 0x0, 0x0, 0xc0003f0000, 0xc00009e200)
        /Users/.../go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:856 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xc0003f0000, 0xd21547, 0x1a, 0xc000012010)
        /Users/.../go/pkg/mod/github.com/spf13/cobra@v1.2.1/command.go:974 +0x375
github.com/spf13/cobra.(*Command).Execute(...)
[...]

PR #536, opened 2021-09-03
Tink Server: Fix runtime panic

jacobweinstock:
## Description

Fix runtime `panic: The unary server interceptor was already set and may not be reset`
The grpc server interceptors need to be chained to not cause a runtime panic.

## Why is this needed

Fixes: #535 

## How Has This Been Tested?

`make test`
`make run`
`docker logs tink_tinkerbell_1` - validate the server started up successfully.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/golang/protobuf v1.5.2
 	github.com/google/go-cmp v0.5.6
 	github.com/google/uuid v1.2.0
+	github.com/grpc-ecosystem/go-grpc-middleware v1.0.1-0.20190118093823-f849b5445de4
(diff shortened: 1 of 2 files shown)


mmlb (approved):

mmlb (commented):
Is there a test we can add that would have caught this?

tobert (approved):
Whooooops. I didn't even notice that and am surprised the compiler didn't!

tobert:
> Is there a test we can add that would have caught this?

I should have started up the think stack before submitting the PR. I'll get my rig set up before I dig in more (tink-worker and tink need some more otel bits). Ideally we can come up with a functional test that lets us exercise the tink binaries a ilttle without bringing up the whole thing.

PR #543, opened 2021-09-29
propagate traceparent and span around core logic in tink-worker

tobert:
## Description

Add more tracing features to tink-worker.

Calls the new helper in otel-init-go to load traceparent from the kernel /proc/cmdline or $TRACEPARENT envvar.

Propagates TRACEPARENT environment variable to workflow Docker containers.

Updates to otel-init-go v0.0.4 for the helpers, which also brings in opentelemetry-go v1.0.0.

I dug into the worker code and discovered I cannot delve further. The code in cmd/tink-worker/internal/worker.go does not appear to have any test coverage. Furthermore, it seems to have unclear semantics and sensitivity to side-effects. Without tests I cannot add tracing to this code responsibly.

With this commit, we should see traces propagate to tink-worker and get one big span around the workflow code. gRPC and HTTP auto-instrumentation should pick us up some more detail beyond that.

We will probably not gain visibility into the Docker stuff unless the docker client happens to add its own spans, that code does not look safe for me to modify.

## Why is this needed

This will put a span around the tink-worker workflow code, and propagates traces down to workflow docker containers via environment variables.

Fixes: #

## How Has This Been Tested?

`make test` and I assume CI will run on this PR

Please see my earlier note about testing gaps that prevent higher-fidelity tracing.

Adds a trace, and Docker containers will now see the TRACEPARENT environment variable show up when tracing is active.


displague:
@tobert 

> We will probably not gain visibility into the Docker stuff unless the docker client happens to add its own spans, that code does not look safe for me to modify.

What changes would you expect or like to make to the client?
Would we benefit from registry servers that have observability enabled? 

```
func NewWorker(client pb.WorkflowServiceClient, regConn *RegistryConnDetails, logger log.Logger, registry string, retries int, retryInterval time.Duration, maxFileSize int64) *Worker {
	registryClient, err := regConn.NewClient()
```

There seem to be a few immediate ways that we could modify the HTTP client if that helps.
https://pkg.go.dev/github.com/docker/docker/client

mmlb (approved):

nshalman:
@tobert do you have time / want to resolve the merge conflicts here or should we close this and revisit after the refactor?

tobert:
Looked easy but too many layers of conflicts from all the linting. Closing.

PR #538, opened 2021-09-03
Introduce more stringent golangci-lint checks & address issues raised

tstromberg:
## Description

This introduces a more stringent `golangci-lint` configuration from the `lint-install` tool, as well as adds some Makefile rules to run lint locally in a way that matches CI.

## Detailed Changes

This PR includes fixes for the following lint changes, as well as some basic reformatting (`gofumpt` for Go, and `prettier` for YAML). This PR does not enforce the Dockerfile lint checks yet, but does warn about them. 

This PR does add some `//nolint` directives for issues that require substantial future refactoring work.

```
client/main.go:20:16: Comment should end in a period (godot)
// gRPC clients
               ^
client/main.go:27:79: Comment should end in a period (godot)
// FullClient aggregates all the gRPC clients available from Tinkerbell Server
                                                                              ^
client/main.go:56:67: Comment should end in a period (godot)
// instead of NewFullClientFromGlobal that will be deprecated soon
                                                                  ^
client/main.go:101:50: Comment should end in a period (godot)
// GetConnection returns a gRPC client connection
                                                 ^
client/main.go:140:41: Comment should end in a period (godot)
// Setup : create a connection to server
                                        ^
client/main.go:152:52: Comment should end in a period (godot)
// TinkHardwareClient creates a new hardware client
                                                   ^
client/main.go:161:52: Comment should end in a period (godot)
// TinkWorkflowClient creates a new workflow client
                                                   ^
cmd/tink-cli/cmd/completion.go:11:45: Comment should end in a period (godot)
// bash or zsh completion script for the CLI
                                            ^
cmd/tink-cli/cmd/completion.go:12:20: unused-parameter: parameter 'name' seems to be unused, consider removing or renaming it as _ (revive)
func completionCmd(name string) *cobra.Command {
                   ^
cmd/tink-cli/cmd/delete/delete.go:1:9: package name delete has same name as predeclared identifier (predeclared)
package delete
        ^
cmd/tink-cli/cmd/delete/delete.go:101:23: error-strings: error strings should not be capitalized or end with punctuation or a newline (revive)
				return errors.New("DeleteByID is not implemented for this resource yet. Please have a look at the issue in GitHub or open a new one.")
				                  ^
cmd/tink-cli/cmd/delete/delete.go:109:13: superfluous-else: if block ends with a continue statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary) (revive)
					} else {
						return err
					}
[...]

PR #546, opened 2021-10-01
Increase timeout for golangci-lint and allow parallel runners

detiber:
## Description

Increases the timeout for golangci-lint and configures the use of parallel runners

## Why is this needed

The github action that pushes the latest images was being flaky and would sometimes hit the default timeout of 1m for golangci-lint when running, causing the images to not be published.

This only affects CI and manual runs of golangci-lint, it does not affect end users

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,3 @@
+run:
+  timeout: 2m
+  allow-parallel-runners: true


tstromberg (approved):

tstromberg (commented):
The timeout portion of this PR was obsoleted by https://github.com/tinkerbell/tink/pull/538

I'm curious on the parallel-runners item though: are you certain it's helpful for Github Actions? This feature allows golangci-lint to be run multiple times simultaneously on the same directory & namespace, which shouldn't happen in CI.

I've previously encountered bad behavior with this option and IDE's (at least VS Code), where they end up flooding the process space with `golangci-lint` processes. Hopefully that's been fixed.

detiber:
@tstromberg sounds like things are covered, thanks!

PR #541, opened 2021-09-13
Add functionality for users to retrieve template by Name as well as UUID

Belchy06:
Signed-off-by: William Belcher <william.belcher@hotmail.com>

## Description

This PR adds the ability for Tink to retrieve workflows by either UUID or by Name. If the value after `tink template get` is not a valid UUID, Tink will default to searching for the value in the Name field of the templates. 

This PR also adds a TemplateClient to the Tink client, allowing other services such as Boots to retrieve templates.

## Why is this needed

This feature is required for the project I am working on but I am sure there are many other use cases I'm unaware of. Also, https://github.com/tinkerbell/boots/issues/178

## How Has This Been Tested?
This has been tested in the VM's based on the [Local Setup with Vagrant](https://docs.tinkerbell.org/setup/local-vagrant/) tutorial, as well as in our personal deployment

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #547, opened 2021-10-04
Fix all the dynamically generated code

mmlb:
Lots of fixes to stuff that is dynamically generated along with a few extra bits here and there in related code to make things better.

- [x] Fixes the `go generate` calls
- [x] Changes protobuf/grpc stuff to use buf for both dependency management and building/calling protoc
- [x] Changes from goimports to gofumpt as we want this anyway
- [x] Build the tools in tools.go via the Makefile
- [x] Dropped protoc-gen-doc support as that seemed unused. (easy to bring back when we have a use for it)
- [x] Add a make target to check that the generated files are up to date
- [x] Prettier-ify the GHA yaml files too (not just *.yml files)
- [ ] Update README/docs
- [ ] Setup CI to ensure files are up to date

Mostly fixes for dynamically generated files.

I noticed that #541 had some protobuf format commits which I was not expecting being done separately. I went looking into it and noticed the mess we're in.

## How Has This Been Tested?

CI

Will have better automation for messing with generated files.

## Checklist:

I have:

- [ ] added unit or e2e tests


.github/workflows/ci.yaml
```diff
@@ -15,8 +15,19 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: "1.17.0"
-      - name: make lint
-        run: make lint
+      - name: make verify
+        run: make verify
```
mmlb, line 19: Hey @tstromberg I modified this so it checks more than just lint steps. I think its in line with the idea behind lint-install's `lint`.

Makefile
```diff
@@ -1,6 +1,10 @@
+help: ## Print this help
+	@grep --no-filename -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sed 's/:.*##/·/' | sort | column -ts '·' -c 120
+
 all: cli server worker ## Build all binaries for host OS and CPU
 
 -include rules.mk
+-include lint.mk
```
mmlb, line 7: I moved the `lint-install` managed stuff here as I'd like to keep the Makefile as simplish as possible... lesson learned from osie's ;)

rules.mk
```diff
@@ -37,25 +37,53 @@ crossbinaries := $(crossbinaries:=386) $(crossbinaries:=amd64) $(crossbinaries:=
...
-	go generate ./protos/**/*
-	goimports -w ./protos/**/mock.go
-
-grpc/gen_doc:
-	protoc \
-		-I./protos \
-		-I./protos/third_party \
-		--doc_out=./doc \
-		--doc_opt=html,index.html \
-		protos/hardware/*.proto protos/template/*.proto protos/workflow/*.proto
```
mmlb, line 61: @displague I got rid of this because I couldn't find where it used. I don't think it is right? I'd love to use it but am not sure how we'd integrate the generated files (html or even markdown...)

.github/workflows/ci.yaml
```diff
@@ -15,8 +15,19 @@ jobs:
...
+  gofumpt:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: "1.14.6"
+      - name: gofumpt
```
tstromberg, line 29: The golangci-lint config already includes gofumpt.

PR #542, opened 2021-09-29
Added Kubernetes API types

micahhausler:
Signed-off-by: Micah Hausler <mhausler@amazon.com>

## Description

This is an initial commit to support the Kubernetes resource model from tinkerbell/proposals#46. I'm starting out with just the API schema, and there are a lot of fields on the Hardware type that I do not understand. I'd like to get some consensus on the API shape and fields, but this is iterative, and structures can be modified in later PRs.

After this is merged, I'll have additional follow on PRs for
* Tink Controller
* Validating webhook
* DB/GRPC server changes behind a feature flag
* Migration tooling from Postgres

## How Has This Been Tested?

* Added CI test for CRD generation

## How are existing users impacted? What migration steps/scripts do we need?

No user change required at this point

## Checklist:

I have:

- [x] updated the documentation and/or roadmap (if required)
- [x] added unit or e2e tests
- [x] provided instructions on how to upgrade

PR #563, opened 2021-12-02
Added Kubernetes Controller

micahhausler:

## Description

This PR adds a new Kubernetes controller binary for the types merged in #542.

* Adds a new `cmd/tink-controller` package with a Dockerfile and a `main.go`
* Removes the `status.data` field  on Workflows. This was just a string of the interpolated template, and was redundant given the strongly typed template actions already on the `.status` of Workflows.
* Adds tests to the methods on the workflow API structure 
* Adds a new `pkg/controllers` package. This introduces a `Manager` abstraction that wraps `sigs.k8s.io/controller-runtime`'s [`Reconciler`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.10.1/pkg/reconcile?utm_source=gopls#Reconciler) type
  * Controllers for individual types (workflows, templates, hardware) will reside in individual sub-packages
  * Controllers patch incoming objects even in the event of an error: this is intentional as a controller in the [Kubernetes Resource Model](https://github.com/kubernetes/design-proposals-archive/blob/main/architecture/resource-management.md) attempts to converge on desired state
* Adds a new `pkg/conversion` package for converting protobuf types to Kubernetes CRD types and vice versa
* Adds an internal `pkg/internal/tests` package with internal testing utilities. 
  * The first utility added is a `FrozenTime` interface with Before/After connivence methods for `time.Time`, as well as protobuf and K8s metav1 package time formats. 

## How Has This Been Tested?

* I added unit tests for nearly all new branching logic
* I added a CI target for building the tink-controller 

There is no impact to existing users. Nothing form this change is called in any existing code path.

## Checklist:

I have:

- [ ] ~~updated the documentation and/or roadmap (if required)~~
- [x] added unit or e2e tests
- [ ] ~~provided instructions on how to upgrade~~


pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
mmlb, line 1: how was this file generated?

mmlb (commented):
I only just skimmed the layout and the autogenerated file question popped up and I wanted to ask that before diving deeper here.

micahhausler:
> This is really great. I'm very excited about this, thank you! There is quite a bit of new code going in. Is it too early for docs? Would be really helpful to know about things like Kubernetes client/server version requirements, how the controller pattern works in general, how the controller code here is laid out, etc.

Thank you! I did want to hold on docs initially, since there is nothing to write about other than "there is a controller you can run that nothing else interacts with yet."

I'll add docs in the next PR, that includes changes to the Tink API.

pkg/apis/core/v1alpha1/zz_generated.deepcopy.go
micahhausler: `make genereate-go` I believe

https://github.com/tinkerbell/tink/blob/main/kube.mk#L25-L26

mmlb: Ah I see. I also just noticed that it says so right in the file https://github.com/tinkerbell/tink/pull/563/files#diff-228d0bde90a51d1ee8bc16fa0c9003354a52185a5898d6e6f47fcf1ba426221bR20 :smile: . I've opened up https://github.com/tinkerbell/tink/pull/564 to rework kube.mk to be simpler imo, can you take a look please?

cmd/tink-controller/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM alpine:3.15
+ENTRYPOINT ["/usr/bin/tink-controller"]
+EXPOSE 42113
+EXPOSE 42114
+
+ARG TARGETARCH
+ARG TARGETVARIANT
+
+RUN apk add --update ca-certificates && \
+    apk add --repository=http://dl-cdn.alpinelinux.org/alpine/edge/testing cfssl
```
mmlb, line 10: do we need cfssl in `tink-controller`? I'm not seeing where its used in this PR.

micahhausler:
I call BS on CodeCov saying we're 2% lower in coverage. I know there are specific tests and cases I added that cover files CC is showing as uncovered.

pkg/apis/core/v1alpha1/hardware_methods.go
```diff
@@ -0,0 +1,25 @@
...
+	return annotations[HardwareIDAnnotation]
+}
+
+// SetTinkID sets the Tinkerbell ID associated with this Hardware.
+func (t *Hardware) SetTinkID(id string) {
+	if t.GetAnnotations() == nil {
+		t.SetAnnotations(make(map[string]string))
+	}
+	t.Annotations[HardwareIDAnnotation] = id
+}
```
mmlb, line 25: Seeing how Annotations is public why not just:

```suggestion
// TinkID returns the Tinkerbell ID associated with this Hardware.
func (t *Hardware) TinkID() string {
	return t.Annotations[HardwareIDAnnotation]
}

// SetTinkID sets the Tinkerbell ID associated with this Hardware.
func (t *Hardware) SetTinkID(id string) {
	if t.Annotations == nil {
		t.Annotations = make(map[string]string)
	}
	t.Annotations[HardwareIDAnnotation] = id
}
```

Alternatively if you want to use the Get/Set methods something like in Set* methods:
```
annotations := t.GetAnnotations()
if annotations == nil {
	annotations = make(map[string]string)
}
annotations[HardwareIDAnnotations] = id
t.SetAnnotations(annotations)
```

That way code isn't mixing and matching use of both `Get`/`Set` methods *and* direct access to the underlying map. Best to pick just one way imo (I prefer direct access as in the first)

mmlb:
> I call BS on CodeCov saying we're 2% lower in coverage. I know there are specific tests and cases I added that cover files CC is showing as uncovered.

codecov is hard to understand some (most?) times. Don't sweat it too much imo.

cmd/tink-controller/Dockerfile
micahhausler: Good call, probably not.  Just copy/pasted from elsewhere

pkg/apis/core/v1alpha1/hardware_methods.go
micahhausler: No objection to this, just copied from the CAPT provider. I'll update

mmlb: thanks, the original way is super weird because not only is it mixing `Get`/`Set` vs direct use but also `len(map...)` vs `map == nil` checks.

micahhausler: On the `TinkID()` case, the map can be `nil`, so we need to check if it exists

mmlb: Got a link that I might be able to follow to see about submitting a PR for doing it just one way?

mmlb: > On the `TinkID()` case, the map can be `nil`, so we need to check if it exists

Actually no, go does the right thing if its nil: https://go.dev/play/p/NDqUXBkSFnd

mmlb: I couldn't remember if that was the case so went and checked earlier.

pkg/conversion/workflow.go
```diff
@@ -0,0 +1,129 @@
...
+					for k, v := range env {
+						merged[k] = v
+					}
+					for k, v := range action.Environment {
+						merged[k] = v
+					}
+					for k, v := range merged {
+						resp = append(resp, fmt.Sprintf("%s=%s", k, v))
+					}
+					sort.Strings(resp)
```
mmlb, line 96: Is this required or just your pref? Either way :chef_kiss: I approve :D

pkg/internal/tests/frozen_time.go
```diff
@@ -0,0 +1,63 @@
...
+
+type ft struct {
+	t time.Time
+}
+
+func (f *ft) Now() time.Time                   { return f.t }
+func (f *ft) Before(d time.Duration) time.Time { return f.Now().Add(d) }
+func (f *ft) After(d time.Duration) time.Time  { return f.Now().Add(-d) }
+func (f *ft) AfterSec(s int64) time.Time       { return f.Now().Add(time.Duration(s) * time.Second) }
+func (f *ft) BeforeSec(s int64) time.Time      { return f.Now().Add(time.Duration(-s) * time.Second) }
```
mmlb, line 51: nit: can you sort these like the other groups? After, Before, AfterSec, BeforeSec... for consistency

workflow/template_validator.go
```diff
@@ -60,36 +60,48 @@ func MustParseFromFile(path string) *Workflow {
...
 	}
 
+	_, buf, err := RenderTemplateHardware(templateID, templateData, hardware)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+// RenderTemplateHardware renderst the workflow template and returns the Workflow andzzzz the interpolated bytes.
```
mmlb, line 79: andzzzz?

mmlb (commented):
here's what I've seen so far. Next up is reviewing the tests. About the import blocks, you can automatically fix everthing with:

`sed -i '/^import (/,/^)/ { /^\s*$/ d}' $(git ls-files '*.go') && gofumpt -w -s .`

I've been meaning to add this to ci-checks.sh but I thought `gofumpt` did that itself :(

workflow/template_validator.go
micahhausler: lol holding z is my layer change key on my Ergodox

pkg/conversion/workflow.go
micahhausler: Not strictly required, but makes testing _way_ easier

pkg/conversion/template.go
```diff
@@ -0,0 +1,64 @@
+package conversion
```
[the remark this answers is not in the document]
micahhausler: +100

pkg/controllers/manager.go
```diff
@@ -0,0 +1,136 @@
...
+		HealthProbeBindAddress:  fmt.Sprintf(":%d", options.HealthProbePort),
+	}
+}
+
+// GetServerOptions returns a set of options used by the Tink API.
+// These options include leader election enabled.
+func GetServerOptions() controllerruntime.Options {
+	return controllerruntime.Options{
+		Logger:                 zapr.NewLogger(logging.FromContext(context.Background()).Desugar()),
+		LeaderElection:         false,
```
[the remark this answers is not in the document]
micahhausler: Ah will fix the doc. The idea here is when you only want one of many async controllers doing work/making mutations. If you have two API servers, they can both make mutations without needing a lock.

pkg/controllers/workflow/controller.go
```diff
@@ -0,0 +1,120 @@
...
+	if c.nowFunc().After(stored.GetStartTime().Add(time.Duration(stored.Status.GlobalTimeout) * time.Second)) {
+		stored.Status.State = protoworkflow.State_name[int32(protoworkflow.State_STATE_TIMEOUT)]
+	}
+
+	// check for any running actions that may have timed out
+	for ti, task := range stored.Status.Tasks {
+		for ai, action := range task.Actions {
+			// A running workflow task action has timed out
+			if action.Status == protoworkflow.State_name[int32(protoworkflow.State_STATE_RUNNING)] &&
+				action.StartedAt != nil &&
```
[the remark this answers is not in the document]
micahhausler: Actually... state changes to "RUNNING" will always be processed by the (as yet unwritten) API. I think the API or Validating Webhook should be responsible for state validation rather than the controller. The controller's job is just to try and reconcile state, not perform all validation

pkg/controllers/manager.go
```diff
@@ -0,0 +1,136 @@
...
+		field        string
+		extractValue client.IndexerFunc
+	}{
+		{
+			&v1alpha1.Workflow{},
+			WorkerAddr,
+			wokerIndexFunc,
+		},
+	}
+	for _, indexer := range indexers {
```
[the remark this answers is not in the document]
micahhausler: I anticipate multiple indexers in successive changes. This just sets it up to add more later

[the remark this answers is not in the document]
mmlb: :+1: I suspected this would be the case.

pkg/controllers/workflow/controller.go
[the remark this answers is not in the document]
mmlb: sgtm :+1:

pkg/internal/tests/frozen_time.go
micahhausler: Sorted in the order mentioned in the interface

pkg/controllers/workflow/controller_test.go
```diff
@@ -0,0 +1,709 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+var runtimescheme = runtime.NewScheme()
+
+var TestTime = tests.NewFrozenTimeUnix(1637361793)
```
mmlb, line 22: maybe a comment for this magic time?

pkg/controllers/workflow/controller_test.go
```diff
@@ -0,0 +1,709 @@
...
+					t.Errorf(`Got unexpected error: got "%v" wanted "%v"`, gotErr, tc.wantErr)
+				}
+				return
+			}
+			if gotErr == nil && tc.wantErr != nil {
+				t.Errorf("Missing expected error: %v", tc.wantErr)
+				return
+			}
+			if tc.want != got {
+				t.Errorf("Got unexpected result. Wanted %v, got %v", tc.want, got)
```
mmlb, line 692: why no return? Also, why not just `t.FailNow()` which if I read the docs right will stop this test but continue others.

mmlb: Yep, works: https://go.dev/play/p/m3xCy3B-onA

pkg/convert/template_test.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
+	"github.com/tinkerbell/tink/pkg/internal/tests"
+	prototemplate "github.com/tinkerbell/tink/protos/template"
+	"google.golang.org/protobuf/testing/protocmp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var TestTime = tests.NewFrozenTimeUnix(1637361794)
```
mmlb, line 14: same question, what is this magic time and maybe lets comment it?

pkg/convert/workflow.go
```diff
@@ -0,0 +1,129 @@
...
+
+	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
+	protoworkflow "github.com/tinkerbell/tink/protos/workflow"
+	"github.com/tinkerbell/tink/workflow"
+)
+
+func WorkflowYamlToCRD(wf *workflow.Workflow, crd *v1alpha1.Workflow) {
```
mmlb, line 15:
```suggestion
func WorkflowYAMLToCRD(wf *workflow.Workflow, crd *v1alpha1.Workflow) {
```
So all the acronyms are in all caps please.

workflow/template_validator.go
```diff
@@ -60,36 +60,48 @@ func MustParseFromFile(path string) *Workflow {
...
 	}
 
+	_, buf, err := RenderTemplateHardware(templateID, templateData, hardware)
+	if err != nil {
+		return "", err
+	}
+	return buf.String(), nil
+}
+
+// RenderTemplateHardware renderst the workflow template and returns the Workflow and the interpolated bytes.
```
mmlb, line 79:
```suggestion
// RenderTemplateHardware renders the workflow template and returns the Workflow and the interpolated bytes.
```

mmlb:
Also needs a rebase

pkg/controllers/workflow/controller_test.go
micahhausler: We also want to compare the mutated object, not just the controller response

micahhausler:
@mmlb I removed go from the Nix install, it was using a different version than what GH Actions uses and caused CI failures

pkg/convert/workflow.go
```diff
@@ -0,0 +1,128 @@
...
+	"sort"
+
+	"github.com/tinkerbell/tink/pkg/apis/core/v1alpha1"
+	protoworkflow "github.com/tinkerbell/tink/protos/workflow"
+	"github.com/tinkerbell/tink/workflow"
+	timestamppb "google.golang.org/protobuf/types/known/timestamppb"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func WorkflowYAMLToCRD(wf *workflow.Workflow, crd *v1alpha1.Workflow) {
```
displague, line 14: The signature of this function is different from the other convert functions which _return_ the target type.

I see that `WorkflowYAMLToCRD` is being used to update the Status of existing CRD objects elsewhere.
Perhaps `WorkflowYAMLToStatus` would be more succinct? I doubt you would want to copy over a new Status, so I imagine you will want to add another function that merges these known Status changes with an existing CRD (or Status) object.

Changing the signature to `func WorkflowYAMLToStatus(wf *workflow.Workflow)  *v1alpha1.WorkflowStatus` would also get past my other gripe with this function, that a `nil` `crd` value is ignored and returned without logging/erroring.

micahhausler: I'm good to change this to `func WorkflowYAMLToStatus(wf *workflow.Workflow) *v1alpha1.WorkflowStatus`

displague (dismissed):

micahhausler:
@mmlb anything else you'd like to see here?

mmlb:
> @mmlb anything else you'd like to see here?

aahhg, sorry for dropping the ball on this. I think I was pretty happy with everything except for dropping go from shell.nix. I'll look again on Tuesday though if thats ok? If its just the go in shell.nix thing then we can merge and I'll sync up with you on why you needed to drop it so we can add it back in. Is that alright?

micahhausler:
> I think I was pretty happy with everything except for dropping go from shell.nix.

This is due to newer point releases of Go in the GH actions than nix has available. We set the Go version in GH actions to `1.17` in #565 which auto updated for two December point releases (including CVEs in 1.17.5) and one in Jan, but the latest Go 1.17 Nix commit is 2 months old pinning to `1.17.3` ([32ef30c](32ef30c)). This resulted in CI build failures when some tooling was built with 1.17.5 and Nix was 1.17.3.

> I'll look again on Tuesday though if thats ok? 

SGTM!

> If its just the go in shell.nix thing then we can merge and I'll sync up with you on why you needed to drop it so we can add it back in. Is that alright?

SGTM!

pkg/convert/workflow.go
```diff
@@ -0,0 +1,129 @@
...
+			resp.ActionList = append(resp.ActionList, &protoworkflow.WorkflowAction{
+				TaskName: task.Name,
+				Name:     action.Name,
+				Image:    action.Image,
+				Timeout:  action.Timeout,
+				Command:  action.Command,
+				WorkerId: task.WorkerAddr,
+				Volumes:  append(task.Volumes, action.Volumes...),
+				// TODO: (micahhausler) Dedupe task volume targets overridden in the action volumes?
+				//   Also not sure how Docker handles nested mounts (ex: "/foo:/foo" and "/bar:/foo/bar")
```
mmlb, line 83: fyi:
```
[manny@zennix:~/t/debug]$ tree
.
├── bar
│   └── from-bar
└── foo
    └── from-foo

2 directories, 2 files

[manny@zennix:~/t/debug]$ docker run --rm -ti -v $PWD/foo:/foo -v $PWD/bar:/foo/bar alpine
/ # echo ==foo==; ls -l /foo/; echo ==bar==; ls -l /foo/bar; 
==foo==
total 33
drwxr-xr-x    2 1000     users            3 Jan 18 20:23 bar
-rw-r--r--    1 1000     users            0 Jan 18 20:23 from-foo
==bar==
total 1
-rw-r--r--    1 1000     users            0 Jan 18 20:23 from-bar
/ # 

[manny@zennix:~/t/debug]$ tree
.
├── bar
│   └── from-bar
└── foo
    ├── bar
    └── from-foo

3 directories, 2 files
```
note how `foo/bar` directory now exists

mmlb (approved):

mmlb:
Looks like mergify is getting stuck on `docker-images`, possibly due to the check being set up as a matrix job. I've opened up a chat with them to see what the fix/work-around is. I'd like to hold this here until tomorrow at most.

mmlb (approved):

micahhausler:
🎉

PR #564, opened 2021-12-03
Add controller-gen to tools.go

mmlb:
## Description

Modifies most of kube.mk to be better integrated with tools.go and accompanying make rules.

## Why is this needed

Gets rid of recursive make calls (which are considered harmful :smile:) by giving make more/better info.
Also removes code that is no longer necessary (and some that never was, envsubst in kube.mk).

## How Has This Been Tested?

I ran `make generate`


kube.mk
```diff
@@ -1,36 +1,18 @@
...
-	$(CONTROLLER_GEN) \
+generate-manifests: bin/controller-gen # Generate manifests e.g. CRD, RBAC etc.
+	controller-gen \
 		paths=./pkg/apis/... \
 		crd:crdVersions=v1 \
 		rbac:roleName=manager-role \
 		output:crd:dir=./config/crd/bases \
 		output:webhook:dir=./config/webhook \
 		webhook
+	prettier --write ./config/crd/
```
micahhausler, line 18: 👍

micahhausler (approved):
LGTM, thanks for the cleanup. I mostly just copied what kubebuilder dumps out and what was in CAPT

kube.mk
mmlb: :nod: but now I have to fix ci.

mmlb: should be fixed now

micahhausler (approved):
LGTM

micahhausler (approved):
LGTM

ScottGarman (approved):

PR #565, opened 2021-12-03
Updated repository config files

micahhausler:

## Description

Extracted repo config changes from #563

## Why is this needed

* Go version is out of date in `ci.yaml`
* `.mk` files were missing from editor config
* Coverage was counted against for generated files

Fixes: #

N/A

## How Has This Been Tested?

Ran verify/tests locally

N/A

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: "1.17.0"
+          go-version: "1.17"
       - name: make verify
         run: make verify
   test:
@@ -25,11 +25,13 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: "1.14.6"
+          go-version: "1.17"
       - name: Generate
         run: make generate
+      - name: go clean -testcache
+        run: go clean -testcache
       - name: go test
-        run: go test -coverprofile=coverage.txt ./...
+        run: go test -coverprofile=coverage.txt ./... -v
       - name: upload codecov
         run: bash <(curl -s https://codecov.io/bash)
   ci-checks:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -25,11 +25,13 @@ jobs:
...
         with:
-          go-version: "1.14.6"
+          go-version: "1.17"
       - name: Generate
         run: make generate
+      - name: go clean -testcache
+        run: go clean -testcache
       - name: go test
-        run: go test -coverprofile=coverage.txt ./...
+        run: go test -coverprofile=coverage.txt ./... -v
```
mmlb, line 34: can you drop this one? I'm not sure we really need it. I don't really care to see the test names when there's no failure (I don't even look at the test run if there's no failure tbh). Go does the right thing when there's failure and only expands the failing tests. And actually having `-v` in there just adds a ton of noise when there is a test, printing out the names of all the passing tests.

.github/workflows/ci.yaml
```diff
@@ -25,11 +25,13 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: "1.14.6"
+          go-version: "1.17"
       - name: Generate
         run: make generate
+      - name: go clean -testcache
+        run: go clean -testcache
```
mmlb, line 32: Do we really need this? If so why?

micahhausler: sure, It was in the Makefile, so I added for consistency

.github/workflows/ci.yaml
micahhausler: sure, It was in the Makefile, so I added for consistency

mmlb: ok makes sense. consistency > sense :D

.github/workflows/ci.yaml
mmlb: ok makes sense. consistency > sense :D

mmlb (approved):

PR #567, opened 2021-12-06
Tinkerbell GRPC Insecure Mode

nshalman:
## Description

This PR, along with some minor cleanup, enables tink-server to run without a TLS certificate and enables tink-cli to connect to it without needing TLS to be set up.

At the time of opening the PR, this change is intended to only extend the capabilities of tink-server and tink-cli rather than fundamentally altering them, though I am open to the latter before this lands if that's the right answer.

## Why is this needed

This will simplify development and testing of the tinkerbell stack, and could theoretically be used in the sandbox.
Alternately, the sandbox could include TLS termination similar to how many applications use Kubernetes ingress for TLS termination.

## How Has This Been Tested?

Template creation and retrieval
```
cat focal.yaml | docker exec -i tink_tink-cli_1 tink template create
docker exec -i tink_tink-cli_1 tink template get
```
- [x] against `make run-stack`
- [ ] with the sandbox

This is intended to not impact 

## Checklist:

I have:

- [ ] added unit or e2e tests


cmd/tink-server/main.go
```diff
@@ -93,6 +95,11 @@ func (c *DaemonConfig) PopulateFromLegacyEnvVar() {
 	if basicAuthPass := os.Getenv("TINK_AUTH_PASSWORD"); basicAuthPass != "" {
 		c.HTTPBasicAuthPassword = basicAuthPass
 	}
+	if insecure, isSet := os.LookupEnv("INSECURE"); isSet {
```
stephen-fox, line 98: I think we should use `TINKERBELL_INSECURE` for consistency, and to avoid possibly conflicting with another application.

client/main.go
```diff
@@ -90,8 +71,23 @@ func NewClientConn(opt *ConnOptions) (*grpc.ClientConn, error) {
...
+		creds, err := fetchCert(opt.CertURL)
+		if err != nil {
+			return nil, err
+		}
+		method = grpc.WithTransportCredentials(creds)
+	}
+	conn, err := grpc.Dial(opt.GRPCAuthority,
+		method,
+		grpc.WithUnaryInterceptor(otelgrpc.UnaryClientInterceptor()),
+		grpc.WithStreamInterceptor(otelgrpc.StreamClientInterceptor()),
```
tobert, line 89: excellent :)

Might want to add a client option to print the trace id in this case e.g. in the logs, so it's easy to take to a search box.

cmd/tink-server/main.go
```diff
@@ -93,6 +95,11 @@ func (c *DaemonConfig) PopulateFromLegacyEnvVar() {
...
+		if b, err := strconv.ParseBool(insecure); err != nil {
```
stephen-fox, line 99: Should this be `err == nil`?

Or, should this result in a panic?

You could also ignore the error I suppose, as that will result in what I consider to be a decent default: that a TLS listener gets created.

client/main.go
```diff
@@ -100,34 +96,27 @@ func NewClientConn(opt *ConnOptions) (*grpc.ClientConn, error) {
...
 	grpcAuthority := os.Getenv("TINKERBELL_GRPC_AUTHORITY")
 	if grpcAuthority == "" {
 		return nil, errors.New("undefined TINKERBELL_GRPC_AUTHORITY")
 	}
-	creds := credentials.NewClientTLSFromCert(cp, "")
+
+	method := grpc.WithInsecure()
+	insecure := os.Getenv("TINKERBELL_INSECURE")
+
+	if insecure == "" {
```
tobert, line 107: no big deal but I usually make flags like `insecure` here go through strconv.ParseBool, technically safer than allowing even " " to get interpreted as true

cmd/tink-server/main.go
```diff
@@ -180,22 +187,30 @@ func NewRootCommand(config *DaemonConfig, logger log.Logger) *cobra.Command {
...
-				ModTime:               modT,
-				GRPCAuthority:         config.GRPCAuthority,
-				HTTPAuthority:         config.HTTPAuthority,
-				HTTPBasicAuthUsername: config.HTTPBasicAuthUsername,
-				HTTPBasicAuthPassword: config.HTTPBasicAuthPassword,
-			}, errCh)
+			if config.Insecure {
+				rpcServer.SetupGRPC(ctx, logger, &rpcServer.ConfigGRPCServer{
+					Facility:      config.Facility,
+					TLSCert:       "insecure",
```
stephen-fox, line 193: Can we leave this field as unspecified / default value (empty string)?

client/main.go
tobert: this also implies `TINKERBELL_INSECURE=0` means to run in insecure mode...

nshalman, line 107: Could probably use a comment here explaining that `insecure == ""` means that we didn't get anything in the environment variable and are thus using the original codepath that *does* use TLS.

docker-compose-insecure.yaml
```diff
@@ -0,0 +1,88 @@
...
+    ports:
+      - 42113:42113/tcp
+      - 42114:42114/tcp
+
+  tink-server-migration:
+    image: quay.io/tinkerbell/tink:latest
+    restart: on-failure
+    environment:
+      ONLY_MIGRATION: "true"
+      INSECURE: "true"
```
stephen-fox, line 39: This will need to be updated if we renamed the environment variable.

client/main.go
```diff
@@ -65,15 +44,17 @@ func NewFullClient(conn grpc.ClientConnInterface) *FullClient {
...
 
 func (o *ConnOptions) SetFlags(flagSet *pflag.FlagSet) {
 	flagSet.StringVar(&o.CertURL, "tinkerbell-cert-url", "http://127.0.0.1:42114/cert", "The URL where the certificate is located")
 	flagSet.StringVar(&o.GRPCAuthority, "tinkerbell-grpc-authority", "127.0.0.1:42113", "Link to tink-server grcp api")
+	flagSet.BoolVar(&o.Insecure, "insecure", false, "Run in insecure mode (no TLS)")
 }
 
-func NewClientConn(opt *ConnOptions) (*grpc.ClientConn, error) {
-	resp, err := http.Get(opt.CertURL)
+func fetchCert(url string) (credentials.TransportCredentials, error) {
```
stephen-fox, line 56: I think the function name could be a bit more descriptive. Maybe `grpcCredentialFromCertEndpoint`, or `trustedCertsFromCertEndpoint`.

Either that, or a comment would be helpful.

client/main.go
```diff
@@ -90,8 +71,23 @@ func NewClientConn(opt *ConnOptions) (*grpc.ClientConn, error) {
 		return nil, errors.Wrap(err, "parse cert")
```
stephen-fox, line 71: I know we didn't create this issue, but I just realized that `err` here is `nil`.

An improvement would be decoding the certificate's PEM and parsing the resulting ASN.1 DER using the `x509` library. That, or creating a new error that says something like `failed to parse cert - cert pool append certs from pem failed`.

client/main.go
nshalman: Amy's idea above is even better.

grpc-server/grpc_server.go
```diff
@@ -60,14 +60,19 @@ func SetupGRPC(ctx context.Context, logger log.Logger, config *ConfigGRPCServer,
 		dbReady: true,
 		logger:  logger,
 	}
-	if cert := config.TLSCert; cert != "" {
```
stephen-fox, line 63: Could we add a field to `ConfigGRPCServer` that specified no TLS?

The original `if` statement might work as-is... But it does leave some ambiguity about what should happen if the field is an empty string.

stephen-fox (changes requested):

client/main.go
nshalman: Do you have an example elsewhere of how to do that I can crib from?

cmd/tink-server/main.go
nshalman: Good catch. Thanks!

client/main.go
nshalman: agreed

client/main.go
micahhausler: +1 to being more explicit.

client/main.go
```diff
@@ -100,34 +96,27 @@ func NewClientConn(opt *ConnOptions) (*grpc.ClientConn, error) {
...
-	}
-
 	grpcAuthority := os.Getenv("TINKERBELL_GRPC_AUTHORITY")
 	if grpcAuthority == "" {
 		return nil, errors.New("undefined TINKERBELL_GRPC_AUTHORITY")
 	}
-	creds := credentials.NewClientTLSFromCert(cp, "")
+
+	method := grpc.WithInsecure()
+	insecure := os.Getenv("TINKERBELL_INSECURE")
```
micahhausler: Since we're using pflag above, can we use Viper to get the env var?

micahhausler (changes requested):
I'm trying to better understand the motivations for this PR. Is it that "generating TLS serving certs is hard"? Or is passing arounds certs trusted by the server is difficult? 

My read of this PR is that we want to provide Tink server operators the ability to turn off client TLS authentication on the server.  

Can the docs be updated with this new mode and the flags to turn it on?

Insecure can have multiple meanings:
* No TLS at all (client or server or both)
* Still use mTLS, but trust any TLS certificate

Can the "insecure" parameters be named more specifically? Maybe something like `insecureNoTLSAuthentication`

client/main.go
nshalman: whynotboth.gif

docker-compose-insecure.yaml
```diff
@@ -0,0 +1,88 @@
...
+      ROLLBAR_DISABLE: ${ROLLBAR_DISABLE:-1}
+      PGDATABASE: tinkerbell
+      PGHOST: db
+      PGPASSWORD: tinkerbell
+      PGPORT: 5432
+      PGSSLMODE: disable
+      PGUSER: tinkerbell
+      TINKERBELL_GRPC_AUTHORITY: :42113
+      TINKERBELL_HTTP_AUTHORITY: :42114
+      TINKERBELL_INSECURE: "true"
```
micahhausler, line 22: Do we need a full new docker-compoose? Can this not default to false and accept an override for true?

nshalman:
> My read of this PR is that we want to provide Tink server operators the ability to turn off client TLS authentication on the server.

We want to turn off TLS as the transport, largely to simplify development, but possibly also for production uses.
On the server side we don't want to use TLS.
On the client side we want the GRPC client connection to be created using the "WithInsecure" DialOption which "disables transport security" (see https://pkg.go.dev/google.golang.org/grpc#WithInsecure)

> Can the docs be updated with this new mode and the flags to turn it on?
If you mean docs within this repo, that's certainly a reasonable thing to request before this is approved, though we should probably resolve the naming issues first.
I'd rather not write docs elsewhere before this is approved, though it seems reasonable to me that they be written before it lands.

> Insecure can have multiple meanings:
> 
> * No TLS at all (client or server or both)
> * Still use mTLS, but trust any TLS certificate
> 
> Can the "insecure" parameters be named more specifically? Maybe something like `insecureNoTLSAuthentication`

I chose "insecure" because that's what GRPC calls the DialOption. I guess "insecureNoTLS" could work, but at that rate we could call it "noTLS" too.

micahhausler:
> We want to turn off TLS as the transport, largely to simplify development, but possibly also for production uses.
On the server side we don't want to use TLS.
 
Introducing a new mode where users can hurt themselves is a bit scary to me, even if it is prefixed with "insecure." Can we simplify development and deployment with certificate generating scripts/more documentation? 

> I chose "insecure" because that's what GRPC calls the DialOption. I guess "insecureNoTLS" could work, but at that rate we could call it "noTLS" too.

If we do go down this route, I'm a fan of `insecureNoTLS`/`TINKERBELL_INSECURE_NO_TLS`/`--insecure-no-tls`. If/when other security features are added (ex: authorization), you'll want to differentiate them and still make it explicit that the configuration option is insecure.

client/main.go
jacobweinstock: food for thought on naming: https://talks.golang.org/2014/names.slide#3

cmd/tink-server/main.go
splaspood: Maybe that should be TINK_INSECURE instead of TINKERBELL_?

cmd/tink-server/main.go
nshalman: I think I cribbed this from ONLY_MIGRATION above. a bug in both locations? Did any of this ever work??

nshalman: It's probably time to collapse the two different codepaths for env vars and command line flags into one piece of correct code... :sigh:

cmd/tink-server/main.go
mmlb: "insecure" is being used as a sentinel value to skip tls stuff in `rpcServer.SetupGRPC`. I'll be cleaning that up next and this will all go away then. In the meantime I don't think its worth re-working this + setupGRPC if its going to be deleted soon.

mmlb:
I'm going to take over this PR so will be dusting this off and responding to feedback.

client/main.go
mmlb: I'd agree, but I'm actually looking to drop the `/cert` completely so this will go away anyway.

docker-compose-insecure.yaml
mmlb: Agree, I'll give this a shot

grpc-server/grpc_server.go
mmlb: I'll be reworking this soon and should do a better job of dealing with tls certs. Lets ignore this for now please.

mmlb:
I have 2 primary reasons for wanting to see this PR through. First is to enable a simpler dev mode where we don't paper over the insecure setup necessary for sandbox (https://github.com/tinkerbell/sandbox/blob/9a04338168efd85be173100eb165bd2d3c6d09c2/deploy/compose/docker-compose.yml#L78). The second reason is to do away with the whole `/cert` thing. `/cert` came around from when cacher was first developed and had to deal with internal stuff. It should have never made its way into tinkerbell. The only reason it worked in cacher and sort of works in tinkerbell is by piggy-backing on tls to fetch `/cert`. We can't do that in sandbox with the self-signed certs and so the one safety anchor goes away.

IMO tinkerbell should support tls via an ingress or tls-terminating proxy (and thus runs in insecure mode itself, so the traffic is in the clear once "on-prem") or if you want to have tls for tinkerbell all the way through we should pass in valid/trusted-ca certs and do away with self-signed certs all together. The former is a nice middle ground of safe-enough for many use cases (remember we do want to be homelab-ish friendly too) and the latter could still be used to support self-signed tls anyway, its just harder. I'm ok with stating "if you already want to deal with the hard work of a self-signed pki infrastructure, then you should be ready to roll your own live environment to handle your 'untrsuted CA' (or embed the CA cert into hook.. somehow".

mmlb:
> > We want to turn off TLS as the transport, largely to simplify development, but possibly also for production uses.
> > On the server side we don't want to use TLS.
> 
> Introducing a new mode where users can hurt themselves is a bit scary to me, even if it is prefixed with "insecure." Can we simplify development and deployment with certificate generating scripts/more documentation?

This should be doable I think in sandbox code, by repackaging hook's initrd to include the generated CA's cert into the trust store, but I don't want to do it myself. I'd rather spend our time on better tinkerbell features than supporting self-signed certs _easiliy_.
 
> 
> _Edit:_ If the primary motivation is development, can we use build tags to make development easier, but release artifacts are don't use the insecure mode?

Its not just development, that was *my* primary want but after thinking some more on this I think "insecure" mode is actually secure enough for many users. Standing up a tls terminating grpc proxy that does clear-text to its backends is very common, especially in homelab/messing-around scenarios. 

> > I chose "insecure" because that's what GRPC calls the DialOption. I guess "insecureNoTLS" could work, but at that rate we could call it "noTLS" too.
> 
> If we do go down this route, I'm a fan of `insecureNoTLS`/`TINKERBELL_INSECURE_NO_TLS`/`--insecure-no-tls`. If/when other security features are added (ex: authorization), you'll want to differentiate them and still make it explicit that the configuration option is insecure.

I like this, but would rather have it be `--no-tls`. We can be secure w/o tinkerbell server running tls itself.

mmlb:
I've forced push with some fixes. I've renamed `insecure` -> `no-tls` in both code, args and env var names. I've also added some info about only recommended for dev or if behind a tls terminating proxy to hopefully avoid misuse. I've also dropped the split docker-compose.yml file since we really could just reuse the old one anyway.

mmlb:
Would love some :eyes: @micahhausler, @stephen-fox and @nshalman

mmlb:
Depends on #583

mmlb:
Alright this should be the last of the changes. I've changed the args and code from negative to positive `--no-tls/TINKERBELL_NO_TLS` (default false) -> `--tls/TINKERBELL_TLS` (default true). This avoided needing to have double negatives in the code (`if !noTLS` for example). Still depends on #583 and as such has its commits, but should be cleared up soon.

markyjackson-taulia (approved):

Issue #574, opened 2021-12-23
trivial sql injection in GetTemplate


mmlb:
Tink's template code does the unforgivable sin of concatenating strings to build up the query which leads to sql injection. hardware and workflow might be similarly affected though a tad more difficult (I think, due to json) but definitely worth looking into.

## Expected Behaviour

calls to `GetTemplate` do not modify the database.

## Current Behaviour

No validation is done to the Id or Name parameters in `GetTemplate`s arguments, which can be easily crafted to drop data.

## Possible Solution

Use postgres query parameters instead string concat.

## Steps to Reproduce (for bugs)

1. Bring up sandbox
2. Compile the example program at the bottom of this issue
3. `docker cp` the statically built program from step 2 into `tink-cli` container
4. Run said program, see simlar output:
```ShellSession
vagrant@ubuntu2004:/vagrant/compose$ docker cp ~/tink-sql-injection compose_tink-cli_1:/ && docker-compose exec tink-cli ./tink-sql-injection
===== before sql injection =====
id:89202fd1-643f-11ec-b9a0-0242ac120006
name:debian_Focal
data:version: "0.1"
name: debian_Focal
global_timeout: 1800
tasks:
  - name: "os-installation"
    worker: "{{.device_1}}"
    volumes:
      - /dev:/dev
      - /dev/console:/dev/console
      - /lib/firmware:/lib/firmware:ro
    actions:
      - name: "stream-ubuntu-image"
        image: image2disk:v1.0.0
        timeout: 600
        environment:
          DEST_DISK: /dev/vda
          IMG_URL: "http://192.168.56.4:8080/focal-server-cloudimg-amd64.raw.gz"
          COMPRESSED: true
      - name: "install-openssl"
        image: cexec:v1.0.0
        timeout: 90
        environment:
          BLOCK_DEVICE: /dev/vda1
          FS_TYPE: ext4
          CHROOT: y
          DEFAULT_INTERPRETER: "/bin/sh -c"
          CMD_LINE: "apt -y update && apt -y install openssl"
      - name: "create-user"
        image: cexec:v1.0.0
        timeout: 90
        environment:
          BLOCK_DEVICE: /dev/vda1
          FS_TYPE: ext4
          CHROOT: y
          DEFAULT_INTERPRETER: "/bin/sh -c"
          CMD_LINE: "useradd -p $(openssl passwd -1 tink) -s /bin/bash -d /home/tink/ -m -G sudo tink"
      - name: "enable-ssh"
[...]

PR #577, opened 2022-01-12
escape pgsql field and values generated by buildGetCondition

displague:
## Description

SQL column names and values should be escaped. Go does not offer a generic function for this (as it tends to be SQL dialect specific).  The "pq" library, which we are using as a Postgresql interface, does provide helpers for this purpose.
 
https://pkg.go.dev/github.com/lib/pq#QuoteIdentifier

Alternatively, `buildGetCondition` is only used by one caller. It should be possible to remove `buildGetCondition` and have `GetTemplate` build a parameterized query with a parameterized list (some SQLs or client libraries don't allow for field names to be parameters, I don't know if that is the case here).
 
## Why is this needed

Fixes: #574

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -111,7 +111,7 @@ func get(ctx context.Context, db *sql.DB, query string, args ...interface{}) (st
 func buildGetCondition(fields map[string]string) (string, error) {
 	for column, field := range fields {
 		if field != "" {
-			return fmt.Sprintf("%s = '%s'", column, field), nil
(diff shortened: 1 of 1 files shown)


mmlb (approved):
lgtm, needs a rebase though. Shall I do it?

cprivite:
@mergify rebase

cprivite:
> @cprivite is not allowed to run commands
> 
> _Hey, I reacted but my real name is @Mergifyio_

Oh well, I tried.

mmlb:
rebased on top of latest main

mmlb (approved):

PR #581, opened 2022-01-19
mergify: Change docker-images check more explicit

mmlb:
## Description

Add one explicit `check-success` condition for each of the docker-image actions.

## Why is this needed

The previous check was never matching because mergify was not being given the name of the matrix job, it is only given the realized name for each matrix instance.

We also need to spell each matrix permuation we care about instead of `check-success~=docker-image` because mergify will happily merge once the first image build succeeds, it has not way to know it needs to wait for all of them.

## How Has This Been Tested?

Verified to be valid using [mergify config editor](https://dashboard.mergify.com/github/tinkerbell/repo/tink/config-editor?repositoryName=tink&pullRequestNumber=563)

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -3,7 +3,9 @@ queue_rules:
     conditions:
       # Conditions to get out of the queue (= merged)
       - check-success=DCO
-      - check-success=docker-images
+      - check-success~=docker-images.*tink-cli
+      - check-success~=docker-images.*tink-server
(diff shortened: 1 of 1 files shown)


ScottGarman (approved):

mmlb:
Rebased on top of main after #582 was merged, PTAL again @ScottGarman

ScottGarman (approved):

splaspood (approved):

PR #583, opened 2022-01-25
Drop json rpc gateway support

mmlb:
## Description

Drop all code dealing with json rpc.

## Why is this needed

Its likely unused code which has some weird code (:eye: runtime.String) that make other PRs more tedious (#567, I want to drop the self-signed `/cert` stuff too) to implement.

## How Has This Been Tested?

`go build` and CI.

No impact I suspect (and hope). I know of only one other tink client and it is also grpc https://github.com/swills/py-tink-cli.


mmlb:
I realized you have a dummy client @micahhausler and we didn't clarify it in slack so pinging you here. I hope its grpc based :D.

micahhausler:
> I realized you have a dummy client @micahhausler and we didn't clarify it in slack so pinging you here. I hope its grpc based :D.

The [virtual worker](https://github.com/micahhausler/tink/tree/feature/k8s-mode/cmd/virtual-worker) just uses the gRPC interface like tink-worker. No risk here from me

cmd/tink-server/main.go
```diff
@@ -188,12 +188,9 @@ func NewRootCommand(config *DaemonConfig, logger log.Logger) *cobra.Command {
 			}, errCh)
 
 			httpServer.SetupHTTP(ctx, logger, &httpServer.Config{
-				CertPEM:               cert,
-				ModTime:               modT,
-				GRPCAuthority:         config.GRPCAuthority,
-				HTTPAuthority:         config.HTTPAuthority,
-				HTTPBasicAuthUsername: config.HTTPBasicAuthUsername,
-				HTTPBasicAuthPassword: config.HTTPBasicAuthPassword,
```
micahhausler, line 196: Is the basic auth only limited to json rpc gateway?

micahhausler (approved):
Yay for cutting code!

cmd/tink-server/main.go
mmlb: Yep, https://github.com/tinkerbell/tink/pull/583/files#diff-441184fd3cd157ac00ab9c45d677b03cf32b3113097d24800ea9e5ee60fa965dL82 is where the auth is used and its for stuff registered off of the grpc http mux https://github.com/tinkerbell/tink/pull/583/files#diff-441184fd3cd157ac00ab9c45d677b03cf32b3113097d24800ea9e5ee60fa965dL50 only. The other http paths are auth free.

mmlb:
which duh, means I have more code I can delete :D

cmd/tink-server/main.go
mmlb: which duh, means I have more code I can delete :D

.envrc
```diff
@@ -1,4 +1,7 @@
+has nix && use nix
+dotenv_if_exists
```
cprivite, line 2: Where is this function defined?

mmlb:
Yep it should, maybe you need to update? https://direnv.net/man/direnv-stdlib.1.html#codedotenvifexists-ltdotenvpathgtcode

cprivite (approved):

markyjackson-taulia (approved):
/lgtm
I love when we shed code so thank you for this!

PR #587, opened 2022-02-09
Fix help and usage for `tink template update`

detiber:
## Description

`tink template update` previously had the wrong help output (stating argument is `--file` instead of `--path`) and did not support providing the updated template data over stdin.

## Why is this needed

- sandbox users run commands through `docker exec`, so supporting stdin makes it easier for them to use the command
- users should expect the help output to match the expected usage.

## How Has This Been Tested?
TODO

Should not affect existing users, only fix issues users are already encountering.


mmlb (approved):
lgtm

PR #588, opened 2022-02-09
Fix issue with `tink template update` using stdin

detiber:
## Description

Fixes an issue with `tink template update` using stdin, follow up to #587 

## Why is this needed

#587 didn't fully add support for sdin with `tink template update`, it still threw an error if content wasn't provided using the `--path` argument.

## How Has This Been Tested?
Tested manually against a Tinkerbell environment created using HEAD of sandbox

Existing users not impacted


detiber:
@mmlb a quick update for #587 that I noticed after finishing up manual testing

mmlb (approved):

PR #596, opened 2022-03-14
Refactored Tink Worker

micahhausler:
## Description

* Externalized worker package for testability  
* Created `ContainerManager` and `ContainerLogger` abstractions
* Added tests for new interfaces
* Cleaned up worker method arguments (logger is now passed via context)
* Left worker logic unchanged
* Refactored `Worker` struct and initialization

## Why is this needed

I was working on #573 and wanted to be able to create a 'virtual worker' with faked-out Docker client calls. This refactor will make it easy immediately test API changes and later refactor the worker logic.

## How Has This Been Tested?

* Tested locally with a workflow
* Added unit tests

No impact to existing users

## Checklist:

I have:

- [x] added unit or e2e tests


cmd/tink-worker/worker/container_manager.go
```diff
@@ -0,0 +1,154 @@
...
+func (m *containerManager) WaitForFailedContainer(ctx context.Context, id string, failedActionStatus chan pb.State) {
+	l := m.getLogger(ctx)
+	// send API call to wait for the container completion
+	wait, errC := m.cli.ContainerWait(ctx, id, container.WaitConditionNotRunning)
+
+	select {
+	case status := <-wait:
+		if status.StatusCode == 0 {
+			failedActionStatus <- pb.State_STATE_SUCCESS
+			return
```
micahhausler, line 131: This method is only ever called in a goroutine, but if it isn't, this code will block on a non-buffered channel so I added the return here

cmd/tink-worker/worker/container_manager_test.go
```diff
@@ -0,0 +1,395 @@
...
+			registry:    "rebelba.se",
+			clientErr:   errors.New("You missed the shot"),
+			wantErr:     errors.New("DOCKER CREATE: You missed the shot"),
+		},
+	}
+	logger, err := log.Init("github.com/tinkerbell/tink")
+	if err != nil {
+		t.Errorf("Error initializing logger: %w", err)
+		return
+	}
```
mmlb, line 134: you're probably better of using https://pkg.go.dev/github.com/packethost/pkg@v0.0.0-20211110202003-387414657e83/log#Test instead, it has the benefit of only printing output when `test.v` is enabled or if the test fails.

cmd/tink-worker/worker/registry.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/docker/docker/api/types"
+	"github.com/pkg/errors"
+)
+
+// RegistryConnDetails are the connection details for accessing a Docker
+// registry.
+type RegistryConnDetails struct {
+	Registry,
+	User,
+	Pwd string
```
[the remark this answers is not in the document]
mmlb: can we also do `Username` and `Password` please, they read easier and (*bonus*) are the same length!

cmd/tink-worker/worker/worker.go
```diff
@@ -376,14 +425,15 @@ func isLastAction(wfContext *pb.WorkflowContext, actions *pb.WorkflowActionList)
...
-	l := w.logger.With("workflowID", actionStatus.GetWorkflowId,
+	l := w.getLogger(ctx).With("workflowID", actionStatus.GetWorkflowId(),
 		"workerID", actionStatus.GetWorkerId(),
 		"actionName", actionStatus.GetActionName(),
 		"taskName", actionStatus.GetTaskName(),
 	)
 	var err error
 	for r := 1; r <= w.retries; r++ {
-		_, err = w.client.ReportActionStatus(ctx, actionStatus)
+		l.Info("Reporting Action Status")
```
mmlb, line 435: log messages should not be `Sentence case`. I thought we had a linter to catch this :thinking: (maybe needs a lint-install update). Would be nice not to introduce cases where we'd need to fix later on.

mmlb (commented):
Looks really good, thanks for the work. I've only got minor issue with a few things, please see the comments.

cmd/tink-worker/worker/container_logger.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"io"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/packethost/pkg/log"
+)
+
+// ContainerLogger emits container logs.
+type ContainerLogger interface {
```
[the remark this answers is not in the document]
micahhausler: How about `LogCapturer`?

cmd/tink-worker/worker/registry.go
```diff
@@ -0,0 +1,69 @@
...
+	"encoding/base64"
+	"encoding/json"
+	"io"
+
+	"github.com/docker/docker/api/types"
+	"github.com/pkg/errors"
+)
+
+// RegistryConnDetails are the connection details for accessing a Docker
+// registry.
```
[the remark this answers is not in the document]
micahhausler: will fix, copied over

cmd/tink-worker/worker/container_manager.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	wfDir := filepath.Join(defaultDataDir, wfID)
+	hostConfig := &container.HostConfig{
+		Privileged: privileged,
+		Binds:      []string{wfDir + ":/workflow"},
+	}
+
+	if pidConfig := action.GetPid(); pidConfig != "" {
+		l.With("pid", pidConfig).Info("creating container")
```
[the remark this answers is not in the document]
micahhausler: I'll just drop this line

cmd/tink-worker/worker/container_manager_test.go
micahhausler: nice

cmd/tink-worker/worker/worker.go
micahhausler: fixed

cmd/tink-worker/worker/container_logger.go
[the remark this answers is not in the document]
mmlb: sounds less awkward yeah :D

cmd/tink-worker/worker/log_capturer_test.go
```diff
@@ -0,0 +1,108 @@
...
+			content: "Line1\nline2\n",
+		},
+		{
+			name:    "empty buffer from error",
+			writer:  *bytes.NewBufferString(""),
+			wanterr: errors.New("Docker failure"),
+			content: "",
+		},
+	}
+	logger := log.Test(t, "github.com/tinkerbell/tink")
```
mmlb, line 57: ditto

micahhausler: fixed

micahhausler:
@mmlb Fixed logger initialization

cmd/tink-worker/worker/log_capturer.go
```diff
@@ -0,0 +1,60 @@
...
+	"fmt"
+	"io"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/packethost/pkg/log"
+)
+
+// LogCapturer emits container logs.
+type LogCapturer interface {
```
chrisdoherty4, line 15: What consumes this interface? Nothing in this package seems to so I wonder if we could just export the struct and leave the interface declaration to consumers?

micahhausler: Sure, I can make the implementation public. I think I'll still leave the interface in this package unless there is a reason not to.

chrisdoherty4: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

micahhausler: I'll move the `LogCapturer` and `ContainerManager` interface declarations into `worker.go`. They're still in the same package, but will be declared at the site of use.

micahhausler:
@mmlb anything else on this?

cmd/tink-worker/cmd/root.go
```diff
@@ -3,16 +3,18 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"time"
 
+	dockerCli "github.com/docker/docker/client"
```
chrisdoherty4, line 10: Relatively minor: alias's should be all lowercase same as package names.

micahhausler: Fixed

thebsdbox (dismissed):
Fantastic, thanks for the additional testing !

cmd/tink-worker/worker/container_manager.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/packethost/pkg/log"
+	"github.com/pkg/errors"
+	pb "github.com/tinkerbell/tink/protos/workflow"
+)
+
+const (
+	ErrFailedToWait   = "failed to wait for completion of action"
+	ErrFailedToRunCmd = "failed to run on-timeout command"
+
+	InfoWaitFinished = "wait finished for failed or timeout container"
```
mmlb, line 20: I don't think these need to/should be public right? They're only used within this package.

mmlb (commented):
lgtm, just the one comment about the public strings that seem unnecessary.

cmd/tink-worker/worker/container_manager.go
micahhausler: I can make them private

mmlb (approved):

PR #599, opened 2022-03-22
Add running workflow for worker index to cache

micahhausler:
## Description

Adds indexer so clients can query for running workflows by a particular worker address

## Why is this needed

Boots needs to import this logic and query for these workflows

Fixes: # N/A

## How Has This Been Tested?

Added unit tests for existing and new index functions

N/A

## Checklist:

I have:

- [x] added unit or e2e tests

### DIFF
--- a/pkg/controllers/manager.go
+++ b/pkg/controllers/manager.go
@@ -17,7 +17,11 @@ import (
 )
 
 const (
-	WorkerAddr = "status.tasks.workeraddr"
+	WorkflowWorkerAddrIndex             = ".status.tasks.workerAddr"
+	WorkflowWorkerNonTerminalStateIndex = ".status.state.nonTerminalWorker"
+	WorkflowStateIndex                  = ".status.state"
+	HardwareMACAddrIndex                = ".spec.interfaces.dhcp.mac"
+	HardwareIPAddrIndex                 = ".spec.interfaces.dhcp.ip"
 )
 
 var (
@@ -47,6 +51,14 @@ func GetControllerOptions() controllerruntime.Options {
 		MetricsBindAddress:     fmt.Sprintf(":%d", options.MetricsPort),
 		HealthProbeBindAddress: fmt.Sprintf(":%d", options.HealthProbePort),
 	}
+}
+
+// GetNamespacedControllerOptions returns a set of options used by the Tink controller.
+// These options include leader election enabled.
+func GetNamespacedControllerOptions(namespace string) controllerruntime.Options {
+	opts := GetControllerOptions()
+	opts.Namespace = namespace
+	return opts
 }
 
 // GetServerOptions returns a set of options used by the Tink API.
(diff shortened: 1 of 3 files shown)


displague (approved):
This looks good and is consistent with our existing index.

I have some questions about the format of keys (should they start with a `.`), names of meta keys (is the snake case field name conventional), and return values (Indexer documentation says only return a single string for Kubernetes compatibility), but our existing index brings up the same questions.

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.11.1/pkg/client/interfaces.go#L125-L136

micahhausler:
> I have some questions about the format of keys (should they start with a .),

I'm happy to prefix with a period

> names of meta keys (is the snake case field name conventional)

Good catch, I'll fix this to be more inline with the API field name convention

> and return values (Indexer documentation says only return a single string for Kubernetes compatibility), but our existing index brings up the same questions. 

I believe this only matters if you wanted to compatibility with a field-selector in the API, which we don't here. We intentionally want all the workers for a particular workflow.

displague (approved):

pkg/controllers/manager.go
```diff
@@ -17,7 +17,8 @@ import (
 )
 
 const (
-	WorkerAddr = "status.tasks.workeraddr"
+	WorkerAddrIndex       = ".status.tasks.workerAddr"
+	NonTerminalStateIndex = ".status.state.nonTerminalWorker"
```
detiber, line 21: Curious to see if it would make sense to break this out on a per-state basis instead of just non-terminal states.

I'm also wondering if we might run into issues based on the field naming here, given the comment about the `FieldIndexer` interface here: https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.11.1/pkg/client#FieldIndexer

detiber (commented):
A couple of questions on the index itself, but I'm curious why this is part of `pkg/controllers/manager.go` if it's expected to be consumed by boots or other clients, would it make more sense to have it in a separate package to be consumed by the clients?

pkg/controllers/manager.go
micahhausler: > Curious to see if it would make sense to break this out on a per-state basis instead of just non-terminal states

I've added an additional index for state, `WorkflowStateIndex` and I've updated the index variable names for clarity. Take a look at the documented example under `FieldIndexer`, the Index name is just a key to lookup an object based on shared understanding of what that index means. In the documented example, the field name is `spec.volumes.secret.secretName` and returns all the secret names mounted on a pod.

micahhausler:
> A couple of questions on the index itself, but I'm curious why this is part of `pkg/controllers/manager.go` if it's expected to be consumed by boots or other clients, would it make more sense to have it in a separate package to be consumed by the clients?

Is there anywhere else you're thinking of specifically? This code is importable by Boots in the current package.  Boots will become just a client of Kubernetes, as opposed to the Tinkerbell API. In that architecture, Boots will need still need indexes on Hardware by MAC and IP, as well as [workflows for a particular worker](https://github.com/tinkerbell/boots/blob/main/packet/endpoints.go#L48).

thebsdbox (dismissed):
This looks OK to me, based upon reading through the code and a review of the comments. Unless @detiber has any strong feelings on the convo re:boots.

pkg/controllers/manager.go
```diff
@@ -119,16 +151,74 @@ func (m *GenericControllerManager) RegisterControllers(ctx context.Context, cont
 }
 
 // workerIndex func returns a list of worker addresses from a workflow.
-func wokerIndexFunc(obj client.Object) []string {
+func workflowWorkerAddrIndexFunc(obj client.Object) []string {
```
mmlb, line 154: comment needs updating to match func name

mmlb (commented):
lgtm after the func comment is fixed

pkg/controllers/manager.go
micahhausler: fixed

mmlb:
After taking a look at #600, does it make more sense to get that merged first and then use the defined constants instead of bare strings here?

micahhausler:
>  After taking a look at https://github.com/tinkerbell/tink/pull/600, does it make more sense to get that merged first and then use the defined constants instead of bare strings here?

Yep! I'll rebase this and reference the constants once #600 merges

mmlb (approved):
lgtm!

PR #602, opened 2022-04-05
Added virtual worker

micahhausler:
Add virtual tink worker command

## Description

* Add a no-op tinkerbell worker

## Why is this needed

This worker is helpful if you want to test Tinkerbell API changes while not actually executing a worker that executes docker containers.

Fixes: #

## How Has This Been Tested?

Tested locally against the Tinkerbell API. I'll have an E2E test in a future PR that uses the no-op logger and container manager.

No impact to existing users.

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/virtual-worker/Dockerfile
+++ b/cmd/virtual-worker/Dockerfile
@@ -1,0 +1,8 @@
+FROM alpine:3.15
+ENTRYPOINT [ "/usr/bin/virtual-worker" ]
+
+ARG TARGETARCH
+ARG TARGETVARIANT
+
+RUN apk add --no-cache --update --upgrade ca-certificates
+COPY virtual-worker-linux-${TARGETARCH:-amd64}${TARGETVARIANT} /usr/bin/virtual-worker
--- a/cmd/virtual-worker/cmd/root.go
+++ b/cmd/virtual-worker/cmd/root.go
@@ -1,0 +1,122 @@
+package cmd
+
+import (
+	"strings"
+	"time"
+
+	"github.com/packethost/pkg/log"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+	"github.com/tinkerbell/tink/client"
+	tinkWorker "github.com/tinkerbell/tink/cmd/tink-worker/worker"
+	"github.com/tinkerbell/tink/cmd/virtual-worker/worker"
+	"github.com/tinkerbell/tink/protos/workflow"
+)
+
+const (
+	defaultRetryIntervalSeconds = 3
+	defaultRetryCount           = 3
+	defaultMaxFileSize          = 10 * 1024 * 1024 // 10MB
+)
+
+// NewRootCommand creates a new Virtual Worker Cobra root command.
+func NewRootCommand(version string, logger log.Logger) *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:   "virtual-worker",
+		Short: "Virtual Tink Worker",
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			return createViper(logger, cmd)
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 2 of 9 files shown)


cmd/virtual-worker/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.11
```
mmlb, line 1: Every other Dockerfile is at 3.15, is 3.11 necessary?

cmd/virtual-worker/main.go
```diff
@@ -0,0 +1,28 @@
...
+	logger, err := log.Init(serviceKey)
+	if err != nil {
+		panic(err)
+	}
+
+	rootCmd := cmd.NewRootCommand(version, logger)
+	if err := rootCmd.Execute(); err != nil {
+		os.Exit(1)
+	}
+	logger.Close()
```
mmlb, line 27: This should probably be defer'd after the err check to ensure Close is called and any buffered messages are flushed.

mmlb (commented):
a couple small fixes needed but otherwise lgtm. I'd *really* love it if #584 would land first and then your root.go could be based off of tink-worker's from then, but if not then rebasing and adapting this in #584 would be fine too.

cmd/virtual-worker/Dockerfile
micahhausler: ah no, I just created this a long time ago, I'll update this

cmd/virtual-worker/main.go
micahhausler: > a couple small fixes needed but otherwise lgtm. I'd _really_ love it if #584 would land first and then your root.go could be based off of tink-worker's from then, but if not then rebasing and adapting this in #584 would be fine too.

I'm happy to wait on #584 and reuse cmd code from tink-worker

micahhausler:
> a couple small fixes needed but otherwise lgtm. I'd _really_ love it if #584 would land first and then your root.go could be based off of tink-worker's from then, but if not then rebasing and adapting this in #584 would be fine too.

I'm happy to wait on #584 and reuse cmd code from tink-worker

cmd/virtual-worker/cmd/root.go
```diff
@@ -0,0 +1,183 @@
...
+			if setupErr := client.Setup(); setupErr != nil {
+				return setupErr
+			}
+
+			ctx := context.Background()
+			if timeOut > 0 {
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithTimeout(ctx, timeOut)
+				defer cancel()
+			}
```
mmlb, line 60: should this go away now that tink-worker doesn't have this either?

cmd/virtual-worker/cmd/root.go
```diff
@@ -0,0 +1,183 @@
...
+	tinkWorker "github.com/tinkerbell/tink/cmd/tink-worker/worker"
+	"github.com/tinkerbell/tink/cmd/virtual-worker/worker"
+	pb "github.com/tinkerbell/tink/protos/workflow"
+	"google.golang.org/grpc"
+)
+
+const (
+	defaultRetryIntervalSeconds       = 3
+	defaultRetryCount                 = 3
+	defaultMaxFileSize          int64 = 10 * 1024 * 1024 // 10MB
```
mmlb, line 24: you can get rid of the `int64` here and leave it a pure/untyped constant for longer and let the compiler figure out the type. Same as tink-worker does.

cmd/virtual-worker/main.go
mmlb: Ah yeah I see that now. Yep lets fix it there too.

micahhausler: Ah you know what, golint-ci complains that the defer won't be called with the `os.Exit(1)` and requires this to be called here.

cmd/virtual-worker/cmd/root.go
micahhausler: done

mmlb (approved):

PR #603, opened 2022-04-05
Create an automated release workflow

detiber:
## Description

- When a tag is created and pushed that has a v prefix:
  - Automatically generate release notes with github
  - Push the pre-built sha-based image with a new tag for the version
  - Automatically create a release in github

- Updates the CI workflow to not run on tags with a v prefix

I modeled this off the automated process we are currently using for the sandbox releases.

## Why is this needed

- Automates the release process, which is currently manual
- Adds a tagged image associated with the release version

## How Has This Been Tested?

It's the same process as what has been merged to boots in https://github.com/tinkerbell/boots/pull/234

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,19 @@
+# Releasing
+
+## Process
+
+For version v0.x.y:
+
+1. Create the annotated tag
+    > NOTE: To use your GPG signature when pushing the tag, use `SIGN_TAG=1 ./contrib/tag-release.sh v0.x.y` instead)
+    - `./contrib/tag-release.sh v0.x.y`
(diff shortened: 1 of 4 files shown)


RELEASING.md
```diff
@@ -0,0 +1,19 @@
+# Releasing
+
+## Process
+
+For version v0.x.y:
+
+1. Create the annotated tag
```
displague, line 7: It seems like we'd be okay to use GitHub's release/tag tooling for the same effect? Is that accurate and if so should we say as much, or say why we should not?

detiber: The idea was to try to automate the release process. Longer term, I could see a manually triggered workflow could be created to automate the tagging instead of requiring the tag to be created and pushed externally.

jacobweinstock (approved):

jacobweinstock (approved):
thanks @detiber !

PR #605, opened 2022-04-11
deprecate Tink worker global timeout cli flag:

jacobweinstock:
## Description

The Tink worker global timeout cli flag doesn't implement the global timeout defined in a Tink server template correctly.

A single template can have multiple workers. So a single Tink worker defining a global timeout for all tasks is erroneous. It doesn't take the multiple workers idea into account.

A long term solution will need to be implemented. Something that takes into account multiple workers. This PR just removes this erroneous behavior.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/tink-worker/cmd/root.go
+++ b/cmd/tink-worker/cmd/root.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


cmd/tink-worker/cmd/root.go
```diff
@@ -59,9 +59,7 @@ func NewRootCommand(version string, logger log.Logger) *cobra.Command {
 
 			ctx := context.Background()
 			if timeOut > 0 {
-				var cancel context.CancelFunc
-				ctx, cancel = context.WithTimeout(ctx, timeOut)
-				defer cancel()
+				logger.Info("timeOut is deprecated")
```
mmlb, line 62: lets just get rid of the code, we have enough workarounds/deprecated things and it just turns into baggage. We're still at v <1.0 and also have sandbox to provide a cushion for breakages.

mmlb (changes requested):

cmd/tink-worker/cmd/root.go
jacobweinstock: sure thing

cmd/tink-worker/cmd/root.go
```diff
@@ -96,6 +86,7 @@ func NewRootCommand(version string, logger log.Logger) *cobra.Command {
 				worker.WithLogCapture(captureActionLogs),
 				worker.WithPrivileged(true))
 
+			ctx := context.Background()
 			err = w.ProcessWorkflowActions(ctx)
```
micahhausler, line 90: Can you use the command context here?
```suggestion
			err = w.ProcessWorkflowActions(cmd.Context())
```

micahhausler (changes requested):
Thanks for doing this! Just one minor change

micahhausler (approved):
LGTM

cmd/tink-worker/cmd/root.go
mmlb: This is currently effectively the same thing. Will change if we do `ExecuteContext` in cmd/tink-worker/main.go which I think would give us ctrl-c support? For a later PR though.

mmlb (approved):

PR #607, opened 2022-04-14
Move to path.Join for repo + image_name:

jacobweinstock:
## Description

With this a tink-worker can be started without a registry defined and templates in Tink server can specify any registry.

## Why is this needed

Currently, if a container registry is not specified then the container image will start with a `/`. This will cause tink worker to fail to pull/run the container. With this change, if a container registry is not specified the `/` will not be added and the contain image will be pulled from the container runtime's default registry. For containerd and docker (which is what Hook runs) the default registry is `docker.io`.

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/cmd/tink-worker/worker/registry.go
+++ b/cmd/tink-worker/worker/registry.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"io"
+	"path"
 
 	"github.com/docker/docker/api/types"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


micahhausler (approved):
LGTM

jacobweinstock:
Hey @mmlb. When you have a few moments, would you mind checking this out?

mmlb:
LGTM, sorry for not reviewing but I *really* didn't want to have something jump in front of #584 and dropping reviews and thus prolonging it even more.

mmlb (commented):
lgtm but needs a rebase

mmlb (approved):

PR #609, opened 2022-04-19
[tink worker] Make registry CLI flags optional

jacobweinstock:
## Description

Make `docker-registry`, `registry-username`, and `registry-password` CLI flags optional:

A previous commit allowed an empty registry to be used when running a tink worker action. This allows tink worker to take advantage of that update by not requiring these registry CLI flags.

Currently, if these values are not passed in via `/proc/cmdline` then tink worker will fail to start.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


micahhausler (approved):
LGTM

mmlb (approved):

PR #610, opened 2022-04-27
Add RBAC for tink-controller and tink-server

abhinavmpandey08:
## Description
This PR adds all deployment and RBAC manifests and `kustomization.yaml`s for `tink-controller` and `tink-server`.

## Why is this needed
This is needed to help users deploy the kubified tink services in their Kubernetes environment
This is what the generated manifest looks like after running `kustomize build config/default` https://gist.github.com/abhinavmpandey08/ed00314e4b738cacbad67ea9e345ec2a

## How Has This Been Tested?
This has been tested by applying the manifests on a KinD cluster and verifying that `tink-controller` and `tink-server` both work as intended.

No user impact

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/config/default/manager_image_patch.yaml
+++ b/config/default/manager_image_patch.yaml
@@ -1,0 +1,12 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+        # Change the value of image field below to your controller image URL
+        - image: tink-controller:latest
+          name: manager
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -1,0 +1,32 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
+  labels:
+    control-plane: controller-manager
+spec:
+  selector:
+    matchLabels:
+      control-plane: controller-manager
+  replicas: 1
+  template:
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
+      labels:
+        control-plane: controller-manager
+    spec:
+      containers:
+        - image: controller:latest
+          imagePullPolicy: IfNotPresent
+          name: manager
+          resources:
+            limits:
+              cpu: 500m
+              memory: 128Mi
+            requests:
+              cpu: 10m
+              memory: 64Mi
+      serviceAccountName: controller-manager
+      terminationGracePeriodSeconds: 10
--- a/config/rbac/leader_election_role.yaml
+++ b/config/rbac/leader_election_role.yaml
@@ -1,0 +1,36 @@
+# permissions to do leader election.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: leader-election-role
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+  - apiGroups:
+      - coordination.k8s.io
+    resources:
+      - leases
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
+  - apiGroups:
+      - ""
+    resources:
+      - events
+    verbs:
+      - create
+      - patch
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -1,0 +1,41 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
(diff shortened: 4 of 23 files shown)


config/crd/patches/cainjection_in_hardware.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
micahhausler, line 1: Why do we need CAs in the CRDs?

config/crd/patches/webhook_in_hardware.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: hardware.tinkerbell.org
+spec:
+  conversion:
+    strategy: Webhook
+    webhook:
+      clientConfig:
+        service:
+          namespace: system
```
micahhausler, line 12: We don't have a webhook yet, should we drop this?

config/manager/manager.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    control-plane: controller-manager
+  name: system
```
micahhausler, line 6: can we call this `tink-system` or `tinkerbell-system`?

config/rbac/leader_election_role.yaml
```diff
@@ -0,0 +1,37 @@
...
+    resources:
+      - configmaps
+    verbs:
+      - get
+      - list
+      - watch
+      - create
+      - update
+      - patch
+      - delete
```
micahhausler, line 18: Why are full config map permissions required in the namespace?

config/rbac/role.yaml
```diff
@@ -0,0 +1,45 @@
...
+  creationTimestamp: null
+  name: manager-role
+rules:
+  - apiGroups:
+      - tinkerbell.org
+    resources:
+      - hardware
+      - hardware/status
+    verbs:
+      - create
```
micahhausler, line 14: I don't think the manager ever needs to create hardware
```suggestion
```

config/rbac/role.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - update
+      - watch
+  - apiGroups:
+      - tinkerbell.org
+    resources:
+      - templates
+      - templates/status
+    verbs:
+      - create
+      - delete
```
micahhausler, line 27: I don't think the manager ever needs to create or delete templates
```suggestion
```

config/rbac/role.yaml
```diff
@@ -0,0 +1,45 @@
...
+      - update
+      - watch
+  - apiGroups:
+      - tinkerbell.org
+    resources:
+      - workflows
+      - workflows/status
+    verbs:
+      - create
+      - delete
```
micahhausler, line 40: I don't think the manager ever needs to create workflows. We might want a sweeper that removes old workflows, so delete can stay
```suggestion
      - delete
```

config/tink-server/role.yaml
```diff
@@ -0,0 +1,36 @@
+---
```
micahhausler, line 1: This looks like a duplicate of `config/tink-server/rbac/role.yaml` right?

micahhausler (changes requested):
Thank you so much for this, awesome work! Just a few minor changes

config/rbac/leader_election_role.yaml
abhinavmpandey08: This role is auto-generated by kubebuilder. I am not sure if all these permissions are really required during leader-election or not. Will check and update here

config/tink-server/role.yaml
abhinavmpandey08: yep, removed

config/rbac/leader_election_role.yaml
abhinavmpandey08: Alright I just tested and looks like it needs all the permissions on cm except for delete

micahhausler (approved):
LGTM!

config/default/manager_image_patch.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: controller-manager
+  namespace: system
```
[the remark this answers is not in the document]
abhinavmpandey08: kustomize overrides the namespace on all the namespaced objects to `tink-system`

jacobweinstock:
@mmlb, can you check this out? approve the workflow run?

gauravgahlot (approved):
The workflow is green and the changes LGTM. Will wait for @mmlb's review though.

micahhausler (approved):
LGTM

mmlb (approved):

PR #614, opened 2022-05-03
Add Hardware resources and BMCRef

micahhausler:

## Description

This PR adds two fields for external orchestrators ([such as CAPT](https://github.com/tinkerbell/cluster-api-provider-tinkerbell/)) to 
* Make scheduling (machine assignment) based on resources declared on a particular hardware
* Manage a machine's lifecycle and maintain a relation to the hardware ([such as Rufio](https://github.com/tinkerbell/rufio))
 
## How Has This Been Tested?

N/A. No 

No change to existing users.

## Checklist:

I have:

- [ ] added unit or e2e tests


chrisdoherty4 (approved):
lgtm

config/crd/bases/tinkerbell.org_hardware.yaml
```diff
@@ -318,6 +341,17 @@ spec:
                     state:
                       type: string
                   type: object
+                resources:
+                  additionalProperties:
+                    anyOf:
+                      - type: integer
```
chrisdoherty4, line 347: Do we need clarification in the `description` on the assumed unit when specifying as an integer?

chrisdoherty4: In-fact I don't think that makes sense because describing a CPUs is a single number. Something still seems odd though, how are users meant to know what value to use?

micahhausler: I don't think so, its a Kubernetes type which has [its own documentation](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#Quantity)

micahhausler: CPUs can be fractional for K8s pods, or whole numbers. See the [Format godoc example](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource#example-Format)

chrisdoherty4: Right, but fractionality doesn't make sense in Tinkerbell. We only talk in terms of whole CPUs hence I'm wondering if we need to better define boundaries seeming as we're leveraging an existing type with behavior beyond our needs.

chrisdoherty4: To clarify, I agree this doesn't need additional documentation on my OP about units assumed with integers. I've shifted to "are our bounds adequately described" because the `resource.Quantity` supports stuff beyond what we want.

chrisdoherty4 (approved):

chrisdoherty4 (approved):

PR #613, opened 2022-05-03
Add make targets to generate release Kubernetes manifests

abhinavmpandey08:
## Description
This PR adds Makefile target `release` to generate Kubernetes manifests for `tink-server` and `tink-controller`

## Why is this needed
This is needed to generate the release manifests using `kustomize`

## How Has This Been Tested?
Ran `make release` and verified that the manifest gets generated properly 

## How are existing users impacted? What migration steps/scripts do we need?
No user impact

## Checklist:

I have:

- [ ] updated the documentation and/or roadmap (if required)
- [ ] added unit or e2e tests
- [ ] provided instructions on how to upgrade

PR #615, opened 2022-05-03
Fix some issues from #613

mmlb:
## Description

#613 introduced a few unwanted bits and a typo.

## Why is this needed

#613 did not use tools.go setup we have for using pinned go tools.

I reworked it so it also didn't do recursive make as that is considered harmful and setup better dependencies so that we let make be better equipped to do the right thing. This way we avoid using make as a dumb command runner, which if not done can lead to racy/broken makefiles (see hook for example).

I fixed a typo that would cause the manager to use tink-server image, likely to not work :D.

In the process I noticed that we don't have `generate-manifests` hooked up to be run as part of ci to detect drift, so I added it as a dependency of `generated` as it should have been from the beginning.

## How Has This Been Tested?

I ran make a bunch.


kube.mk
```diff
@@ -35,59 +11,55 @@ generate-go: bin/controller-gen bin/gofumpt # Generate Go code.
 	gofumpt -w -s ./pkg/apis
 
 .PHONY: generate-manifests
-generate-manifests: bin/controller-gen # Generate manifests e.g. CRD, RBAC etc.
+generate-manifests: generate-crds generate-rbacs generate-server-rbacs # Generate manifests e.g. CRD, RBAC etc.
```
micahhausler, line 14: want this to be a doc?
```suggestion
generate-manifests: generate-crds generate-rbacs generate-server-rbacs ## Generate manifests e.g. CRD, RBAC etc.
```

mmlb: Hmm good catch this was from before I just tacked it onto `generate`.

mmlb: Oh wait I misread the diff. No I don't think its necessary, `make generate` will do the right thing. If you'd like it to be documented I have no problem with it showing up in make help. Do *you* want this to be a doc?

micahhausler: Ha It doesn't matter to me. Since this isn't a top level target, I'm fine with keeping it not doc'd

micahhausler (approved):

kube.mk
mmlb: Yeah thats been my thinking.

PR #617, opened 2022-05-04
Add image substitutions in kustomization instead of doing a patch

abhinavmpandey08:
## Description
This PR refactors the `release-manifests` target to use `kustomize` to substitute the tink-controller and tink-servers images instead of using patches.

## Why is this needed
There was a bug in the previous `release-manifests` target where the tink-server image wasn't being updated because its dependencies were all being met already causing it to not be re-build. This PR refactors it to not use the image patches removing the need to update these patch files.

## How Has This Been Tested?
ran `make release-manifests` and verified that the kustomization.yaml and tink.yaml have the expected outputs

No user impact

## Checklist:

I have:

- [ ] added unit or e2e tests


kube.mk
```diff
@@ -40,25 +40,19 @@ generate-server-rbacs: bin/controller-gen
...
+TINK_CONTROLLER_IMAGE ?= quay.io/tinkerbell/tink-controller
+TINK_SERVER_TAG ?= latest
+TINK_CONTROLLER_TAG ?= latest
 
 out/release/default/kustomization.yaml: config/default/kustomization.yaml
 	rm -rf out/
 	mkdir -p out/
 	cp -a config/ out/release/
+	sed -e 's|TINK_CONTROLLER_IMAGE|$(TINK_CONTROLLER_IMAGE)|g' -e 's|TINK_CONTROLLER_TAG|$(TINK_CONTROLLER_TAG)|g' \
+		-e 's|TINK_SERVER_IMAGE|$(TINK_SERVER_IMAGE)|g' -e 's|TINK_SERVER_TAG|$(TINK_SERVER_TAG)|g' $^ > $@
```
mmlb, line 53: Instead of having the `images:` block in the original file and forcing the use of the file to always have to go through `make` (to avoid what are effectively useless names/tags) why not just do something like:

```
cat >>$@ <<-EOF
images:
  - name: tink-controller
    newName: $(TINK_CONTROLLER_IMAGE)
    newTag: $(TINK_CONTROLLER_TAG)
  - name: tink-server
    newName: $(TINK_SERVER_IMAGE)
    newTag: $(TINK_SERVER_TAG)
EOF
```
?

kube.mk
```diff
@@ -40,25 +40,19 @@ generate-server-rbacs: bin/controller-gen
...
+TINK_SERVER_IMAGE ?= quay.io/tinkerbell/tink-server
+TINK_CONTROLLER_IMAGE ?= quay.io/tinkerbell/tink-controller
+TINK_SERVER_TAG ?= latest
+TINK_CONTROLLER_TAG ?= latest
 
 out/release/default/kustomization.yaml: config/default/kustomization.yaml
 	rm -rf out/
 	mkdir -p out/
 	cp -a config/ out/release/
+	sed -e 's|TINK_CONTROLLER_IMAGE|$(TINK_CONTROLLER_IMAGE)|g' -e 's|TINK_CONTROLLER_TAG|$(TINK_CONTROLLER_TAG)|g' \
```
jacobweinstock, line 52: kustomize does have an edit command. might be something to consider. for example:

```bash
./bin/kustomize edit set image tink-controller=quay.io/tinkerbell/tink-controller:v0.1.1
```

mmlb (commented):
I do like this way over the {manager|server}_image.yaml from before :+1:

config/default/kustomization.yaml
```diff
@@ -18,6 +18,10 @@ bases:
...
-patchesStrategicMerge:
-  - manager_image.yaml
-  - server_image.yaml
+images: 
+- name: tink-controller
+  newName: TINK_CONTROLLER_IMAGE
+  newTag: TINK_CONTROLLER_TAG
+- name: tink-server
+  newName: TINK_SERVER_IMAGE
+  newTag: TINK_SERVER_TAG
```
jacobweinstock, line 27: looks like some yamllint formatting will be required here.

kube.mk
abhinavmpandey08: Removed the images block from kustomization. It will be added dynamically using kustomize like @jacobweinstock suggested

config/default/kustomization.yaml
abhinavmpandey08: removed this section so this shouldn't be an issue anymore

kube.mk
mmlb: nice :+1:

kube.mk
abhinavmpandey08: done

kube.mk
```diff
@@ -40,25 +40,18 @@ generate-server-rbacs: bin/controller-gen
...
+TINK_CONTROLLER_TAG ?= latest
 
 out/release/default/kustomization.yaml: config/default/kustomization.yaml
 	rm -rf out/
 	mkdir -p out/
 	cp -a config/ out/release/
 
-out/release/tink.yaml: bin/kustomize generate-manifests out/release/default/manager_image.yaml out/release/default/manager_image.yaml
+out/release/tink.yaml: bin/kustomize generate-manifests out/release/default/kustomization.yaml
+	cd out/release/default && kustomize edit set image server=$(TINK_SERVER_IMAGE):$(TINK_CONTROLLER_TAG) controller=$(TINK_CONTROLLER_IMAGE):$(TINK_CONTROLLER_TAG)
```
mmlb, line 54: we have .ONESHELL set, so the whole recipe is run as one shell command instead of one per-line. This should thus be `(cd out/release/default && kustomize edit set...)` so the `cd` happens in a sub shell.

abhinavmpandey08: done

mmlb (approved):

jacobweinstock (approved):

PR #620, opened 2022-05-06
Expose indexing funcs

micahhausler:

## Description

Making these functions public will allow other packages (such as [Boots](https://github.com/tinkerbell/boots/blob/4d162dd/client/kubernetes/cluster.go#L83-L135)) to reuse these functions without needing to import the entire manager.


jacobweinstock (approved):
lgtm

PR #622, opened 2022-05-13
Add kube namespace override and default to in cluster namespace

chrisdoherty4:
See https://github.com/tinkerbell/hegel/pull/106 for explanation.


micahhausler (approved):
LGTM

PR #623, opened 2022-05-19
Add clusterctl labels to Tink CRDs to allow them to be moved

abhinavmpandey08:
Signed-off-by: Abhinav Pandey <abhinavmpandey08@gmail.com>

## Description
Add these two labels to the Tink CRD definitions:
```yaml
  clusterctl.cluster.x-k8s.io: ""
  clusterctl.cluster.x-k8s.io/move: ""
```

## Why is this needed
Adding these labels allows clusterctl to recognize these objects during cluster move operations 

## How Has This Been Tested?
ran `make release-manifests` and verified that the labels were added properly to the Hardware, Template and Workflow CRDs

No existing user impact

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -8,6 +8,11 @@ resources:
   - bases/tinkerbell.org_workflowdata.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
+# Labels to add to all resources and selectors.
+commonLabels:
+  clusterctl.cluster.x-k8s.io: ""
+  clusterctl.cluster.x-k8s.io/move: ""
+
 # the following config is for teaching kustomize how to do kustomization for CRDs.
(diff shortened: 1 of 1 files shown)


chrisdoherty4 (approved):
Lgtm

config/crd/kustomization.yaml
```diff
@@ -8,6 +8,11 @@ resources:
   - bases/tinkerbell.org_workflowdata.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
+# Labels to add to all resources and selectors.
+commonLabels:
```
chrisdoherty4, line 12: Given recent discussions do we still want this? Is it going to apply it to all CRDs? I suspect we don't want to move Workflow's given they're ephemeral.

PR #628, opened 2022-06-13
Validate mac address on Hardware spec:

jacobweinstock:
## Description

This only allows lower case mac addresses. This simplifies matching/looking up mac addresses. Without this values could be upper case or mixed case. Upper case and mixed cases will cause clients to miss lookups if using the wrong case. This standardized to lowercase to remove this error prone interaction.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


chrisdoherty4 (approved):

micahhausler (approved):
LGTM

PR #630, opened 2022-06-15
Fixed handling of failed task

micahhausler:

## Description

Failed tasks did not properly update the workflow status, and the overall workflow status perpetually showed as `STATE_RUNNING`

## Why is this needed

To properly report a failed workflow

Fixes: #

## How Has This Been Tested?

Added unit tests

N/A

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/server/kubernetes_api_workflow.go
+++ b/server/kubernetes_api_workflow.go
@@ -127,8 +127,7 @@ cont:
 			t := metav1.NewTime(s.nowFunc())
 			return &t
 		}()
-	case pb.State_STATE_FAILED:
-	case pb.State_STATE_TIMEOUT:
+	case pb.State_STATE_FAILED, pb.State_STATE_TIMEOUT:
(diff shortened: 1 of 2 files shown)


server/kubernetes_api_workflow.go
```diff
@@ -128,6 +128,7 @@ cont:
 			return &t
 		}()
 	case pb.State_STATE_FAILED:
+		fallthrough
 	case pb.State_STATE_TIMEOUT:
```
mmlb, line 132: can we just do `case pb.State_STATE_FAILED, pb.State_STATE_TIMEOUT:` here? Doing it this way reinforces that they are both to have the same exact code.

micahhausler: done

chrisdoherty4 (approved):

jacobweinstock (approved):

mmlb (approved):

Issue #531, opened 2021-09-01
RPC `Hardware.HardwareServer.DeprecatedWatch` panics server when called


jacobweinstock:
Anyone familiar with what's going on with [`DeprecatedWatch`](https://github.com/tinkerbell/tink/blob/18bc57d9a8f67752eaee11078e714e6b4b5a69b7/protos/hardware/hardware.proto#L66) in the [hardware.proto](https://github.com/tinkerbell/tink/blob/master/protos/hardware/hardware.proto) file?

It has an implementation [here](https://github.com/tinkerbell/tink/blob/18bc57d9a8f67752eaee11078e714e6b4b5a69b7/grpc-server/hardware.go#L184), which I would assume a deprecated RPC implementation would just return an ["Unimplemented" status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html) or similar instead. What we have is a full implementation that if called, will panic the Tink service.
```bash
panic: assignment to entry in nil map
tink/grpc-server/hardware.go:194
```
Does anyone with commits in this hardware file have any context here, by chance? @gauravgahlot @parauliya @kqdeng @mmlb 

## Expected Behaviour

A deprecated RPC should be removed or return a [common gRPC status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html). Overall, it should not panic the service.

## Current Behaviour

The `DeprecatedWatch` RPC panics when called.

## Possible Solution

1. Remove the RPC from [hardware.proto](https://github.com/tinkerbell/tink/blob/master/protos/hardware/hardware.proto) and remove the [implementation](https://github.com/tinkerbell/tink/blob/18bc57d9a8f67752eaee11078e714e6b4b5a69b7/grpc-server/hardware.go#L184).
2. Update the implementation [here](https://github.com/tinkerbell/tink/blob/18bc57d9a8f67752eaee11078e714e6b4b5a69b7/grpc-server/hardware.go#L184) to only return a [common gRPC status code](https://grpc.github.io/grpc/core/md_doc_statuscodes.html). For example:
    ```go
    return status.Error(codes.Unimplemented, "DeprecatedWatch is, well, deprecated")
    ```
3. Get it working. Rename `DeprecatedWatch` to `Watch` and add `watch:  make(map[string]chan string),` after [line 61](https://github.com/tinkerbell/tink/blob/d0535766650c85528e37dad40b1fc703604713d2/grpc-server/grpc_server.go#L61) in `grpc_server.go`.

[...]

PR #633, opened 2022-07-16
Reenable Hardware watch

gauravgahlot:
## Description

When we introduced events to Tink, we changed `Watch` to `DeprecatedWatch` because the new event system was to be used for watching any resource type (template, hardware, workflow). However, with events being dropped, we should reenable the Hardware watch.

## Why is this needed

Fixes: #

#531  

## Checklist:

I have:

- [ ] added unit or e2e tests


chrisdoherty4:
Hey @gauravgahlot. What's the rationale for providing an API in Tink to watch Hardware? What events are generated and what makes the changes important to a consumer?

chrisdoherty4:
Having had no response we'll close and circle back should we need this in the future.

PR #638, opened 2022-08-25
Expose custom funcs to template

chrisdoherty4:
Having exposed a subset of hardware data to templates there's a need to provide some helper functions for manipulating the data. The concrete use case is formatting of disk partitions.

This PR introduces `formatPartition` to templates accepting the device path. The function currently supports historic block devices (`/dev/sd<char>`) and NVMe drives. Additionally it adds `strings.Contains`, `strings.HasPrefix` and `strings.HasSuffix` which have proven useful. 

**Example template snippet**
```
name: {{ .Hardware.Name }}
disk: {{ formatPartition ( index .Hardware.Disks 0 ) 1 }}
```

Documentation will need adding to the docs website in the on-going documentation effort.

### DIFF
--- a/workflow/funcs.go
+++ b/workflow/funcs.go
@@ -1,0 +1,30 @@
+package workflow
+
+import (
+	"fmt"
+	"strings"
+)
+
+// templateFuncs defines the custom functions available to workflow templates.
+var templateFuncs = map[string]interface{}{
+	"contains":        strings.Contains,
+	"hasPrefix":       strings.HasPrefix,
+	"hasSuffix":       strings.HasSuffix,
+	"formatPartition": formatPartition,
+}
+
+// formatPartition formats a device path with partition for the device type. If it receives an
+// unidentifiable device path it returns the dev.
+//
+// Examples
+// 		formatPartition("/dev/nvme0n1", 0) -> /dev/nvme0n1p1
+// 		formatPartition("/dev/sda", 1) -> /dev/sda1
+func formatPartition(dev string, partition int) string {
+	switch {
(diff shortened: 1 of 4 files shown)


displague:
> Documentation will need adding to the docs website in the on-going documentation effort.

In this PR and in added documentation, specifying that the device partition format is the Linux kernel format may avoid some confusion.

chrisdoherty4:
@displague Good call out. Are you thinking in the function comment (and the docs site when it goes up)?

workflow/funcs.go
```diff
@@ -0,0 +1,30 @@
...
+	"formatPartition": formatPartition,
+}
+
+// formatPartition formats a device path with partition for the device type. If it receives an
+// unidentifiable device path it returns the dev.
+//
+// Examples
+// 		formatPartition("/dev/nvme0n1", 0) -> /dev/nvme0n1p1
+// 		formatPartition("/dev/sda", 1) -> /dev/sda1
+func formatPartition(dev string, partition int) string {
```
chrisdoherty4, line 22: Are we OK with this name? I originally had `formatDevicePartition` and @displague called out its specific to Linux.

displague (approved):
This will be very helpful in future workflows - no doubt we will want to build on this with other internal helper functions (for other string comparisons or modifications) and also for metadata mining.

I could see use in having a helper that chooses the most appropriate disk based on some heuristics, or one that returns the appropriate NIC or bonding device names.

I found the existing template operators helpful when reviewing this https://pkg.go.dev/text/template#pkg-overview (I wanted to make sure we weren't duplicating functionality with the string comparators).

PR #642, opened 2022-09-14
Add VLAN ID to hardware interface spec

jacobweinstock:
## Description

This adds vlan id to interfaces in the hardware spec.

FYI, hidden inside of all this generated code, linting, and dependency version updates is the reason for this PR. The small addition of vlan id to the hardware interface struct. Look here: `/pkg/apis/core/v1alpha1/hardware_types.go`.

Also, important to note that this PR updates the Go version from `1.17` to `1.18`, this was required to get the make targets for generating code functioning properly for me.

## Why is this needed

Needed to enabled trunk port support across the stack. See proposal https://github.com/tinkerbell/proposals/pull/60

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests


http-server/http_server.go
```diff
@@ -24,7 +24,7 @@ func SetupHTTP(ctx context.Context, logger log.Logger, authority string, errCh c
 	http.HandleFunc("/version", getGitRevJSONHandler())
 	http.HandleFunc("/healthz", healthCheckHandler)
 
-	srv := &http.Server{
+	srv := &http.Server{ //nolint:gosec // TODO: fix Potential Slowloris Attack because ReadHeaderTimeout is not configured
```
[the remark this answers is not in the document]
jacobweinstock: @sonatype-lift ignore

pkg/apis/core/v1alpha1/hardware_types.go
```diff
@@ -130,6 +129,7 @@ type DHCP struct {
 	UEFI        bool     `json:"uefi,omitempty"`
 	IfaceName   string   `json:"iface_name,omitempty"`
 	IP          *IP      `json:"ip,omitempty"`
+	VLANID      uint     `json:"vlan_id,omitempty"`
```
chrisdoherty4, line 132: Any particular reason we're using unsigned?

jacobweinstock: valid VLANs are from 0-4096 so i wanted to eliminate negative numbers. This makes me think i should add a regex check to this affect.

jacobweinstock: Thinking about this, i actually want this to be a string. so that the default value is not 0 but “”.

jacobweinstock: I've changed it to be a string and added a regex validation.

chrisdoherty4 (approved):

PR #648, opened 2022-11-04
Fix tag selection on cli make target

electrocucaracha:
## Description

This change reduces the number of output displayed by `git tag` command.

## Why is this needed

Some commits (like 9c098c135ff221757e9baa4116900600dbeefe07) contains multiple tags, causing issues during the execution of `make cli`.

## How Has This Been Tested?

```bash
$ git clone --depth 1 -b v0.8.0 https://github.com/tinkerbell/tink /tmp/tink
$ cd /tmp/tink
$ make cli
```

NA

## Checklist:

I have:

- [ ] added unit or e2e tests


jacobweinstock (approved):
Great catch, thank you @electrocucaracha!

PR #656, opened 2022-12-23
Simplify and tidy Makefile

chrisdoherty4:
The project Makefile had lots of targets some of which perform duplicate functions and are, at times, tough to reconcile. 

This PR seeks to simplify the targets used and clearly document how they work for future maintainers. 

There's too much happening to explain in detail. In summary:

- `build` can be used to build all binaries and supports cross platform builds with GOOS and GOARCH.
- Each binary can be built individually with `tink-server` etc targets.
- Images can be built with `images`
- Individual images can be built with the name appended with `-image` e.g. `tink-server-image`.
- Generated YAML now gets formatted such that it satisfies the linters.


Makefile
```diff
@@ -1,82 +1,94 @@
...
+		-o ./bin/$@-$(GOOS)-$(GOARCH) \
+		./cmd/$@
+
+# IMAGE_ARGS is resolved when its used in the `%-image` targets. Consequently, the $* automatic 
+# variable isn't evaluated until the target is called.  
+# See https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html.
+IMAGE_ARGS ?= -t $*
+
+.PHONY: images
+images: $(addsuffix -image, $(BINARIES)) ## Build all tink container images. All images will be for Linux but target the host arch.
```
jacobweinstock, line 73: Hmm....this target is not working for me. `make images`: `make: Nothing to be done for 'images'`
same with the individual image targets. `make: Nothing to be done for 'tink-server-image'`

Makefile
```diff
@@ -88,23 +100,16 @@ check-pbfiles: pbfiles
...
-e2etest-setup: tools
-	setup-envtest use
-
-## --------------------------------------
-## Generate
-## --------------------------------------
-
 .PHONY: generate
-generate: generate-go generate-manifests # Generate code, manifests etc.
+generate: generate-go generate-manifests ## Generate code, manifests etc.
```
jacobweinstock, line 104: when i do a `make generate` and then a `make lint`, `make lint` fails. is that expected behavior?

Makefile
jacobweinstock: fixed, thanks!

jacobweinstock (approved):

jacobweinstock:
@mergify queue

PR #657, opened 2022-12-23
Update, simplify and remove redundant steps in CI config

chrisdoherty4:
- Update action version for every action
- Remove superfluous steps such as setting up QEMU
- Reference the same Go version throughout

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-name: For each commit and PR
+name: Tink
 on:
   push:
     branches:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -92,36 +84,30 @@ jobs:
...
       - name: Login to quay.io
-        uses: docker/login-action@v1
+        uses: docker/login-action@v2
         if: ${{ startsWith(github.ref, 'refs/heads/main') }}
         with:
           registry: quay.io
           username: ${{ secrets.QUAY_USERNAME }}
           password: ${{ secrets.QUAY_PASSWORD }}
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v1
```
jacobweinstock, line 109: dont we need qemu for Docker multi-arch builds?

chrisdoherty4: Nope. I think this is a relic of when multi-arch builds included binary compilation. The system now cross compiles the binaries in an earlier job. `docker build` doesn't need QEMU to build arch specific images.

jacobweinstock (approved):

jacobweinstock:
@mergify queue

PR #663, opened 2022-12-27
Clean up and rearrange packages

chrisdoherty4:
Execute on some general cleanup for unused/deprecated/obsolete functionality and rearrange packages.

- Remove TLS support in Tink Server
- Remove Postgres configurability
- Remove metrics package; it contained cache related metrics that were possibly related to the old cacher backend and didn't seem useful, everything else leftover didn't seem useful either
- Move packages that can be easily moved and are internal to /internal
- Update some package names to be more representative of their function
- Remove obsolete/unused code

Controllers contain code generation integration that needs to be reworked separately so as to avoid lots of churn on this PR.


cmd/tink-worker/cmd/root.go
```diff
@@ -92,7 +89,6 @@ func NewRootCommand(version string, logger log.Logger) *cobra.Command {
 	rootCmd.Flags().Int("max-retry", defaultRetryCount, "Maximum number of retries to attempt (MAX_RETRY)")
 	rootCmd.Flags().Int64("max-file-size", defaultMaxFileSize, "Maximum file size in bytes (MAX_FILE_SIZE)")
 	rootCmd.Flags().Bool("capture-action-logs", true, "Capture action container output as part of worker logs")
-	rootCmd.Flags().Bool("tinkerbell-tls", true, "Connect to server via TLS or not (TINKERBELL_TLS)")
```
jacobweinstock, line 95: I believe we still want to have this optionally enabled for tink-worker. We could conceivably run the tink server behind a TLS-terminated ingress controller and then we'd want to have the tink-worker <-> server communication with TLS.

jacobweinstock (commented):
Looking good, thanks @chrisdoherty4

cmd/tink-worker/cmd/root.go
chrisdoherty4: That's fair. I figured we'd reimplement across the board in a saner way but no harm maintaining it until then.

chrisdoherty4: Added TLS back to server, worker and virtual worker as before.

jacobweinstock: Ah sorry, I actually wasn't suggesting we put it all back. Only for tink worker to use TLS. And not TLS that requires a cert to be passed in. So only sever side TLS from a know root CA's.

chrisdoherty4: So we want to officially announce no TLS support in Tink Server, for now? That's the direction Hegel has gone also and I expect the rest of the stack to go because the TLS implementations felt a little hacky.

Long term I think it would be good to support TLS configuration as an opt-in, thoughts?

jacobweinstock: In tink server, long term, I don't know if we should support TLS termination. Definitely a conversation to have though. I think, for sure, some detailed documentation needs to be created to describe the different TLS termination options and highlight that TLS communication between server and worker can still be achieved without functionality to terminate in tink server.

jacobweinstock (dismissed):

cmd/tink-worker/cmd/root.go
chrisdoherty4: Sounds good. Will strip it out of this PR.

chrisdoherty4: Stripped TLS from the server

internal/tlscert/tlscert.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"crypto/tls"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+)
+
+// Load TLS certificate from the given directory. The public key is expected to be named
+// "bundle.pem" and the private key "server-key.pem".
+func Load(certsDir string) (*tls.Certificate, error) {
```
jacobweinstock, line 12: Is this needed still?

chrisdoherty4: Oops, nope.

jacobweinstock (approved):

PR #670, opened 2023-01-18
Fix of template CRD example

AleksandrNull:
## Description

1. Prepend writefile image with quay registry path
2. kexec version fix (there is no 1.0.1 tag)

## Why is this needed

Examples do not work out-of-the-box

## Tested at my local setup as-is


jacobweinstock:
thanks @AleksandrNull ! mind signing the commit? `git commit --amend --signoff --no-edit` so the DCO check passes, please?

jacobweinstock (approved):

PR #675, opened 2023-03-21
Add v1alpha2 Hardware validating admission webhook

chrisdoherty4:
Add a [validating admission webhook](https://book.kubebuilder.io/reference/admission-webhook.html) to perform complex Hardware validation that can't be performed with [CEL](https://kubernetes.io/docs/reference/using-api/cel/).


internal/hardware/admission.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// Ensure there's no hardware in the cluster with the same IP addresses.
+	if resp := a.validateUniqueIPs(ctx, &hw); !resp.Allowed {
+		return resp
+	}
+
+	return admission.Allowed("")
+}
+
+func (a *Admission) InjectDecoder(d *admission.Decoder) error {
```
chrisdoherty4, line 58: Comment

api/v1alpha2/hardware.go
```diff
@@ -41,19 +41,31 @@ type NetworkInterfaces map[MAC]NetworkInterface
 
 // NetworkInterface is the desired configuration for a particular network interface.
 type NetworkInterface struct {
-	// DHCP is the basic network information for serving DHCP requests.
-	DHCP DHCP `json:"dhcp,omitempty"`
+	// DHCP is the basic network information for serving DHCP requests. Requires when DisbaleDHCP
```
jacobweinstock, line 44: `Requires` -> `Required`?

internal/hardware/admission.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/tinkerbell/tink/api/v1alpha2"
+	"github.com/tinkerbell/tink/internal/hardware/internal"
+	ctrl "sigs.k8s.io/controller-runtime"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// AdmissionWebhookEndpoint is the endpoint serving the Admission handler.
+const AdmissionWebhookEndpoint = "/validate-tinkerbell-org-v1alpha2-hardware"
```
jacobweinstock, line 18: does this need to exported?

internal/hardware/admission_mac.go
```diff
@@ -0,0 +1,70 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/tinkerbell/tink/api/v1alpha2"
+	"github.com/tinkerbell/tink/internal/hardware/internal"
+	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+func (a *Admission) validateMACs(hw *v1alpha2.Hardware) admission.Response {
```
jacobweinstock, line 16: Could a cel validation be used instead? `+kubebuilder:validation:XValidation:rule=` https://kubernetes.io/blog/2022/09/29/enforce-immutability-using-cel/

jacobweinstock (changes requested):
Thanks @chrisdoherty4. Question, how is this deployed?

internal/hardware/admission.go
chrisdoherty4: Yes, see the broader comment on usage.

internal/hardware/admission_mac.go
chrisdoherty4: I don't think we can run regex using the `XValidation` rule - I can't find anything in the link or other resources that shows how.

Even if it could, I'm not sure it would work for _map keys_. We have CEL validation configured on the map key type which I'd hoped would translate to applying the regex, but it seems the open API spec supported by the `CustomResourceDefinition` kind doesn't support validation of map keys (not that I've found yet anyway).

chrisdoherty4: > Thanks @chrisdoherty4. Question, how is this deployed?

Great question.

This will be registered with the new controller manager component (we already have a manager under `/internal/controllers`). The manager is a controller-runtime thing that handles initialization of controllers, webhooks, clients and has responsibilities such as leader election etc. 

The `Admission` object satisfies the [`Handler`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#Handler) interface that is consumed by the [`Webhook`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#Webhook) data structure. The `Webhook` is [registerable](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#Webhook.ServeHTTP), as an `http.Handler`, with the [`webhook.Server`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook#Server.Register).

chrisdoherty4:
> Thanks @chrisdoherty4. Question, how is this deployed?

Great question.

This will be registered with the new controller manager component (we already have a manager under `/internal/controllers`). The manager is a controller-runtime data structure that handles initialization of controllers, webhooks, clients and has responsibilities such as leader election etc. 

The `Admission` object satisfies the [`Handler`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#Handler) interface that is consumed by the [`Webhook`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#Webhook) data structure. The `Webhook` is [registerable](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook/admission#Webhook.ServeHTTP), as an `http.Handler`, with the [`webhook.Server`](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.6/pkg/webhook#Server.Register). You can see this registration happening in the [`SetupWithManager`](https://github.com/tinkerbell/tink/pull/675/files#diff-4b691c8732d3e3b19f81109b5e641ca5e8d042957dfc3810f8e82422b84ce2f0R71) function. The `webhook.Server` is 1 of the structures managed by the manager.

An example of how this is plugged together can be seen in the [kubebuilder book](https://book.kubebuilder.io/reference/webhook-for-core-types.html).

jacobweinstock (approved):

PR #679, opened 2023-03-26
Proposal: Workflow v2 protobuf definition for Tink Agent -> Tink Server communication

chrisdoherty4:
This protobuf definition is a proposed contract for Tink Worker-Server communication.


internal/proto/workflow/v2/workflow.proto
```diff
@@ -0,0 +1,83 @@
...
+option go_package = "github.com/tinkerbell/tink/internal/proto/workflow/v2;workflow";
+
+service WorkflowService {
+  // Workflows configures a bidirectional stream for a client to receive new workflows to run
+  // and submit workflow events.
+  rpc Workflows(stream WorkflowEvent) returns (stream Workflow) {}
+}
+
+
+message Workflow {
```
chrisdoherty4, line 14: Users may try to delete workflows mid-execution. How does the server communicate 'stop'?

PR #680, opened 2023-03-26
PoC: Implement tink agent

chrisdoherty4:
This is a PoC PR raised for discussion purposes. We intend on pursuing smaller PRs to merge work.


internal/agent/runtime.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+)
+
+type ContainerRuntime interface {
+	// PullImage pulls the image specified by the given action into the local registry.
+	PullImage(Action) error
+
+	// FromAction creates a new ActionContainer instance based on the Action.
+	ContainerFromAction(Action) (ActionContainer, error)
```
chrisdoherty4, line 12: Consider `NewContainer`

internal/agent/event.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "context"
+
+type EventRecorder interface {
+	RecordActionStarted(context.Context, ActionUID) error
+	RecordActionSucceeded(context.Context, ActionUID) error
+	RecordActionFailed(_ context.Context, _ ActionUID, reason, message string) error
+}
+
+type ActionUID struct {
```
chrisdoherty4, line 11: Can we get a better name than this? Feels peculiar even if its accurate.

internal/agent/workflow.go
```diff
@@ -0,0 +1,107 @@
...
+				Reason:     FailureReasonRuntimeError,
+				Message:    err.Error(),
+			})
+		}
+
+		start := time.Now()
+		logger.Info("Starting action")
+
+		started := event.ActionStarted{ID: action.ID, WorkflowID: wflw.ID}
+		if err := events.RecordEvent(ctx, started); err != nil {
```
chrisdoherty4, line 49: Need to add a timestamp to the event

internal/agent/event/zz_from_package.go
```diff
@@ -0,0 +1,16 @@
+package event
```
displague, line 1: This filename suggests that it is generated.

chrisdoherty4: There's a very clear file comment explaining otherwise. Assuming you mean the `zz`, I'm not against removing it.

PR #729, opened 2023-05-10
Add GOBIN for protoc binaries installs:

jacobweinstock:
## Description

This puts the protoc binaries in our `bin` directory so that they are available when we update the PATH at the top of the Makefile.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,5 +240,5 @@ yamllint: $(YAMLLINT_BIN)
 
 .PHONY: _protoc ## Install all required tools for use with this Makefile.
 _protoc:
-	$(GO) install $(PROTOC_GEN_GO)
-	$(GO) install $(PROTOC_GEN_GO_GRPC)
+	GOBIN=$${PWD}/bin $(GO) install $(PROTOC_GEN_GO)
+	GOBIN=$${PWD}/bin $(GO) install $(PROTOC_GEN_GO_GRPC)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -240,5 +240,5 @@ yamllint: $(YAMLLINT_BIN)
 
 .PHONY: _protoc ## Install all required tools for use with this Makefile.
 _protoc:
-	$(GO) install $(PROTOC_GEN_GO)
-	$(GO) install $(PROTOC_GEN_GO_GRPC)
\ No newline at end of file
+	GOBIN=$${PWD}/bin $(GO) install $(PROTOC_GEN_GO)
```
chrisdoherty4, line 243: I don't think `make generate-proto` will find these binaries unless someone adds `$PWD/bin` so their path.

I'd be keen to stick them under `out/bin` or `out/tools` also. I intend on moving the binaries we build to `out/bin` at some point so we have easy clean up.

jacobweinstock: the makefile does add `$PWD/bin` to the PATH. https://github.com/tinkerbell/tink/blob/eef654590132fbd291449904d3385100e386120f/Makefile#L4

chrisdoherty4 (approved):

chrisdoherty4:
@mergifyio queue

PR #730, opened 2023-05-11
Add docker runtime for agent

chrisdoherty4:
# Summary

This PR introduces a Docker runtime that is usable by the agent to run actions. The code is intended to introduce a functional runtime, therefore it doesn't implement the full set of features available on actions such as volumes; the additional features will be implemented separately.

This PR also doesn't address debugability of actions - for example, action logs are discarded -  though this may not be required if we continue with syslog and routing container logs to an external service.

# Future work

The tests are integration tests that ensure the code is robust enough. Some functionality like ensuring volume configuration, environment variables, commands and args are all configured correctly would be better implemented as unit tests. A subsequent PR will look to introduce an interface so we can easily test these feature as they're added.


internal/ptr/ptr.go
```diff
@@ -0,0 +1,90 @@
+/*
```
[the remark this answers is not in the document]
chrisdoherty4: We tend to avoid [top level util packages](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common) and sign up to every package having at least some scoped purpose.

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	// as being the command launched in the container hence it is used as the entrypoint. Args
+	// on the action are therefore the command portion in Docker.
+	if a.Cmd != "" {
+		cfg.Entrypoint = append(cfg.Entrypoint, a.Cmd)
+	}
+	if len(a.Args) > 0 {
+		cfg.Cmd = append(cfg.Cmd, a.Args...)
+	}
+
+	// TODO: Figure out container logging. We probably want to save it somewhere for debugability.
```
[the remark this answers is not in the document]
chrisdoherty4: Possibly. I've never looked into what Docker does by default.

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,191 @@
...
+func toDockerEnv(env map[string]string) []string {
+	var de []string
+	for k, v := range env {
+		de = append(de, fmt.Sprintf("%v=%v", k, v))
+	}
+	return de
+}
+
+// NewDocker creates a new Docker instance.
+func NewDocker(opts ...DockerOption) (*Docker, error) {
```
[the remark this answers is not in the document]
chrisdoherty4: Eventually we'll have this but I've refrained from adding unit tests until I implement the other features. The overview has some more details.

moadqassem (approved):
LGTM 🚀

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	return o, nil
+}
+
+// DockerOption defines optional configuration for a Docker instance.
+type DockerOption func(*Docker)
+
+// WithLogger returns an option to configure the logger on a Docker instance.
+func WithLogger(log logr.Logger) DockerOption {
```
chrisdoherty4, line 174: I've realized the logger is actually required because we log when a container couldn't be gracefully stopped. I'll tweak this when adding the other features as noted in the OP.

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,191 @@
...
+
+	create, err := d.client.ContainerCreate(ctx, &cfg, &hostCfg, nil, nil, containerName)
+	if err != nil {
+		return fmt.Errorf("docker: %w", err)
+	}
+
+	// Always try to remove the container on exit.
+	defer func() {
+		// We can't use the context passed to Run() as it may have been cancelled.
+		err := d.client.ContainerRemove(context.Background(), create.ID, types.ContainerRemoveOptions{})
```
jacobweinstock, line 94: should we enable all the `ContainerRemoveOptions`?

chrisdoherty4:
```
type ContainerRemoveOptions struct {
	RemoveVolumes bool
	RemoveLinks   bool
	Force         bool
}
```

On Mac with Docker Desktop I've seen force leave objects lying on disk that couldn't be cleaned up by a following `docker system prune --volumes` (I had to manually remove files). I don't know how widespread that impact is, or how challenging it would be to figure out, so I'm not sure if we'd want that enabled, particularly given this is all in-memory and goes away on restart. Given action names are unique I don't foresee any conflicts with respect to launching more actions.

I don't think we want to remove volumes because we allow users to share data between actions via volumes. 

I don't see any issues with link removal.

chrisdoherty4: On action name uniqueness, I've realized [the action ID is unique within the context of a workflow](https://github.com/tinkerbell/tink/blob/main/api/v1alpha2/workflow.go#L57), but if a second workflow executes there could be a collision. I think we could fix that by committing to UUIDs for action IDs.

jacobweinstock: thanks, Chris.

You're right we probably shouldn't remove volumes.

Currently, the main use case for the agent is to run in Hook. The status-quo is to use [`Force: true`](https://github.com/tinkerbell/tink/blob/cc1f8f13d92299fde6bb46a90298c720ee742711/cmd/tink-worker/worker/container_manager.go#L145), so we know it works in Hook. I'm concerned about resource constrained machines where memory and disk space in Hook are an issue.

internal/agent/runtime/docker.go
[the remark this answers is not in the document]
jacobweinstock: Just FYI, the docker lib has already defined some interfaces. https://github.com/moby/moby/blob/cb14534d7877c7eb940c97b18f228451aa8e322e/client/interface.go
Even though they are large there are ways to not have to build an implementation of all methods. For example: [here](https://github.com/jacobweinstock/tinklet/blob/main/pkg/container/docker/docker_test.go) and [here](https://github.com/tinkerbell/boots/blob/main/client/mock.go)

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,191 @@
...
+// Docker is a docker runtime that satisfies agent.ContainerRuntime.
+type Docker struct {
+	log    logr.Logger
+	client *client.Client
+}
+
+// Run satisfies agent.ContainerRuntime.
+func (d *Docker) Run(ctx context.Context, a workflow.Action) error {
+	// We need the image to be available before we can create a container.
+	image, err := d.client.ImagePull(ctx, a.Image, types.ImagePullOptions{})
```
jacobweinstock, line 31: might think about adding retries here. a transient issue to the registry could cause a whole workflow to fail.

internal/agent/runtime/docker.go
chrisdoherty4: OK, lets stick with the status quo and force delete. The volumes will likely be in-memory and cleaned on restart anyway.

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,191 @@
...
+
+// NewDocker creates a new Docker instance.
+func NewDocker(opts ...DockerOption) (*Docker, error) {
+	o := &Docker{}
+
+	for _, fn := range opts {
+		fn(o)
+	}
+
+	if o.log.GetSink() == nil {
```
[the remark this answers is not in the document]
chrisdoherty4: Thanks

internal/agent/runtime/docker.go
chrisdoherty4: I went with [github.com/avast/retry-go](github.com/avast/retry-go) as it seems quite modular and has all the bits you'd expect from a retry library.

internal/agent/runtime/docker.go
chrisdoherty4: Should be addressed.

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,202 @@
...
+		return fmt.Errorf("docker: %w", err)
+	}
+
+	// Always try to remove the container on exit.
+	defer func() {
+		// Force remove containers in an attempt to preserve space in memory constraints environments.
+		// In rare cases this may create orphaned volumes that the Docker CLI won't clean up.
+		opts := types.ContainerRemoveOptions{
+			Force:       true,
+			RemoveLinks: true,
```
chrisdoherty4, line 105: This is causing issues in tests and its not clear why. I no longer think I understand what it does so I'm disabling; we can enable it at a later date.

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	// TODO: Support all the other things on the action such as volumes.
+	cfg := container.Config{
+		Image: a.Image,
+		Env:   toDockerEnv(a.Env),
+	}
+
+	failureFiles, err := internal.NewFailureFiles()
```
jacobweinstock, line 60: Coming back to this after a bit the interaction between a runtime implementation, the agent, the eventing system, and the action, etc being run in terms of failures feels difficult to follow with just the code. Even if we had documentation (which is prone to get out of date with the code) it's feeling a bit delicate/complicated. Maybe we don't need all these places to be aware of these files? Can we talk/revisit this?

internal/agent/runtime/internal/failure_files.go
```diff
@@ -0,0 +1,104 @@
...
+// FailureFiles provides mountable files for runtimes that can be used to extract
+// a reason and message from actions.
+type FailureFiles struct {
+	reason  *os.File
+	message *os.File
+}
+
+// Close closes all files tracked by f.
+func (f *FailureFiles) Close() error {
+	os.Remove(f.reason.Name())
```
jacobweinstock, line 40: if the model is going to be to not capture the errors here, maybe we should change the interface to not return an error?

internal/agent/runtime/internal/failure_files.go
```diff
@@ -0,0 +1,104 @@
...
+	}()
+
+	var message bytes.Buffer
+	if _, err := message.ReadFrom(f.message); err != nil {
+		return "", err
+	}
+	return strings.TrimRight(message.String(), "\n"), nil
+}
+
+func (f *FailureFiles) ToError() error {
```
jacobweinstock, line 85: Code comment?

internal/agent/runtime/internal/failure_files.go
jacobweinstock: Also, the name `Close` actually doesn't feel accurate here. deleting files is different than closing them. is this satisfying an interface method?

internal/agent/runtime/docker.go
```diff
@@ -0,0 +1,204 @@
...
+
+	err := retry.Do(pullImage, retry.Attempts(5), retry.DelayType(retry.BackOffDelay))
+	if err != nil {
+		return err
+	}
+
+	// TODO: Support all the other things on the action such as volumes.
+	cfg := container.Config{
+		Image: a.Image,
+		Env:   toDockerEnv(a.Env),
```
jacobweinstock, line 57: we might want to set the failure file locations to be env vars for all containers. that way actions dont need to "know" where this location is and we can change it if needed. action writers just look up the env var and get the files.

jacobweinstock (approved):

internal/agent/runtime/internal/failure_files.go
chrisdoherty4: If we inject a data structure for failure files to runtimes I suspect this method will be unnecessary so defering.

chrisdoherty4:
In an effort to move this along we've decided to merge this. This comment serves as a summary of outstanding items:

1. Inject failure files as they should be common across runtimes.
2. Rename `Close()` for failure files to `Remove()` and drop returning an error.
3. Restructure the sources to follow Code Review comments guideliens for source organization.

PR #734, opened 2023-05-15
Add transport implementation for agent

chrisdoherty4:
Builds on #730.

The new agent requires a transport to retrieve workflows. The initial transport to be implemented is gRPC as its worked well for the tink-worker (the agents predecessor). 

The transport is responsible for connecting to the Tink Server, identifying the node by submitting the nodes unique identifier and subsequently passing along workflows. Unlike the Tink Worker implementation, this transport provides a mechanism for publishing workflow events. The events are what provides insight into the progression of the workflow and are ultimately used by the Tink Server to update the workflow object in Kubernetes.


internal/agent/agent.go
```diff
@@ -43,40 +49,94 @@ func (agent *Agent) Start(ctx context.Context) error {
...
+func (agent *Agent) CancelWorkflow(workflowID string) {
+	agent.mtx.RLock()
+	defer agent.mtx.RUnlock()
+
+	if agent.executionContext == nil {
+		agent.Log.Info("No workflow running; ignoring cancellation request", "workflow_id", workflowID)
+		return
 	}
+
+	if agent.executionContext.Workflow.ID != workflowID {
```
jacobweinstock, line 126: does the transport also handle this type of check?

internal/agent/transport/grpc.go
```diff
@@ -0,0 +1,170 @@
...
+	"io"
+
+	"github.com/avast/retry-go"
+	"github.com/go-logr/logr"
+	"github.com/tinkerbell/tink/internal/agent/event"
+	"github.com/tinkerbell/tink/internal/agent/workflow"
+	workflowproto "github.com/tinkerbell/tink/internal/proto/workflow/v2"
+)
+
+var _ event.Recorder = &GRPC{}
```
jacobweinstock, line 15: should we add a check that `&GRPC{}` implements the [`agent.Transport`](https://github.com/tinkerbell/tink/blob/29052da86f6551ce443a7c3a434da523543da389/internal/agent/transport.go#L10) interface too?

chrisdoherty4: Unfortunately that creates an import cycle - pretty common when locating interfaces next to the consumer.

internal/agent/agent.go
chrisdoherty4: No, the transport doesn't make any decisions like this.

internal/agent/transport/grpc.go
```diff
@@ -0,0 +1,170 @@
...
+		client: client,
+	}
+}
+
+type GRPC struct {
+	log    logr.Logger
+	client workflowproto.WorkflowServiceClient
+}
+
+func (g *GRPC) Start(ctx context.Context, agentID string, handler WorkflowHandler) error {
```
jacobweinstock, line 29: hmmm.....we dont have a linting check for code comments on exported functions/methods?

internal/proto/workflow/v2/workflow.proto
```diff
@@ -0,0 +1,117 @@
...
+package internal.proto.workflow.v2;
+
+option go_package = "github.com/tinkerbell/tink/internal/proto/workflow/v2;workflow";
+
+// WorkflowService is responsible for retrieving workflows to be executed by the agent and
+// publishing events as a workflow executes.
+service WorkflowService {
+  // Stream creates a stream that will receive workflows intended for the agent identified
+  // by the StreamWorkflowsRequest.agent_id.
+  rpc StreamWorkflows(StreamWorkflowsRequest) returns (stream StreamWorkflowsResponse) {}
```
jacobweinstock, line 12: thought this was interesting. Its from the Buf website. https://buf.build/docs/best-practices/style-guide#recommendations 

_"While controversial, our recommendation is to avoid streaming RPCs. While they certainly have specific use cases that make them extremely valuable, on the whole they generally cause a lot of problems, push RPC framework logic up the stack, and usually prevent developing more reliable architectures."_

jacobweinstock: Also, would you me open to a name change here? maybe `StreamWorkflows` -> `GetWorkflows`

jacobweinstock (approved):
one loosely held suggestion for a name change. other than that all good. Thanks for all the work on this!

internal/proto/workflow/v2/workflow.proto
chrisdoherty4: > thought this was interesting. Its from the Buf website. [buf.build/docs/best-practices/style-guide#recommendations](https://buf.build/docs/best-practices/style-guide#recommendations)

I'm familiar with this but chose to ignore it because the alternatives I thought about have tradeoffs also. The options I came up with include: (1) polling from the agent to the server for workflows; (2) launching a server in the agent and have the tink-server (now a confusing name) issue unary calls to run workflows. 

I wasn't fond of polling even with HTTP2s sticky connections and I wanted to, initially at least, avoid the complexity of launching a server and dealing with the agent going offline etc. 

I very open to adopting a different approach and welcome more ideas. Long term I'm interested in migrating to the agent containing a server even though its more complex.

> Also, would you me open to a name change here? maybe StreamWorkflows -> GetWorkflows

Yes, no problem.

internal/agent/transport/grpc.go
chrisdoherty4: Apparently not 🤷🏻 . We need to tweak the `ci-scripts` stuff and align Tinkerbell repositories. As always, a low priority task.

jacobweinstock (approved):

chrisdoherty4:
@mergifyio queue

PR #762, opened 2023-06-22
Add tink-agent binary

chrisdoherty4:
_Depends on #734_

This PR adds the tink-agent binary to that includes minimal configuration to enable launching an agent with connectivity over gRPC to a server.


internal/cli/agent.go
```diff
@@ -0,0 +1,59 @@
...
+				ID:        opts.AgentID,
+				Transport: trnport,
+				Runtime:   rntime,
+			}).Start(cmd.Context())
+		},
+	}
+
+	flgs := cmd.Flags()
+	flgs.StringVar(&opts.AgentID, "agent-id", "", "An ID that uniquely identifies the agent instance")
+	flgs.StringVar(&opts.TinkServerAddr, "tink-server-addr", "127.0.0.1:42113", "Tink server address")
```
chrisdoherty4, line 56: Not sure about the option here. It might be better to use `grpc-addr` complimenting the idea that we could have multiple transports. Introducing the file transport may inform us better.

jacobweinstock (approved):

PR #763, opened 2023-06-22
Add file based agent transport

chrisdoherty4:
This change adds a file based transport that will execute a single workflow. It is useful for testing the agent e2e.


chrisdoherty4:
The previous iteration scanned a directory. However, because the `WorkflowHandler` does block and there was no tracking of running workflows the implementation would fire all workflows with everything except the first failing. 

Given the purpose of this addition is to enable easy e2e testing, I've reduced the implementation to something similar to a suggestion by @jacobweinstock that executes a single file.

jacobweinstock (approved):

chrisdoherty4:
@mergifyio queue

PR #807, opened 2023-10-06
Expand Hardware spec fields available as template data in Templates:

jacobweinstock:
## Description

This improves the templating of Templates. This was requested by quite a few folks in the community.

## Why is this needed

Fixes: #

## Checklist:

I have:

- [ ] added unit or e2e tests

### DIFF
--- a/docs/Template.md
+++ b/docs/Template.md
@@ -1,0 +1,26 @@
+# Templating a Template
+
+For this document we will be explaining how to templatize your Template. This will allow you to use data that lives in your Hardware objects to populate values in your Template object. Let's get started.
+
+## Background
+
+In Tinkerbell, we have the concept of a Template. This is a defined as a Kubernetes Custom Resource (CR). It describes the steps that will run during a workflow. Look at the CR Definition (CRD) [here](../config/crd/bases/tinkerbell.org_templates.yaml) and an example Template object [here](../config/crd/examples/template.yaml). We also have the concept of Hardware. This is also a Kubernetes Custom Resource (CR). It describes your physical hardware. Look at the CR Definition (CRD) [here](../config/crd/bases/tinkerbell.org_hardware.yaml) and an example Hardware object [here](../config/crd/examples/hardware.yaml).
+
+## Templating
+
+Inside of this Template there is the ability to use a templating language to get data from your Hardware objects. See the docs [here](https://developer.hashicorp.com/nomad/tutorials/templates/go-template-syntax) for more details on use of the templating language. We don't expose the full Hardware object to the templating engine. We have created an abstraction on top of the Hardware object that is available to you. This was done so that we can change the underlying Hardware spec without breaking your Templates. Right now we expose the following fields:
+
+### v1alpha1.Hardware
+
+| Spec Field                            | Template Access                               | Field Type    | Example                                                   |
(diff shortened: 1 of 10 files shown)


internal/workflow/reconciler.go
```diff
@@ -126,7 +126,26 @@ func (r *Reconciler) processNewWorkflow(ctx context.Context, logger logr.Logger,
...
+	Interfaces []netInterface
+}
+
+// netInterface defines a network interface for a Hardware instance.
+// It is different than v1alpha1.Interface so that we don't couple ourselves to the API,
+// especially when we already have a v1alpha2.
+type netInterface struct {
+	MAC    string
+	VLANID string
+	DHCP   dhcp
```
chrisdoherty4, line 139: I know we've been around this loop but lets go again: do we think DHCP is here to stay with the network interface data? I.e. would we introduce any feature where DHCP isn't specified? If so, how can I easily determine if the DHCP config is available from template language?

internal/workflow/reconciler.go
```diff
@@ -126,7 +126,26 @@ func (r *Reconciler) processNewWorkflow(ctx context.Context, logger logr.Logger,
 
 // templateHardwareData defines the data exposed for a Hardware instance to a Template.
 type templateHardwareData struct {
-	Disks []string
+	Disks      []string
+	Interfaces []netInterface
```
chrisdoherty4, line 130: In the v1alpha2 API we're doubling down on clarity by using `NetworkInterfaces`. Should we do the same here?

jacobweinstock: Sure thing.

internal/workflow/reconciler.go
chrisdoherty4: I think I'm inventing possibilities to challenge this because its meant to be a stable contract. I don't know how productive trying to provide stability here really is though. I feel like we'd be better off just plumbing through the raw `Hardware.Spec` and when that becomes stable this contract inherently becomes stable.

Given we've had numerous requests to get various fields, mostly network interface data, how do you feel about including that and declaring this 'stable' contract a deprecated experiment?

docs/Template.md
```diff
@@ -0,0 +1,72 @@
...
+        worker: "{{.device_1}}"
+```
+
+## Templating functions
+
+There are a number of built in functions that Go provides and that can be used in your templating. See [here](https://developer.hashicorp.com/nomad/tutorials/templates/go-template-syntax#function-list). Tinkerbell has also defined a few custom functions that can be used.
+
+| Function Name     | Description | Use     | Examples |
+| -------------     | ----------- | ------- | -------- |
+| `contains`        | Contains returns a bool for whether `substr` is within `s`. | `{{ contains "HELLO" "H" }}` | `contains <s> <substr>` |
```
chrisdoherty4, line 69:
```suggestion
| `contains`        | contains returns a bool for whether `substr` is within `s`. | `{{ contains "HELLO" "H" }}` | `contains <s> <substr>` |
```

internal/workflow/reconciler.go
chrisdoherty4: @jacobweinstock What did we conclude - I don't recall?

chrisdoherty4 (dismissed):

chrisdoherty4 (approved):

PR #825, opened 2023-10-24
Update Event interface code doc

chrisdoherty4:
The `Event` interface is implemented by all events. The code docs were not clear on how a consumer leverages the `Event` interface to handle the event (for example transport the event).

This code doc clarifies the intended usage of the Event interface.

### DIFF
--- a/internal/agent/event/event.go
+++ b/internal/agent/event/event.go
@@ -7,7 +7,18 @@ import "context"
 // Name is a unique name identifying an event.
 type Name string
 
-// Event is a recordable event.
+// Event is an event generated during execution of a Workflow. Each event in the event package
+// implements this interface. Consumers may type switch the Event to the appropriate type for
+// event handling.
+//
+// E.g.
+//
+//	switch ev.(type) {
+//	case event.ActionStarted:
+//		// Handle ActionStarted event.
+//	default:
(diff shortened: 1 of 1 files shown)


moadqassem (approved):

internal/agent/event/event.go
```diff
@@ -7,7 +7,17 @@ import "context"
 // Name is a unique name identifying an event.
 type Name string
 
-// Event is a recordable event.
+// Event is a recordable event. Each event in the event package implements this interface. Consumers
```
jacobweinstock, line 10: This is great, thanks. A few suggestions for things to add that I think would be helpful for implementors. 
- what "recordable event" is, means, and when/where events are recorded (i believe its just in the agent logic).
- transports implementations are responsible for sending the event. 
- concrete event types will contain more context that can be used when sending an event.

chrisdoherty4: I can add something for the first 2. Isn't the third point covered by the example?

chrisdoherty4: I've tweaked this a little but I decided not to make any mention of transport because thats 1 of possibly more consumers. I think the documentation in this package needs to be tight and relevant to the package itself.

jacobweinstock (approved):

PR #826, opened 2023-10-24
Refactor v1alpha2 Linux namespace definitions

chrisdoherty4:
There are a handful of Linux namespaces (see https://man7.org/linux/man-pages/man7/namespaces.7.html). Containers can be configured to run inside the bounds of precreated namespaces. The original v1alpha2 API allowed users to specify a network namespace only, but we have already acquired a use-case for the PID namespace. 

This change refactors the v1alpha2 API (unreleased) to accommodate future namespace customization.


api/v1alpha2/template.go
```diff
@@ -49,11 +49,9 @@ type Action struct {
 	// +optional
 	Volumes []Volume `json:"volumes,omitempty"`
 
-	// NetworkNamespace defines the network namespace to run the container in. This enables access
-	// to the host network namespace.
-	// See https://man7.org/linux/man-pages/man7/namespaces.7.html.
+	// Namespaces defines the Linux namespaces this container should execute in.
 	// +optional
-	NetworkNamespace *string `json:"networkNamespace,omitempty"`
+	Namespaces *Namespaces `json:"namespaces,omitempty"`
```
moadqassem, line 54: Shouldn't it be called `Namespace` instead of `Namespaces`? It has a struct type and within the struct there are no arrays/list defined.

api/v1alpha2/template.go
```diff
@@ -71,6 +69,18 @@ type Action struct {
 // See https://docs.docker.com/storage/volumes/ for additional details.
 type Volume string
 
+// Namespace defines the Linux namespaces to use for the container.
```
moadqassem, line 72: Based on the type name and the field above in the Actions I assume this is Namespaces :)

api/v1alpha2/template.go
chrisdoherty4: I pluralized it because there are multiple namespaces in the Namespaces data structure.

You can specify multiple namespaces simultaneously.

moadqassem: Sure, but in this context, you only have can one namespace per Action, or am I missing something here? I mean I found it a bit confusing, especially if you are looking at it from a pure data structural point of view ;)

chrisdoherty4: I see what you mean. I'll update.

jacobweinstock (approved):

moadqassem (approved):
LGTM

PR #860, opened 2024-01-16
Update grpc telemetry instrumentation with new APIs

chrisdoherty4:
When updating the gRPC Otel libraries in a dependabot update the verify command passed. However, there are deprecated APIs that we were still using. This updates Tink Worker and Tink Server to use the new Otel gRPC API.

### DIFF
--- a/buf.lock
+++ b/buf.lock
@@ -4,4 +4,4 @@ deps:
   - remote: buf.build
     owner: googleapis
     repository: googleapis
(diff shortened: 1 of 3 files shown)


chrisdoherty4:
@mergify refresh

chrisdoherty4:
Merged in https://github.com/tinkerbell/tink/pull/863

PR #874, opened 2024-01-29
v1alpha2 Workflow Controller

chrisdoherty4:
Introduce basic functionality for a v1alpha2 Workflow Controller. This implementation is independent of the existing controller as we won't support the v1alpha1 API when we move to v1alpha2. The first commit moves deprecated code into /internal/deprecated and isn't that interesting: the work is mostly in the [second commit](https://github.com/tinkerbell/tink/pull/874/commits/eb5eb09b37ab79ea4c346422712d9e3929beacc0).

The implementation renders the template actions onto the Workflow only which will aid in E2E tests with Tink Server. Logic such as remediating Workflow states is deferred to a later PR.


Makefile
```diff
@@ -49,21 +49,21 @@ help: ## Print this help
 VERSION ?= $(shell git rev-parse --short HEAD)
 
 # Define all the binaries we build for this project that get packaged into containers.
-BINARIES := tink-server tink-agent tink-worker tink-controller virtual-worker
+BINARIES := tink-server tink-agent tink-worker tink-controller tink-controller-v1alpha2 virtual-worker
```
jacobweinstock, line 52: not too excited we have to use `tink-controller-v1alpha2` :(

api/v1alpha2/conditions.go
```diff
@@ -30,7 +30,7 @@ type Condition struct {
 	Status ConditionStatus `json:"status"`
 
 	// LastTransition is the last time the condition transitioned from one status to another.
-	LastTransition *metav1.Time `json:"lastTransitionTime"`
+	LastTransition metav1.Time `json:"lastTransitionTime"`
```
jacobweinstock, line 33: I know you love pointers ;) so why move away from one here?

cmd/tink-controller-v1alpha2/main.go
```diff
@@ -0,0 +1,179 @@
+package main
```
jacobweinstock, line 1: is v1alpha2 so different from a cmd/cli perspective that we have to break it out entirely? Could a flag be used in the existing tink-controller?

cmd/tink-controller-v1alpha2/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM alpine:3.15
```
jacobweinstock, line 1: any reason not to use `scratch`?

cmd/tink-controller-v1alpha2/main.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	config.AddFlags(cmd.Flags())
+	return cmd
+}
+
+func createViper(logger logr.Logger) (*viper.Viper, error) {
+	v := viper.New()
+	v.AutomaticEnv()
+	v.SetConfigName("tink-controller")
+	v.AddConfigPath("/etc/tinkerbell")
```
jacobweinstock, line 140: do we need a default location for a config file? does this provide much value, you think?

internal/workflow/internal/reconcile_test.go
```diff
@@ -0,0 +1,135 @@
+package internal_test
```
jacobweinstock, line 1: im not fond of `internal_test` files that use dot notation import for the package it's testing. I don't see the value. Not anything to hold up the PR, just wanted to voice it. : )

internal/workflow/reconciler.go
```diff
@@ -2,171 +2,72 @@ package workflow
...
 }
 
 // +kubebuilder:rbac:groups=tinkerbell.org,resources=hardware;hardware/status,verbs=get;list;watch;update;patch
 // +kubebuilder:rbac:groups=tinkerbell.org,resources=templates;templates/status,verbs=get;list;watch;update;patch
-// +kubebuilder:rbac:groups=tinkerbell.org,resources=workflows;workflows/status,verbs=get;list;watch;update;patch;delete
+// +kubebuilder:rbac:groups=tinkerbell.org,resources=workflows;workflows/status,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=tinkerbell.org,resources=workflows;workflows/finalizers,verbs=update
 
-func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+func (r *Reconciler) Reconcile(ctx context.Context, req reconcile.Request) (result reconcile.Result, rerr error) {
```
jacobweinstock, line 36: are we gaining any value by having named return values? I don't see them being used, feels unnecessary.

internal/workflow/reconciler.go
```diff
@@ -2,171 +2,72 @@ package workflow
...
-			if action.Status == v1alpha1.WorkflowStateRunning && action.StartedAt != nil &&
-				r.nowFunc().After(action.StartedAt.Add(time.Duration(action.Timeout)*time.Second)) {
-				// Set fields on the timed out action
-				stored.Status.Tasks[ti].Actions[ai].Status = v1alpha1.WorkflowStateTimeout
-				stored.Status.Tasks[ti].Actions[ai].Message = "Action timed out"
-				stored.Status.Tasks[ti].Actions[ai].Seconds = int64(r.nowFunc().Sub(action.StartedAt.Time).Seconds())
-				// Mark the workflow as timed out
-				stored.Status.State = v1alpha1.WorkflowStateTimeout
-			}
+	// Always attempt to patch.
```
jacobweinstock, line 59: Why always attempt a patch?

jacobweinstock (commented):
Thanks for the work on this, @chrisdoherty4. Left some comments for discussion.

cmd/tink-controller-v1alpha2/main.go
chrisdoherty4: I'm open to using a single binary. This temporary extra binary just felt simpler to me. There won't be a transition period where both APIs are supported so eventually `tink-controller-v1alpha2` will become `tink-controller`.

cmd/tink-controller-v1alpha2/Dockerfile
chrisdoherty4: I don't know. The goal of the PR is to get something functional so I've not experimented.

cmd/tink-controller-v1alpha2/main.go
chrisdoherty4: Probably not, copy-n-paste from the existing tink-controller. If we maintain the temporary v1alpha2 binary, I'll reduce this down to bare minimum so we don't end up with cruft down the road.

internal/workflow/internal/reconcile_test.go
chrisdoherty4: Do you mean specifically for `internal` packages, or do you mean generally the `_test` idiom with dot import?

jacobweinstock (approved):

jacobweinstock:
Hey @chrisdoherty4 , feel free to merge as is. We can address these before the official removal of v1alpha1.

internal/workflow/reconciler.go
chrisdoherty4: The named returns are included for the defer to work because we change the return variable if patching fails. I don't think the code needs to work this way, and honestly I'm not fond of the defer approach - defering the patch is a common method for always patching.

internal/workflow/reconciler.go
chrisdoherty4: The rationale is that you do as much as possible to bring the resource to the desired state and you shouldn't update the `.Status` state if nothing changed so always patching ensures we always try to persist the updates.

internal/workflow/reconciler.go
chrisdoherty4: Well, that's a good pont: `_`.

jacobweinstock (approved):

PR #960, opened 2024-07-03
add tinkerbell-insecure-tls param to control InsecureSkipVerify

rpardini:
#### add tinkerbell-insecure-tls param to control InsecureSkipVerify

- this allows using TLS but without verifying certificates/CAs/hostnames etc
- fix e2e tests for new tlsInsecure parameter
- add `// #nosec G402` so we can actually use InsecureSkipVerify
- make gofumpt happy

Signed-off-by: Ricardo Pardini <ricardo@pardini.net>

### DIFF
--- a/cmd/tink-worker/cmd/root.go
+++ b/cmd/tink-worker/cmd/root.go
@@ -57,6 +57,7 @@ func NewRootCommand(version string) *cobra.Command {
 			conn, err := client.NewClientConn(
 				viper.GetString("tinkerbell-grpc-authority"),
 				viper.GetBool("tinkerbell-tls"),
+				viper.GetBool("tinkerbell-insecure-tls"),
 			)
 			if err != nil {
 				return err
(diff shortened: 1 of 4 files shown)


rpardini:
Small justification here: when using an Ingress in front of tink (eg: ingress-nginx), having TLS enabled is essential, as non-TLS gRPC is generally not supported by Ingresses (as it would require disabling http/1.1 support in favor of http/2 on port 80). But enabling TLS before this _implied_ the full TLS verification as well (CA/certs/CN+SAN matching etc). This allows to have TLS enabled, but with InsecureSkipVerify. It defaults to false so no unexpected changes should be introduced.

rpardini:
updated after making CI pass (gofumpt, gosec, etc)

rpardini:
Reference https://github.com/kubernetes/ingress-nginx/issues/3897

jacobweinstock (approved):

